PR #32, opened 2025-05-22
Add CI workflows for CPU and GPU

rtburns-jpl:


.github/conda_envs/environment.yml
```diff
@@ -0,0 +1,31 @@
...
+channels:
+  - conda-forge
+  - nodefaults
+dependencies:
+  - backoff
+  - cmake>=3.19
+  - cxx-compiler
+  - eigen>=3.3
+  - fftw>=3.3
+  - gdal>=3.0,!=3.9.2,<3.11
```
gmgunter: Why are we capping the GDAL version here?

.github/conda_envs/environment.yml
```diff
@@ -0,0 +1,31 @@
...
+  - gmock>=1.10
+  - gtest>=1.10
+  - h5py>=3.5
+  - hdf5>=1.10.2
+  - libgdal-hdf5
+  - libgdal-netcdf
+  - ninja
+  - numpy>=1.20,<2
```
gmgunter: Same here. I think isce3 supports numpy>=2, doesn't it?

.github/conda_envs/environment.yml
```diff
@@ -0,0 +1,31 @@
+name: isce3
+channels:
+  - conda-forge
+  - nodefaults
+dependencies:
+  - backoff
+  - cmake>=3.19
```
gmgunter: This is inconsistent with the minimum CMake version in the top-level `CMakeLists.txt`

https://github.com/isce-framework/isce3/blob/1af2d3b5ad885b3adfcc516c17e8aa95eafaa23e/CMakeLists.txt#L1

.github/conda_envs/environment.yml
```diff
@@ -0,0 +1,31 @@
...
+  - gdal>=3.0,!=3.9.2,<3.11
+  - gmock>=1.10
+  - gtest>=1.10
+  - h5py>=3.5
+  - hdf5>=1.10.2
+  - libgdal-hdf5
+  - libgdal-netcdf
+  - ninja
+  - numpy>=1.20,<2
+  - pandas
```
gmgunter: pandas isn't a direct dependency anymore since it was removed in https://github-fn.jpl.nasa.gov/isce-3/isce/pull/2196

```suggestion
```

.github/conda_envs/environment.yml
gmgunter, line 1: Could we just use the [`environment.yml`](https://github.com/isce-framework/isce3/blob/9ef0fdd43add2392ddb8538a45b2f864615ae2d4/environment.yml) file in the root directory of the repo instead of adding this file to `.github/conda_envs/`?


Your CI pipelines are setup to allow you to supply extra package specs to further constrain the environment, if necessary. (I was hoping we could also use this for the "minimum reqs" job instead of creating a separate `minimum-reqs.yml` file, but don't think it's possible to remove the `pyre` dependency this way, hmm...)

https://github.com/isce-framework/isce3/blob/9ef0fdd43add2392ddb8538a45b2f864615ae2d4/.github/workflows/build-and-run.yml#L69-L73

tests/python/extensions/pybind/core/orbit.py
```diff
@@ -6,6 +6,7 @@
 import h5py
 import tempfile
 import isce3
+import journal
```
gmgunter: Unused import

```suggestion
```

.github/workflows/build-and-run-gpu.yml
```diff
@@ -0,0 +1,127 @@
+name: Build-And-Run Tests on GPU
+
+on:
+  pull_request:
+    types: [ labeled, opened, synchronize, reopened ]
```
gmgunter, line 7: Let's add a comment here that clarifies that the default `types` are `opened`, `synchronize`, and `reopened`, to which we are adding `labeled`.

https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request

```suggestion
    # Trigger the workflow whenever a PR receives a special label
    # in addition to the default triggering events.
    types: [ labeled, opened, synchronize, reopened ]
```

.github/workflows/build-and-run-gpu.yml
```diff
@@ -0,0 +1,127 @@
...
+  push:
+    branches:
+      - 'ci-updates-gpu'
```
gmgunter: I think this was just used for your initial testing and could be removed now.

```suggestion
```

.github/workflows/build-and-run-gpu.yml
```diff
@@ -0,0 +1,127 @@
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref }}
+  cancel-in-progress: true
+
+jobs:
+  build_and_test_gpu:
+    # Trigger on pushes to CI updates branch, or PRs with specific label
+    if: |
+      github.event_name == 'push' ||
+      contains(github.event.pull_request.labels.*.name, 'GPU')
```
gmgunter:
```suggestion
    # Trigger on PRs with specific label
    if: contains(github.event.pull_request.labels.*.name, 'GPU')
```

.github/workflows/build-and-run.yml
```diff
@@ -0,0 +1,128 @@
+name: Build-And-Run Tests
+
+on: [pull_request, push]
```
gmgunter: Is the intention that this would run on pushes to *all* branches, or just pushes to `develop`?

I have something like this in the snaphu-py:

```yaml
on:
  pull_request:
  push:
    branches:
      - develop
      - release-*
    tags:
      - v*.*.*
```

So that the CI is triggered only by PRs and pushes to specific branches/tags, rather than pushes to any branch.

gmgunter (commented):
Thanks @rtburns-jpl, exciting stuff! Few questions/suggestions below.

.github/workflows/build-and-run.yml
rtburns-jpl: I was thinking that all our branches here will be develop or release branches, so we'll want to run this workflow on all of them

gmgunter: This `ci` branch isn't a develop or release branch! 🤔

image

rtburns-jpl: Yeah, I think I have to do changes to github action workflows on branches on the main repo, because I thought changes to these files from forks are ignored for security reasons

.github/conda_envs/environment.yml
rtburns-jpl: I think I added this previously when the workflow was building pyre from source, because pyre needs 3.19+. I don't think that happens anymore so I'll try changing it back.

.github/conda_envs/environment.yml
rtburns-jpl: Hehe, this line is old enough that isce3 didn't fully support numpy 2 when written. Will remove

.github/conda_envs/minimum-reqs.yml
```diff
@@ -0,0 +1,32 @@
+name: isce3
+channels:
+  - conda-forge
+  - nodefaults
+dependencies:
+  - backoff
+  - cmake=3.19
```
gmgunter: Let's add a comment here explaining why this isn't `cmake=3.18`

```suggestion
  # Building pyre from source requires cmake>=3.19
  - cmake=3.19
```

.github/conda_envs/minimum-reqs.yml
gmgunter, line 1: In the future, it'd be nice to have a job (or jobs) that test the minimum supported versions of `hdf5`, `pybind11`, `python`, and `shapely` as well, since I believe we've run into issues with API breakages in each of those at some point in the past.

In addition, the top-level [`environment.yml`](https://github.com/isce-framework/isce3/blob/1af2d3b5ad885b3adfcc516c17e8aa95eafaa23e/environment.yml) file specifies minimum supported versions of `h5py`, `numpy`, and `pysolid`, but we're not constraining their versions here. Ideally, we should add a CI job that ensures we maintain compatibility with those versions.

It might be necessary to break the "Minimum" job up into multiple jobs that each test a distinct subset of minimum dependencies in order to avoid over-constraining the conda environment specs.

This is a great start for now, though.

.github/workflows/build-and-run-gpu.yml
```diff
@@ -0,0 +1,124 @@
...
+    types: [ labeled, opened, synchronize, reopened ]
+
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref }}
+  cancel-in-progress: true
+
+jobs:
+  build_and_test_gpu:
+    # Trigger on pushes to CI updates branch, or PRs with specific label
```
gmgunter: This comment is out-of-date now

```suggestion
    # Trigger on PRs with specific label
```

.github/workflows/build-and-run.yml
gmgunter: >I was thinking that all our branches here will be develop or release branches, so we'll want to run this workflow on all of them

https://github.com/isce-framework/isce3/pull/32#discussion_r2103452198

Hmm, I still think it's correct to constrain which branches this workflow will trigger on.

Even if we ensure that all non-develop, non-release branches are created on forks, this CI workflow will still run on pushes to the forked repo. So, if I push a commit to a PR originating from a fork, the workflow will run twice -- once in the upstream repo and once in my forked repo. This seems like an unnecessary waste of compute resources and also means that I'll get twice as many email notifications if I push a bad commit that causes the CI to fail.

```suggestion
on:
  pull_request:
  push:
    branches:
      - develop
      - release-*
    tags:
      - v*.*.*
```

.github/conda_envs/minimum-reqs.yml
```diff
@@ -0,0 +1,32 @@
...
+  - gxx_linux-64=11
+  - eigen=3.3
+  - fftw=3.3
+  # GDAL 3.0 is too old to satisfy python>=3.9
+  # GDAL 3.1 is too old to satisfy hdf5>=1.12
+  # GDAL 3.6.0 was the first to build with cmake, needed for FindGDAL module
+  # https://github.com/conda-forge/gdal-feedstock/pull/651
+  - gdal>=3.6
```
gmgunter: Typo? Shouldn't we ensure that this job builds with GDAL 3.6.X (and not some newer GDAL version)?

```suggestion
  - gdal=3.6
```

environment.yml
```diff
@@ -11,14 +11,14 @@ dependencies:
   - gdal>=3
```
gmgunter:
```suggestion
  - gdal>=3.6
```

Should we specify `gdal>=3.6` here now since that seems to be the de facto minimum supported version?

https://github.com/isce-framework/isce3/blob/3040629593a08a536e57923204251f8407501787/.github/conda_envs/minimum-reqs.yml#L11-L15

environment.yml
```diff
@@ -11,14 +11,14 @@ dependencies:
...
-  - h5py>=3
+  - h5py>=3.5
   - hdf5>=1.10.2,!=1.14.0
   - libgdal-hdf5
   - libgdal-netcdf
   - ninja
   - numpy>=1.20
   - pyaps3>=0.3
-  - pybind11>=2.5
+  - pybind11>=2.6
```
gmgunter: Can you explain these changes to the minimum `h5py` and `pybind11` versions?

rtburns-jpl: Hmm, I want to say these were from my minimum-requirements testing, and were needed because of specific features in h5py/pybind11 that we are using in isce3 , but I think my commit messages for these changes got squashed down so I'm not able to check why. I'll remove these constraints for now but we may need to add them back in the future.

.github/conda_envs/minimum-reqs.yml
rtburns-jpl: Yes, completely agree, and I think making a minimum requirements job to test just a single dependency is satisfied will be easier that trying to resolve an entire environment where everything is at its minimum supported version.

.github/conda_envs/minimum-reqs.yml
rtburns-jpl: I was utterly unable to get an environment to resolve with gdal 3.6 and had to drop the hard constraint. It may be possible to pin it to a later version that's still less than our latest-build version, not sure.

gmgunter: How about we drop the min-reqs job for now, iterate on it some more, and revive it in a subsequent PR?

It looks like the GPU runner and the rest of the CI build matrix are working well -- the min-reqs job is the only holdup from this being merged and proceeding forward with public GitHub migration. The existing internal CI system doesn't test minimum package versions, so there's no regression if we remove it for now.

Some more iteration could let us re-structure the min-reqs job as multiple jobs, each targeting a different dependency (or set of dependencies).

WDYT?

rtburns-jpl: Yeah I think that's a good idea. Maybe the min-reqs could be structured as you proposed above - one min-reqs jobs for each dependency we want to enforce a minimum version on. In any case, I agree that we can cook on this a little longer, while the rest of the CI matrix is already in a good place.

gmgunter (approved):
LGTM, thanks!

.github/workflows/build-and-run-gpu.yml
```diff
@@ -0,0 +1,124 @@
...
+          environment-file: ${{ matrix.deps.env-file }}
+          environment-name: isce3
+          create-args: >-
+            ${{ matrix.deps.extra-specs }}
+
+      # Fix missing conda command on macOS
+      - name: macOS conda command fixup
+        if: runner.os == 'macOS'
+        run: |
+          micromamba install conda
```
Tyler-g-hudson, line 73: I don't see any macOS runners for this workflow - is this left in with the intention to add them in later?

.github/workflows/build-and-run-gpu.yml
```diff
@@ -0,0 +1,124 @@
...
+        run: |
+          cmake --build build --target install --parallel
+
+      # Set the LD_LIBRARY_PATH and DYLD_LIBRARY_PATH variables
+      - name: Set Link Library Environment Variable
+        # Append to DYLD_LIBRARY_PATH on macOS or to LD_LIBRARY_PATH on Linux.
+        # Depending on the Linux flavor, the install dir name could be 'lib64' or 'lib'.
+        run: |
+          if [ "$RUNNER_OS" == "macOS" ]; then
+            conda env config vars set DYLD_LIBRARY_PATH=$DYLD_LIBRARY_PATH:$ISCE3_PREFIX/lib
```
Tyler-g-hudson, line 115: Same question as above regarding macOS runners

.github/workflows/build-and-run-gpu.yml
```diff
@@ -0,0 +1,124 @@
...
+      matrix:
+        os:
+          - label: Linux
+            runner: ubuntu-gpu-t4
+            ctest_exclude: -E ".*($ALL_EXCLUDE)"
+        deps:
+          - label: Latest
+            env-file: environment.yml
+            extra-specs: ""
+        build-type: [ RelWithDebInfo ]
```
Tyler-g-hudson, line 42: I don't recall when we decided which build-types we'd use in the GHA CI system, but the GPU runner uses RelWithDebInfo while the CPU runners use either Release or Debug - could you document the reasoning for these selections in the files?

Tyler-g-hudson (approved):
LGTM! Some small items to look at, none of them blocking approval

.github/workflows/build-and-run-gpu.yml
rtburns-jpl: Sure - Release and Debug are two opposite extremes so it's nice to test with both when possible, so that's what I'm doing for the free CPU-only builds. The GPU builds cost us per-minute, so I'm setting up a single GPU build and avoiding a whole matrix of possibilities for the build. RelWithDebInfo makes the tests fast but still builds with debugging checks, so it seems like the best compromise if you need to just pick one.

.github/workflows/build-and-run-gpu.yml
rtburns-jpl: Yeah, no macOS builds for GPU, and I don't think that's possible since CUDA doesn't support macOS. I'm just keeping this file as similar to the CPU workflow as possible so it's easier to port over changes from one to the other - but yeah, this will never execute on the GPU workflow.

.github/workflows/build-and-run-gpu.yml
Tyler-g-hudson: Good - let's add a comment here and in the same line on CPU that explains this decision since it seems like it could be useful to refresh on when you or someone else come back to modify the file later on
```suggestion
        # It would be ideal to test with both Release and Debug build types, but GPU runners cost per-minute;
        # to avoid excess runtime costs, we compromise to run with RelWithDebInfo build type, which runs fast
        # but still produces debugging checks.
        build-type: [ RelWithDebInfo ]
```

.github/workflows/build-and-run-gpu.yml
Tyler-g-hudson: My personal preference is to remove any steps that will not run unless we have a short- to medium-term expectation of using them, simply to avoid confusion when maintaining the file later on. If not, then a comment explaining the reasoning for including the step would be fine just so anyone who wants to read the file or use it for reference doesn't have to spend as much time figuring out if they need the step or not.

Tyler-g-hudson: @gmgunter I would be interested to hear if you have any thoughts or preferences here as well

PR #35, opened 2025-06-27
Compute bounding polygon for a geocoded grid (MCR #90132)

gmgunter:
Adds a function `make_geo_grid_bounding_polygon()` that returns a polygon in geodetic coordinates that approximately encloses an input `GeoGridParameters` object. The result is an `osgeo.ogr.Geometry` object with counterclockwise orientation.

This is useful for the NISAR Static Layers product. Unlike other L2 products whose bounding polygon represents only the footprint of the swath inside the output geo grid, the bounding polygon should span the entire geo grid for Static Layers.

### DIFF
--- a/python/packages/isce3/geometry/bounding_polygon.py
+++ b/python/packages/isce3/geometry/bounding_polygon.py
@@ -1,12 +1,124 @@
 from __future__ import annotations
 
-from collections.abc import Iterator
+from collections.abc import Iterable, Iterator
 
 import numpy as np
+import shapely
+from numpy.typing import ArrayLike
 from osgeo import ogr
 
 import isce3
 from isce3.geometry import DEMInterpolator
+
+Point2D = tuple[float, float]
+Point3D = tuple[float, float, float]
+
+
+def wrap_to_interval(val: float, lower: float, upper: float) -> float:
+    """
+    Wrap a value to the range [lower, upper) using modular arithmetic.
+
+    Parameters
+    ----------
+    val : float
+        Value to wrap.
+    lower : float
+        Lower endpoint of the interval (inclusive).
+    upper : float
+        Upper endpoint of the interval (exclusive). Must be > `lower`.
+
+    Returns
+    -------
+    float
+        Wrapped value in the range [lower, upper).
+    """
+    if not (upper > lower):
+        raise ValueError(f"{upper=} must be greater than {lower=}")
+
+    # Python's `%` (modulo) operator computes the remainder with the same sign as the
+    # right-hand side operand, so the result of the modulo operation below is in the
+    # range [0, b-a), where b is the upper endpoint and a is the lower endpoint. Then,
+    # if we add a, the result will be in the range [a, b). See
+    # https://docs.python.org/3/reference/expressions.html#binary-arithmetic-operations.
+    return (val - lower) % (upper - lower) + lower
+
+
+def unwrap_degrees(angles: ArrayLike) -> np.ndarray:
+    """
+    Unwrap a sequence of angles in degrees.
+
+    Unwraps the input angles such that the absolute differences between adjacent
+    elements are never greater than 180 degrees by adding a multiple of 360 degrees to
+    each element.
+
+    Parameters
+    ----------
+    angles : array_like
+        The input sequence of angles, in degrees.
+
+    Returns
+    -------
+    numpy.ndarray
+        The unwrapped angles, in degrees.
+    """
+    # `numpy.unwrap` doesn't correctly unwrap inputs in degrees prior to NumPy 1.21.0.
+    # See https://github.com/numpy/numpy/pull/16987.
+    # ISCE3 currently supports NumPy >=1.20.
+    return np.rad2deg(np.unwrap(np.deg2rad(angles)))
+
+
+def unwrap_longitudes(llhs: Iterable[Point3D]) -> Iterator[Point3D]:
+    """
+    Unwrap longitude coordinates of a sequence of LLH points.
+
+    Parameters
+    ----------
+    llhs : iterable of (float, float, float)
+        An ordered series of (longitude, latitude, height) triplets, with longitudes
+        specified in degrees.
+
+    Returns
+    -------
+    iterator of (float, float, float)
+        An iterator over the input sequence with longitude values unwrapped. The order
+        of the points is preserved.
+    """
+    lons, lats, heights = map(list, zip(*llhs))
+    return zip(unwrap_degrees(lons), lats, heights)
+
+
+def create_ogr_polygon(
+    exterior_vertices: Iterable[Point2D] | Iterable[Point3D],
+) -> ogr.Geometry:
+    """
+    Create a polygon with the specified exterior vertices as an OGR Geometry object.
+
+    Parameters
+    ----------
+    exterior_vertices : iterable of (float, float) or iterable of (float, float, float)
+        The vertices of the exterior boundary of the polygon. An ordered series of 2D or
+        3D points. The final point should not be identical to the first -- the boundary
+        ring will be automatically closed.
+
+    Returns
+    -------
+    osgeo.ogr.Geometry
+        The output polygon.
+    """
+    # Create the exterior ring of the polygon.
+    ring = ogr.Geometry(ogr.wkbLinearRing)
+    for vertex in exterior_vertices:
+        ring.AddPoint(*vertex)
+
+    # Ensure the ring is closed by appending the first vertex to the end of the list of
+    # points in the ring.
+    ring.CloseRings()
+
+    # Create the output polygon object.
+    polygon = ogr.Geometry(ogr.wkbPolygon)
+    polygon.AddGeometry(ring)
+
+    return polygon
 
 
 def make_geo_grid_bounding_polygon(
@@ -19,16 +131,17 @@ def make_geo_grid_bounding_polygon(
     Get the perimeter of a geocoded grid as an LLH polygon.
 
     Construct a polygon whose corners consist of (longitude, latitude, height) points
-    sampled along the perimeter of `geo_grid`. The polygon has counterclockwise
-    orientation (in the native coordinates of `geo_grid`), starting from the upper-left
-    corner of the grid.
+    sampled along the perimeter of `geo_grid`.
+
+    The polygon is convex with counterclockwise winding order on the ellipsoid. The
+    first boundary vertex is the upper-left corner of the grid. The polygon is closed
+    (i.e. the first and last point in the boundary are the same).
 
     In the resulting polygon, longitude and latitude coordinates are in degrees and
-    height is in meters above the reference ellipsoid of the digital elevation model
-    (DEM).
-
-    The polygon is closed (i.e. the first and last point in the polygon perimeter ring
-    are the same).
+    height is in meters above the reference ellipsoid of the input digital elevation
+    model (DEM). Longitude values are in the range [-180, 180], except that they may
+    extend outside this range in the case where `geo_grid` spans the antimeridian.
+    Latitude values are in the range [-90, 90].
 
     Parameters
     ----------
@@ -58,6 +171,32 @@ def make_geo_grid_bounding_polygon(
     # `geo_grid`.
     proj = isce3.core.make_projection(geo_grid.epsg)
 
+    # Convert an (x, y) point in projected coordinates to an LLH point on the DEM
+    # surface.
+    def xy_to_llh(x: float, y: float) -> Point3D:
+        # Transform from projected coordinates to geodetic coordinates.
+        # XXX: `ProjectionBase.inverse()` doesn't necessarily guarantee that the
+        # resulting longitude is in [-pi, pi] or that the resulting latitude is in
+        # [-pi/2, pi/2].
+        lon_rad, lat_rad, _ = proj.inverse((x, y, 0.0))
+        lon_rad = wrap_to_interval(lon_rad, lower=-np.pi, upper=np.pi)
+        if not (-0.5 * np.pi <= lat_rad <= 0.5 * np.pi):
+            raise ValueError(
+                "geo_grid latitude coordinates must be within the interval"
+                f" [-pi/2, pi/2], got latitude={lat_rad} (radians)"
+            )
+
+        # Interpolate the DEM height. In general, the DEM grid may be in a different
+        # coordinate system than the input (x, y) points, so use `interpolate_lonlat()`,
+        # which internally converts (lon, lat) points to the native DEM coordinates,
+        # instead of `interpolate_xy()`.
+        height = dem.interpolate_lonlat(lon_rad, lat_rad)
+
+        # Convert to degrees.
+        lon_deg, lat_deg = np.rad2deg((lon_rad, lat_rad))
+
+        return lon_deg, lat_deg, height
+
     # The transformation from projected coordinates to LLH coordinates is not affine in
     # general, so it's not sufficient to just use the four corners of `geo_grid` as the
     # corners of the bounding polygon -- we need to ensure that the entire perimeter of
@@ -67,7 +206,7 @@ def make_geo_grid_bounding_polygon(
 
     # Yield (x, y) points sampled uniformly along the perimeter of `geo_grid` in native
     # coordinates, in counterclockwise order, starting from the upper-left corner.
-    def boundary_pts() -> Iterator[tuple[float, float]]:
+    def boundary_pts() -> Iterator[Point2D]:
         # Get x & y coordinates uniformly sampled along the extents of the `geo_grid`.
         xcoords = np.linspace(geo_grid.start_x, geo_grid.end_x, num=pts_per_edge)
         ycoords = np.linspace(geo_grid.start_y, geo_grid.end_y, num=pts_per_edge)
@@ -92,20 +231,31 @@ def make_geo_grid_bounding_polygon(
         for x in xcoords[1:-1][::-1]:
             yield x, y0
 
-    # Create a ring of LLH points sampled along the perimeter of `geo_grid`.
-    ring = ogr.Geometry(ogr.wkbLinearRing)
-    for x, y in boundary_pts():
-        lon_rad, lat_rad, _ = proj.inverse((x, y, 0.0))
-        height = dem.interpolate_lonlat(lon_rad, lat_rad)
-        lon_deg, lat_deg = np.rad2deg((lon_rad, lat_rad))
-        ring.AddPoint(lon_deg, lat_deg, height)
+    # Get points sampled along the perimeter of `geo_grid` in LLH coordinates. Unwrap
+    # longitude values to handle antimeridian crossings.
+    perimeter_points_llh = (xy_to_llh(x, y) for (x, y) in boundary_pts())
+    perimeter_points_llh = list(unwrap_longitudes(perimeter_points_llh))
 
-    # Ensure the ring is closed by appending the first point to the end of the list of
-    # points in the ring.
-    ring.CloseRings()
+    # Construct a `LinearRing` from the perimeter points so we can ensure that the
+    # output polygon is simple (i.e. not self-intersecting) and has the correct winding
+    # order.
+    ring = shapely.LinearRing(perimeter_points_llh)
 
-    # Create the output polygon object.
-    polygon = ogr.Geometry(ogr.wkbPolygon)
-    polygon.AddGeometry(ring)
+    # Check that the exterior of the polygon is a valid LinearRing (that is, it does not
+    # cross or touch itself). This could happen, for example, if `geo_grid` extends
+    # beyond +/-90 degrees latitude, which might result in a self-intersecting shape on
+    # the ellipsoid (see
+    # https://github.com/isce-framework/isce3/pull/35#discussion_r2178276599).
+    if not ring.is_valid:
+        raise RuntimeError("polygon boundary is invalid")
 
-    return polygon
+    # Ensure that the polygon has counterclockwise (CCW) winding order on the ellipsoid.
+    # Reverse the order of points along the exterior boundary if necessary. This could
+    # happen, for example, if the geo grid orientation is not north-up/west-left, or if
+    # the projection doesn't preserve the north-south ordering and east-west ordering of
+    # the points.
+    if not ring.is_ccw:
+        perimeter_points_llh = perimeter_points_llh[::-1]
+        assert shapely.LinearRing(perimeter_points_llh).is_ccw
+
+    return create_ogr_polygon(perimeter_points_llh)
--- a/python/packages/isce3/geometry/bounding_polygon.py
+++ b/python/packages/isce3/geometry/bounding_polygon.py
@@ -1,0 +1,111 @@
+from __future__ import annotations
+
+from collections.abc import Iterator
+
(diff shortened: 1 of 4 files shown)


gshiroma (commented):
Nice PR, @gmgunter . I really like how you implemented this functionality and the way you're testing it. I have just one concern with this PR: the assumption that a counter-clockwise (CCW) winding in projected coordinates will be preserved when converting the polygon to geographic coordinates (lat/lon). While that might hold for UTM, I don’t think we can safely assume it for polar stereographic projections like EPSG:3031 or EPSG:3413. Even though the unit test using the dem_south_pole.tif in EPSG:3031 passes, I’m not confident this will always be the case. Please, let me know if I'm missing something. There are ways we can adapt your code to always produce CCW in geographic coordinates, but we also need to be careful with the antimeridian crossing.

gmgunter:
Thanks, @gshiroma. Good points.

I think as long as the projection preserves the west-east ordering and north-south ordering of the points (and doesn't transpose the coordinate axes), then a CCW ring in projected coordinates will also be CCW in geodetic coordinates. But, since it's possible to implement the function to ensure that the ring is always CCW in LLH coordinates, it seems it's not necessary to make this assumption (and I should probably document the assumption otherwise).

>There are ways we can adapt your code to always produce CCW in geographic coordinates

One way I can imagine to do this is to first construct the ring, then check if it has CCW orientation, and then, if not, construct a new ring by adding the points in reverse order.

Can you think of a better approach?

>we also need to be careful with the antimeridian crossing

Good note. I can add a test case with an antimeridian crossing as well.

gshiroma:
> I think as long as the projection preserves the west-east ordering and north-south ordering of the points

Hi @gmgunter , I don't think polar projections preserve west-east ordering and north-south ordering of the points. I should be able to create a plot to exemplify this and I'll follow up once I have it.

> One way I can imagine to do this is to first construct the ring, then check if it has CCW orientation, and then, if not, construct a new ring by adding the points in reverse order.

This seems okay. But I'm not sure how you'll test for CCW orientation when the longitude values "wrap" around +/- 180. You may need to "unwrap" the longitude values first before doing the test.

gshiroma:
The plot below shows a hillshade of the Copernicus DEM projected in the NSIDC Sea Ice Polar Stereographic North projection (EPSG:3413). The northernmost point (North Pole) is at the center of the plot.

In regions near Alaska, the geographic North points to the right, while in areas near Northern Europe, it points to the left. So, the west-east and north-south ordering of points should not be preserved in this projection.

However, as I was creating this plot, I started to wonder: even if the linear ordering isn’t preserved, maybe the clockwise (CW) and counter-clockwise (CCW) orientation are still maintained. Someone rotating CW or CCW in any of those regions would still appear to rotate in the same sense when viewed in geographic coordinates.

So I guess my initial concern is not really an issue and we likely don’t need to worry about CW/CCW inversion when converting from projected to geographic coordinates. Let me think a bit more about it and I'll follow up soon.

Screenshot 2025-06-30 at 12 01 50 PM

gshiroma (approved):
It appears that CW/CCW orientation is preserved when converting between polar stereographic and geographic coordinates (more info below). The only caveats are polygons that either cover the pole or cross the antimeridian (when longitude values are wrapped).

The former case is a bit odd. It's unclear whether CW/CCW orientation is even well-defined in that case. The latter, however, should be correctly handled by your PR, since maintaining CCW ordering is primarily intended to help identify antimeridian crossings (even if longitude values are "wrapped").

To test if CW/CCW ordering is preserved, I've added some clockwise arrows to DEMs in polar stereographic projection, as shown below:

Screenshot 2025-06-30 at 12 27 57 PM

Screenshot 2025-06-30 at 12 29 54 PM

Then, I geocoded those modified DEMs into geographic coordinates:

Screenshot 2025-06-30 at 12 32 17 PM

As you can see. The modified DEM in geographic coordinates has all the arrows in clockwise order, which matches the input DEMs and suggests that the order is preserved! Sorry, for the noise, but now I'm more confident that the PR is doing the right thing and I'm approving it. Thanks, @gmgunter !

python/packages/isce3/geometry/bounding_polygon.py
```diff
@@ -0,0 +1,111 @@
...
+    # `geo_grid` is contained within the polygon. As a concession to computational
+    # feasibility, we'll sample a finite number of points along each edge of `geo_grid`
+    # to form the perimeter of the output polygon.
+
+    # Yield (x, y) points sampled uniformly along the perimeter of `geo_grid` in native
+    # coordinates, in counterclockwise order, starting from the upper-left corner.
+    def boundary_pts() -> Iterator[tuple[float, float]]:
+        # Get x & y coordinates uniformly sampled along the extents of the `geo_grid`.
+        xcoords = np.linspace(geo_grid.start_x, geo_grid.end_x, num=pts_per_edge)
+        ycoords = np.linspace(geo_grid.start_y, geo_grid.end_y, num=pts_per_edge)
```
bhawkins, line 212: This is probably fine for the intended use case but isn't obviously correct in general:
1. The CCW ordering only occurs when dx > 0 and dy < 0 (or dx < 0 and dy > 0).
2. The ctor `linspace(start, end, n)` doesn't work correctly if either coordinate wraps.

While it may be correct given the particular implementation of `GeoGridParameters`, checking signs and using `start + step * arange(length)` would be easier to understand.

python/packages/isce3/geometry/bounding_polygon.py
```diff
@@ -0,0 +1,111 @@
...
+
+        # Top edge from right to left (excluding both endpoints).
+        y0 = ycoords[0]
+        for x in xcoords[1:-1][::-1]:
+            yield x, y0
+
+    # Create a ring of LLH points sampled along the perimeter of `geo_grid`.
+    ring = ogr.Geometry(ogr.wkbLinearRing)
+    for x, y in boundary_pts():
+        lon_rad, lat_rad, _ = proj.inverse((x, y, 0.0))
```
bhawkins: It might be worth mentioning in the docstring that the polygon is CCW on the sphere, not necessarily on a lat/lon projection. For example, the lat/lon polygon for a frame that crosses the dateline or contains the pole would fail the `shapely.is_ccw` check despite being correctly ordered on the sphere.

bhawkins (commented):
This is a nice feature. I just have a couple of nits based on experience with the RSLC polygons.

python/packages/isce3/geometry/bounding_polygon.py
gmgunter: >The CCW ordering only occurs when dx > 0 and dy < 0 (or dx < 0 and dy > 0).

I was under the impression that `GeoGridParameters` only supports north-up, west-left grids (or, more specifically, that it only supports grids for which dx > 0 and dy < 0). It looks like that's only half true -- the y-spacing must be negative-valued but the class doesn't necessarily enforce any constraints on the x-spacing.

https://github.com/isce-framework/isce3/blob/968e78fb90e09fd69d34770fc6ab238ac2b0fdcb/cxx/isce3/product/GeoGridParameters.cpp#L32-L37

I'd prefer to resolve this by clarifying what the invariants of `GeoGridParameters` are. Do we want that class to only support dx > 0 and dy < 0? If so, I could just leave a comment and a few `assert` statements in my code like this:

```python
    # Other orientations are impossible due to the constraints imposed by 
    # `GeoGridParameters`.
    assert geo_grid.spacing_x > 0.0
    assert geo_grid.spacing_y < 0.0
```

Otherwise, I'm fine with making the implementation of this PR more general so that it can support grids with different orientations.

>The ctor `linspace(start, end, n)` doesn't work correctly if either coordinate wraps.
>
>While it may be correct given the particular implementation of `GeoGridParameters`, checking signs and using `start + step * arange(length)` would be easier to understand.

How should I get `step`, though? The obvious way is

```python
step = (geo_grid.end_x - geo_grid.start_x) / (pts_per_edge - 1)
```

but that approach is similarly vulnerable to coordinate wrapping. (And I think computing `step` that way followed by `start + step * arange(length)` is basically equivalent to the linspace approach, ignoring floating point rounding errors.)

Did you have in mind a better way to compute `step`? Would you be more satisfied with something like this:

```python
# `geo_grid.end_x` might be unreliable due to coordinate wrapping.
end_x_unwrapped = geo_grid.start_x + geo_grid.spacing_x * geo_grid.width

x_coords = np.linspace(geo_grid.start_x, end_x_unwrapped, num=pts_per_edge)
```

?

I may have to go back and audit my previous uses of `GeoGridParameters.end_{x,y}` if it's not safe to assume that the x & y coordinates are strictly increasing / strictly decreasing. It seems to me that others might make similar assumptions. Maybe as part of this effort we could try to better document what the invariants of `GeoGridParameters` are/aren't.

bhawkins: > I'd prefer to resolve this by clarifying what the invariants of GeoGridParameters are.

I was kinda just hoping to be able to review this PR without really understanding GeoGridParameters at all, at the expense of a couple of extra lines of code. Those asserts fit the bill, though they make me wonder whether spacing=0 is okay (not okay if you want a well-defined winding order but occurs with default-constructed `GeoGridParameters` objects). It'd be great to improve the docs in the C++ header and add enforcement in the ctor/setters, too.

> How should I get step, though? The obvious way is

Now I see the difficulty. I commented (half) thinking you should just use `geo_grid.spacing_x`, but you need to support `pts_per_edge != geo_grid.length` so using the endpoint really is more convenient. The actual class behavior is just like in your example, so maybe asserts are warranted there as well.

While it's kind of obvious since the implementation is in the header, it'd be nice to clarify in the API docs that `end_x >= start_x` and `end_y <= start_y` regardless of the projection. More concretely for EPSG:4326 that means lats and lons are not bounded/modular, though that seems pretty busted for latitude...

gmgunter: Oof yeah I forgot about the default constructor. That makes it more difficult to define the strict invariants that I want. OK, I'll abandon that fantasy for now and add the couple of extra lines of code here to make the code robust to different grid orientations.

>While it's kind of obvious since the implementation is in the header, it'd be nice to clarify in the API docs that `end_x >= start_x` and `end_y <= start_y` regardless of the projection. More concretely for EPSG:4326 that means lats and lons are not bounded/modular, though that seems pretty busted for latitude...

I'll propose some verbiage in a separate PR.

I'm not sure it's sufficient to document that `end_x >= start_x`. For example, you could have a grid with a north pole crossing, where latitude values are initially increasing and then decreasing. In that case, `end_x` might be > `start_x`, but the maximum latitude value could still be > `end_x`.

The specific behavior I'd like to document is that the coordinate spacing is fixed and `end_x` is equivalent to `start_x + spacing_x * width`. Without that guarantee, I don't think anyone can safely use `GeoGridParameters.end_x` without jumping through a bunch of hoops.

bhawkins: > though that seems pretty busted for latitude...

Elaborating on this a bit more, it might be smart to have the `GeoGridParameters` constructor throw an error if it detects `abs(startY) > 90 or abs(endY) > 90` in the EPSG:4326 case. GDAL coordinate transforms fail for $|\phi| > 90^\circ$. ISCE3 coordinate transforms allow it, treating the point $(\lambda, 90^\circ + \delta)$ as equivalent to $(-\lambda, 90^\circ - \delta)$. However, for a rectangular `GeoGridParameters` that implies a self-intersecting shape on the sphere as shown below:

polar_poly_small

That seems like an erroneous usage to me. It also leads to an inconsistent winding order of the polygon on the sphere.

In case it's helpful, here's the code to generate that polygon:

```Python
import isce3
from osgeo import ogr, osr

grid = isce3.product.GeoGridParameters(
    start_x = -118.0,
    start_y = -88.0,
    spacing_x = 1.0,
    spacing_y = -1.0,
    width = 10,
    length = 5,
    epsg = 4326)

poly = isce3.geometry.bounding_polygon.make_geo_grid_bounding_polygon(grid)
print(poly.ExportToWkt())

srs_lonlat = osr.SpatialReference()
srs_lonlat.ImportFromEPSG(4326)
srs_lonlat.SetAxisMappingStrategy(osr.OAMS_TRADITIONAL_GIS_ORDER)

poly.AssignSpatialReference(srs_lonlat)

srs_polar = osr.SpatialReference()
srs_polar.ImportFromEPSG(3031)

try:
    poly.Transform(osr.CoordinateTransformation(srs_lonlat, srs_polar))
    print(poly.ExportToWkt())
except:
    print("GDAL failed to transform polygon with lat>90 to polar stereo")

# isce3 to the "rescue"
proj_lonlat = isce3.core.make_projection(4326)
proj_polar = isce3.core.make_projection(3031)
points_polar = [proj_polar.forward(proj_lonlat.inverse(pt))
    for pt in poly.GetBoundary().GetPoints()]

ring = ogr.Geometry(ogr.wkbLinearRing)
for x, y, z in points_polar:
    ring.AddPoint(x, y, z)
poly_polar = ogr.Geometry(ogr.wkbPolygon)
poly_polar.AddGeometry(ring)
print(poly_polar.ExportToWkt())

```

python/packages/isce3/geometry/bounding_polygon.py
gshiroma: @bhawkins , it's not uncommon to see geogrids exceeding  +/- 90 degrees lat. See for example this geoid model (EGM08) provided by OSGeo [here](https://github.com/OSGeo/proj-datumgrid/blob/master/world/egm08_25.gtx) with latitude extents equal to `[-90.02083333333334,  90.02083333333331]`. The same applies to the Copernicus-based DEM we prepared to use for the procesing of NISAR products, whose latitude extents are `[-89.99986111111112,  90.00013888888888]`.

bhawkins: (I think you meant to reply to my other comment.) I hadn't really considered the case of a global dataset, where I can see the value in having a small amount of double-coverage. Thanks for pointing that out.

gshiroma: Yes, it was related to the other comment =). Thank you, @bhawkins .

gmgunter:
Thanks for all your feedback, @gshiroma and @bhawkins. This task ended up being much more prickly than I'd imagined!

Summarizing the concerns that were brought up in the above discussion:

* In general, the polygon may not necessarily have CCW winding order in geodetic coordinates even if it has CCW winding order in projected coordinates (though this isn't an issue for the UTM or UPS projections used for NISAR).
* The function may not behave as documented in the case where the input geo grid crosses the antimeridian.
* The polygon may not have CCW winding order if the geo grid x-spacing was negative xor the y-spacing was positive.
* The polygon may be self-intersecting if the latitude coordinates of the geo grid extended beyond the range [-90, 90]. In this case, the polygon clearly does not have CCW winding order.

I've made the following changes in 620fc14d7897a35f3356b46b4115e6b664cb5c1a to try to address these issues:

* Ensure the longitude coordinates computed by `proj.inverse()` are wrapped to the interval [-180, 180], except in the case of an antimeridian crossing. Unwrap longitude values so that the polygon is still convex with CCW winding order in case of an antimeridian crossing.
* Only allow latitude values in the interval [-90, 90]. Raise an exception if latitude values outside this range are encountered.
* Ensure that the polygon has CCW winding order in geodetic coordinates, even if the geo grid was not north-up/west-left and/or the projection did not preserve north-south and east-west relative ordering. (This is done by constructing the polygon boundary in LLH coordinates, checking if it's counterclockwise, and, if not, reversing the order of the points.)

I've also added the following to the test suite to exercise the new behavior:

* A test case with a geo grid that crosses the antimeridian.
* A test case with a geo grid with north-down orientation (i.e. positive-valued y-spacing).
* A test that ensures the function raises an exception if the latitude coordinates of the geo grid extend outside the interval [-90, 90].

gshiroma:
> Thanks for all your feedback, @gshiroma and @bhawkins. This task ended up being much more prickly than I'd imagined!
> 
> Summarizing the concerns that were brought up in the above discussion:
> 
> * In general, the polygon may not necessarily have CCW winding order in geodetic coordinates even if it has CCW winding order in projected coordinates (though this isn't an issue for the UTM or UPS projections used for NISAR).
> * The function may not behave as documented in the case where the input geo grid crosses the antimeridian.
> * The polygon may not have CCW winding order if the geo grid x-spacing was negative xor the y-spacing was positive.
> * The polygon may be self-intersecting if the latitude coordinates of the geo grid extended beyond the range [-90, 90]. In this case, the polygon clearly does not have CCW winding order.
> 
> I've made the following changes in [620fc14](https://github.com/isce-framework/isce3/commit/620fc14d7897a35f3356b46b4115e6b664cb5c1a) to try to address these issues:
> 
> * Ensure the longitude coordinates computed by `proj.inverse()` are wrapped to the interval [-180, 180], except in the case of an antimeridian crossing. Unwrap longitude values so that the polygon is still convex with CCW winding order in case of an antimeridian crossing.
> * Only allow latitude values in the interval [-90, 90]. Raise an exception if latitude values outside this range are encountered.
> * Ensure that the polygon has CCW winding order in geodetic coordinates, even if the geo grid was not north-up/west-left and/or the projection did not preserve north-south and east-west relative ordering. (This is done by constructing the polygon boundary in LLH coordinates, checking if it's counterclockwise, and, if not, reversing the order of the points.)
> 
> I've also added the following to the test suite to exercise the new behavior:
> 
> * A test case with a geo grid that crosses the antimeridian.
> * A test case with a geo grid with north-down orientation (i.e. positive-valued y-spacing).
> * A test that ensures the function raises an exception if the latitude coordinates of the geo grid extend outside the interval [-90, 90].

Thank you, @gmgunter , for addressing the PR comments. I just reviewed the updated code and unit tests, and everything looks great. One thing to keep in mind is that `make_geo_grid_bounding_polygon()` returns an OGR Polygon with unwrapped longitude values, but I believe the polygon saved in NISAR products should be wrapped within the [-180, 180] interval. @bhawkins can confirm.

python/packages/isce3/geometry/bounding_polygon.py
```diff
@@ -0,0 +1,261 @@
...
+    # order.
+    ring = shapely.LinearRing(perimeter_points_llh)
+
+    # Check that the exterior of the polygon is a valid LinearRing (that is, it does not
+    # cross or touch itself). This could happen, for example, if `geo_grid` extends
+    # beyond +/-90 degrees latitude, which might result in a self-intersecting shape on
+    # the ellipsoid (see
+    # https://github.com/isce-framework/isce3/pull/35#discussion_r2178276599).
+    if not ring.is_valid:
+        raise RuntimeError("polygon boundary is invalid")
```
bhawkins, line 250: The "south_pole" test case doesn't contain the actual pole (-90 deg latitude). A grid that does contain the pole triggers this `RuntimeError`. For example,
```Python
def south_pole():
    grid = GeoGridParameters(
        start_x=-100e3,
        start_y=100e3,
        spacing_x=10e3,
        spacing_y=-10e3,
        width=21,
        length=21,
        epsg=3031,
    )
    dem = DEMInterpolator()
    return grid, dem
```

It would be possible to work around this case by cutting at the dateline. Since that wouldn't cut all the way through the polygon, it'd still be a simple polygon (not a multipolygon). It would just have three extra edges (one at lon=+180°, one at lon=-180°, and one at lat=-90°).

python/packages/isce3/geometry/bounding_polygon.py
```diff
@@ -0,0 +1,261 @@
...
+    sampled along the perimeter of `geo_grid`.
+
+    The polygon is convex with counterclockwise winding order on the ellipsoid. The
+    first boundary vertex is the upper-left corner of the grid. The polygon is closed
+    (i.e. the first and last point in the boundary are the same).
+
+    In the resulting polygon, longitude and latitude coordinates are in degrees and
+    height is in meters above the reference ellipsoid of the input digital elevation
+    model (DEM). Longitude values are in the range [-180, 180], except that they may
+    extend outside this range in the case where `geo_grid` spans the antimeridian.
```
bhawkins, line 143: This is a nice addition, though I expect folks will continue to encounter issues. For example, I tried dumping one of the unit tests to KML, but the `ExportToKML()` method wraps the values back to [-180, 180] deg. Then I dumped the WKT (which didn't re-wrap) and loaded it in QGIS using a plugin. But my basemap was in web mercator instead of EPSG:4326, so it wrapped the wrong way around again.

In principle we could cut these at the dateline, resulting in a multi-polygon. But I think it makes sense to leave this to users rather than randomly spit out a different topology.

bhawkins (approved):
Thanks for the improvements! There are a few things worth pointing out, but I don't consider them blockers.

python/packages/isce3/geometry/bounding_polygon.py
gmgunter: >The "south_pole" test case doesn't contain the actual pole (-90 deg latitude).

Sorry for the confusing name! I initially just had two test cases, "winnipeg" and "south_pole", and decided to name them after the two pre-existing DEM raster files in the `tests/data` directory: "winnipeg_dem.tif" and "dem_south_pole.tif".

The intention was just to test using a variety of projections -- not to test with a geo grid containing the pole. I'll rename this test case from "south_pole" to "antarctica" to reduce confusion.

>It would be possible to work around this case by cutting at the dateline. Since that wouldn't cut all the way through the polygon, it'd still be a simple polygon (not a multipolygon). It would just have three extra edges (one at lon=+180°, one at lon=-180°, and one at lat=-90°).

This is a clever approach, but I think it might be best to maintain the invariant that the bounding polygon has 4 edges (in the geo grid's native coordinates) with an equal number of points along each. The QA code [relies](https://github.com/isce-framework/nisarqa/blob/cd16089ae436be7815c4a95d00d36351858b658b/src/nisarqa/products/product_reader.py#L1533-L1539) on this property in order to extract the [LatLonQuad](https://developers.google.com/kml/documentation/kmlreference#gxlatlonquad) that's used to form its output KML file. Currently there's no QA for Static Layers, but I'm not sure how we would handle this if we decide to add Static Layers QA in the future.

We could revisit this later on if/when we need to add support for geo grids that contain a pole.

python/packages/isce3/geometry/bounding_polygon.py
gmgunter: >For example, I tried dumping one of the unit tests to KML, but the `ExportToKML()` method wraps the values back to [-180, 180] deg.

That's a shame -- my initial inspiration for unwrapping longitude values came from the KML spec for [`LatLonBox`](https://developers.google.com/kml/documentation/kmlreference#latlonbox):

> * **\<east\>** Specifies the longitude of the east edge of the bounding box, in decimal degrees from 0 to ±180. (For overlays that overlap the meridian of 180° longitude, values can extend beyond that range.)
> * **\<west\>** Specifies the longitude of the west edge of the bounding box, in decimal degrees from 0 to ±180. (For overlays that overlap the meridian of 180° longitude, values can extend beyond that range.)

As long as there's no explicit requirement that the longitude coordinates of these bounding boxes must be wrapped to [-180, 180], I think my slight preference would be to continue to unwrap cases that straddle the antimeridian. At least it makes it much more convenient to check that the polygon has CCW winding order in the unit test suite. But it'd be no big deal to rewrap these longitude coordinates if that's the behavior that we'd prefer.

gmgunter:
(Rebased to fix some merge conflicts)

PR #37, opened 2025-06-30
Add option to compute the ground velocity using rdr2geo (geolocation grid cubes)

gshiroma:
This PR adds option to compute the ground velocity using rdr2geo for geolocation grid cubes (L1 products) in addition to the current method that uses a theoretical formula. The new method is added to the function `makeGeolocationGridCubes()` selected through the flag `flag_ground_velocity_from_rdr2geo`. Since the new method is believed to be more accurate, this PR also makes the rdr2geo the default method for computing the ground-track velocity for geolocation grid cubes.

### DIFF
--- a/cxx/isce3/geometry/metadataCubes.cpp
+++ b/cxx/isce3/geometry/metadataCubes.cpp
@@ -719,7 +719,7 @@ void makeGeolocationGridCubes(
                 info << "estimating the ground-track velocity using rdr2geo" << pyre::journal::endl;
             }
 
-        _Pragma("omp parallel for")
+            _Pragma("omp parallel for")
             /* 
             Compute ground-track velocity based on estimated distance between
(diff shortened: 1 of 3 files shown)


hfattahi (approved):
LGTM. Thank you @gshiroma . We have reviewed this PR in the internal repository before the public repository was open for PRs and all comments have been addressed. As part of the PR review I evaluated the along track spacing computations from the metadata cube with two different approaches. 1) by evaluating the ground velocity cube and multiplying by azimuth time interval, and 2) by evaluating the x and y coordinates cube in L1 product for two points on ground whose distance in radar coordinates is one PRF.
The outcome is that the two are consistent:
```
Along track spacing from groound velocity cube: 22.420074494376827 meters
Along track spacing derived from x, y coordinates in the cube: 22.4200743822321 meters
```

bhawkins (approved):
Reviewed internally

PR #46, opened 2025-06-30
Add option to GCOV to use the RSLC radiometric calibration LUTs

gshiroma:
This PR adds an option to the GCOV workflow to apply radiometric calibration using the RSLC radiometric calibration LUTs `beta0`, `sigma0`, or `gamma0` located in the input RSLC metadata `science/LSAR/RSLC/metadata/calibrationInformation/geometry/{beta0|sigma0|gamma0}`. The radiometric calibration is applied to the RSLC before RTC. Therefore, if RTC is enabled (`apply_rtc` is `True`), this field should be consistent with the RTC group field `input_terrain_radiometry`. Otherwise, the software will raise an error. The value "gamma0" implies that the radiometric calibration will be applied instead of RTC. In that case, the field `apply_rtc` should be `False`, otherwise the software will also raise an error.


Nafeazizi:
مرحبا استاذ هل تستطيع ارشادي لطريقة معالجة داتا نيسار مستوى 1 لاكمل بالطريقة التي اريد هل يوجد موقع سحابي او برنامج او حزم لويندوز

PR #49, opened 2025-06-30
Patch `cheby_equi_ripple_filter` for SciPy v1.16.0

### LINKED ISSUE #47: The unit test `test_build_multi_rate_fir_filter` is failing
The unit test `test_build_multi_rate_fir_filter` is currently failing, likely due to a change in SciPy’s `remez()` function, which is invoked by ISCE3 [here](https://github.com/isce-framework/isce3/blob/968e78fb90e09fd69d34770fc6ab238ac2b0fdcb/python/packages/isce3/signal/fir_filter_func.py#L79). Below is the CI error message:

```
../../../../../../tests/python/packages/isce3/signal/fir_filter_func.py:121: 
../../../../../../install/packages/isce3/signal/fir_filter_func.py:395: in build_multi_rate_fir_filter
    coefs = cheby_equi_ripple_filter(
../../../../../../install/packages/isce3/signal/fir_filter_func.py:79: in cheby_equi_ripple_filter
    coeffs = spsg.remez(len_flt, 0.5 / samprate

numtaps = 111, bands = array([0.        , 0.08333333, 0.1       , 0.5       ])
desired = array([1., 0.])

    def remez(numtaps, bands, desired, *, weight=None, type='bandpass',
              maxiter=25, grid_density=16, fs=None):
        """
        Calculate the minimax optimal filter using the Remez exchange algorithm.
    
        Calculate the filter-coefficients for the finite impulse response
        (FIR) filter whose transfer function minimizes the maximum error
        between the desired gain and the realized gain in the specified
        frequency bands using the Remez exchange algorithm.
    
        Parameters
        ----------
        numtaps : int
            The desired number of taps in the filter. The number of taps is
            the number of terms in the filter, or the filter order plus one.
        bands : array_like
            A monotonic sequence containing the band edges.
            All elements must be non-negative and less than half the sampling
            frequency as given by `fs`.
        desired : array_like
            A sequence half the size of bands containing the desired gain
            in each of the specified bands.
        weight : array_like, optional
        This example shows a band-pass filter with a pass-band from 2 kHz to
        5 kHz.  The transition width is 260 Hz and the length of the filter
        is 63, which is smaller than in the other examples:
    
        >>> band = [2000, 5000]  # Desired pass band, Hz
        >>> trans_width = 260    # Width of transition from pass to stop, Hz
        >>> numtaps = 63         # Size of the FIR filter.
        >>> edges = [0, band[0] - trans_width, band[0], band[1],
        ...          band[1] + trans_width, 0.5*fs]
        >>> taps = signal.remez(numtaps, edges, [0, 1, 0], fs=fs)
        >>> w, h = signal.freqz(taps, [1], worN=2000, fs=fs)
        >>> plot_response(w, h, "Band-pass Filter")
        >>> plt.show()
    
        The low order leads to higher ripple and less steep transitions.
    
        The next example shows a band-stop filter.
    
        >>> band = [6000, 8000]  # Desired stop band, Hz
        >>> trans_width = 200    # Width of transition from pass to stop, Hz
        >>> numtaps = 175        # Size of the FIR filter.
        >>> edges = [0, band[0] - trans_width, band[0], band[1],
        ...          band[1] + trans_width, 0.5*fs]
        >>> taps = signal.remez(numtaps, edges, [1, 0, 1], fs=fs)
        >>> w, h = signal.freqz(taps, [1], worN=2000, fs=fs)
        >>> plot_response(w, h, "Band-stop Filter")
        >>> plt.show()
    
        """
        xp = array_namespace(bands, desired, weight)
        bands = np.asarray(bands)
        desired = np.asarray(desired)
>       if weight:
           ^^^^^^
E       ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

/home/runner/micromamba/envs/isce3/lib/python3.12/site-packages/scipy/signal/_fir_filter_design.py:837: ValueError
____________ test_bandpass[0.4166666666666667--0.192[708](https://github.com/isce-framework/isce3/actions/runs/15982576518/job/45080120442?pr=39#step:11:709)33333333334] ____________
```

@rad-eng-59 , is this error familiar to you?

gmgunter:
Closes #47

SciPy 1.16.0 was released recently with [this update](https://github.com/scipy/scipy/commit/f45ba7bd311ee60c5c2f74c272cd3cd6c1dece86) that seems to break the `scipy.signal.remez()` function when the `weight` argument is not `None`. This seems to be causing CI failures for us.

Let's exclude this SciPy version from the environment spec.

### DIFF
--- a/python/packages/isce3/signal/fir_filter_func.py
+++ b/python/packages/isce3/signal/fir_filter_func.py
@@ -78,8 +78,8 @@ def cheby_equi_ripple_filter(samprate, bandwidth, rolloff=1.2, ripple=0.1,
     # get LPF coeffs
     coeffs = spsg.remez(len_flt, 0.5 / samprate
                         * np.array([0, bandwidth, fstop, samprate]),
-                        np.array([1.0, 0.0]),
-                        weight=np.array([1, weight_fact]),
+                        [1.0, 0.0],
+                        weight=[1, weight_fact],
                         fs=1, type='bandpass', maxiter=50)
 
     # up/down conversion
--- a/tests/python/packages/isce3/signal/fir_filter_func.py
+++ b/tests/python/packages/isce3/signal/fir_filter_func.py
@@ -245,6 +245,9 @@ def test_bandpass(width, fc):
                                       transition_width=tw, force_odd_len=odd)
     # compute frequency response over entire unit circle
     f, H = sig.freqz(h, fs=1.0, whole=True)
+    # Workaround for scipy=1.16.0
+    # (see https://github.com/isce-framework/isce3/pull/49#issuecomment-3025125318).
+    f = f.real
     a, b = transition_region(width, tw)
     # should have unit gain in passband with ripple < -att
     mask_pass = wrapped_distance(f - fc) < a


gmgunter:
I opened an upstream PR with a fix here: https://github.com/scipy/scipy/pull/23266

rad-eng-59 (approved):
LGTM. Thanks @gmgunter for the quick fix.
@gshiroma, The scipy folks has introduced a bug in line `836`. The problem is not on our side given the doc clearly says the optional input argument `weight` can be an array-like object. 
Alternatively, we can modify the `weight` to be a non-numpy container such as a python list rather than a numpy array on our side to avoid this scipy bug while still using the latest version of scipy. 
Let me know if you want to go with latter and I will open a PR to fix it.

rad-eng-59:
@gmgunter Alternatively, as stated earlier, you can simply remove `np.array` from [this line](https://github.com/isce-framework/isce3/blob/968e78fb90e09fd69d34770fc6ab238ac2b0fdcb/python/packages/isce3/signal/fir_filter_func.py#L82) to let `scipy 1.16.0` go through successfully. That is
```python
...
weight=[1, weight_fact],
...
```
No need to limit scipy version in isce3.

Perhaps, we should remove `np.array()` from [*desired*, the third argument to remez](https://github.com/isce-framework/isce3/blob/968e78fb90e09fd69d34770fc6ab238ac2b0fdcb/python/packages/isce3/signal/fir_filter_func.py#L81) as well to avoid similar issue in the future.

gmgunter:
Thanks, @rad-eng-59. I modified the PR to adopt your suggestion instead of excluding SciPy v1.16.0 from the environment spec.

rad-eng-59 (approved):
LGTM. Thank you so much @gmgunter and sorry for the trouble. 
I believe this is the second time the `remez` function being modified in SciPy :(

gshiroma (approved):
Thank you, @gmgunter !

gshiroma:
> LGTM. Thanks @gmgunter for the quick fix. @gshiroma, The scipy folks has introduced a bug in line `836`. The problem is not on our side given the doc clearly says the optional input argument `weight` can be an array-like object. Alternatively, we can modify the `weight` to be a non-numpy container such as a python list rather than a numpy array on our side to avoid this scipy bug while still using the latest version of scipy. Let me know if you want to go with latter and I will open a PR to fix it.

Sounds good. Thank you, @rad-eng-59 !

hfattahi (approved):
LGTM. Thanks @gmgunter.

gshiroma:
Uhm. Tests are still failing =/

gmgunter:
Looks like there's another bug in SciPy v1.16.0. The unit tests are now failing because the frequencies `f` computed by `scipy.signal.freqz` here:

https://github.com/isce-framework/isce3/blob/968e78fb90e09fd69d34770fc6ab238ac2b0fdcb/tests/python/packages/isce3/signal/fir_filter_func.py#L247

are coming out complex-valued now instead of real-valued.

Here's a snippet from the CI log:

```
x = array([0.36458333+0.j, 0.36653646+0.j, 0.36848958+0.j, 0.37044271+0.j,
       0.37239583+0.j, 0.37434896+0.j, 0.376302...1.35091146+0.j, 1.35286458+0.j, 1.35481771+0.j,
       1.35677083+0.j, 1.35872396+0.j, 1.36067708+0.j, 1.36263021+0.j])
fs = 1.0

    def wrapped_distance(x, fs=1.0):
>       xm = x % fs
             ^^^^^^
E       TypeError: ufunc 'remainder' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```

https://github.com/isce-framework/isce3/actions/runs/16004235062/job/45146789710?pr=49#step:11:558

We're getting a `TypeError` here because the `%` operator doesn't support complex values.

I'm not sure if there's a similarly simple bandaid we can apply here? Otherwise, I can revert back to excluding SciPy v1.16.0 from the environment spec.

rad-eng-59:
@gmgunter, please give me an hour or so to see if I can fix it and test it. Thank you

rad-eng-59:
It is odd that the frequency values are complex type with zero imaginary part. The frequency response shall be complex given it is a single-side band-pass filter and thus whole frequency range is required to cover the full filter response (given asymmetric filter response within +/- Nyquist `= fs / 2`). However, the SciPy code also sets the datatype of frequency values, `f within [0, 1) ` to complex. I would say it is safe to simply use real part of frequency bins to avoid this bug that is 
```python
f, H = sig.freqz(h, fs=1.0, whole=True)
f = f.real
```
I have just tested it with SciPy `v1.15.2`  and `v1.16.0`. The difference is in `f` data type. Basically the real part in new version  is the same as older version with extra zero-filled imaginary part which I have found it very very strange! So, let's simply use the real part which works for both old and new versions.

Btw, in other instances like [this one](https://github.com/isce-framework/isce3/blob/968e78fb90e09fd69d34770fc6ab238ac2b0fdcb/tests/python/packages/isce3/signal/fir_filter_func.py#L215), the code uses absolute value for the frequency bins provided by `signal.freqz` just to be cautious , I guess.

what do you say?

gmgunter:
Thanks @rad-eng-59 for investigating this. I've implemented the suggested patch and tested it locally with SciPy v1.13.1 and v1.16.0.

rad-eng-59 (approved):
LGTM. Hopefully this is the last change in `fir_filter_func.py`, at least for a while. 
Thank you @gmgunter for the fixes and apologies for all the trouble.
Thank you @gshiroma for reporting the issue in the first place and for your patience.

gmgunter:
Thanks @rad-eng-59 for all your help resolving this!

PR #50, opened 2025-07-01
Update attribute `grid_mapping_name` from unicode to byte string

gshiroma:
This PR updates attribute `grid_mapping_name` from unicode to byte string. The function `get_grid_mapping_name()` returns a string. This PR converts that string to byte array before saving it as an attribute of the projection dataset.

### DIFF
--- a/python/packages/nisar/workflows/h5_prep.py
+++ b/python/packages/nisar/workflows/h5_prep.py
@@ -797,7 +797,7 @@ def set_get_geo_info(hdf5_obj, root_ds, geo_grid, z_vect=None,
     sr = osr.SpatialReference()
     sr.ImportFromEPSG(epsg_code)
 
-    projds.attrs['grid_mapping_name'] = get_grid_mapping_name(sr)
+    projds.attrs['grid_mapping_name'] = np.bytes_(get_grid_mapping_name(sr))
 
     # Set up units
     # Geodetic latitude / longitude


bhawkins (approved):

python/packages/nisar/workflows/h5_prep.py
```diff
@@ -797,7 +797,7 @@ def set_get_geo_info(hdf5_obj, root_ds, geo_grid, z_vect=None,
     sr = osr.SpatialReference()
     sr.ImportFromEPSG(epsg_code)
 
-    projds.attrs['grid_mapping_name'] = get_grid_mapping_name(sr)
+    projds.attrs['grid_mapping_name'] = np.bytes_(get_grid_mapping_name(sr))
```
gmgunter, line 800: LGTM, thanks.

I recall that @hfattahi raised a [concern](https://github-fn.jpl.nasa.gov/isce-3/isce/pull/2179#issuecomment-23083) about whether this change would affect NISAR product compatibility with GIS software (panoply in particular). Is that something we should test before merging this? I'm not sure who on the team is familiar with panoply or has experience reading NISAR products with any other GIS software.

gshiroma: I believe the change is safe. I tested it by generating a GCOV product and successfully opening it with Panoply, QGIS, and HDFView. All tools were able to open the product without issues. Notice that in the same H5 dataset `projection`, we already have other byte string attributes: `ellipsoid` and `spatial_ref`. But that was a good point. Thanks for checking, @gmgunter .

gmgunter: Thanks for testing that! :rocket:

gmgunter (approved):

PR #52, opened 2025-07-03
Disable polarimetric symmetrization by default in the GCOV runconfig (MCR-93801)

gshiroma:
This PR disables polarimetric symmetrization by default in the GCOV runconfig. The change is motivated by the following reasons:
- Faraday rotation will probably not be corrected in GCOV products by default, which can impact the reciprocity between the HV and VH polarizations and impair polarimetric symmetrization, especially at L-band.
- Polarimetric crosstalk calibration will not be applied to RSLC products, potentially degrading the quality of symmetrization.
- Only about 1% of products are expected to be quad-pol, so disabling symmetrization by default should have a minimum impact on storage requirements and runtime.


hfattahi:
Thanks @gshiroma This requires MCR approval. Let's get the MCR approved before adding to R4.1.0 Milestone.

hfattahi (approved):
Thanks @gshiroma . LGTM. For full transparency to SDS please also turn this to False in Parameter Set Files.

bhawkins (approved):

gshiroma:
> Thanks @gshiroma . LGTM. For full transparency to SDS please also turn this to False in Parameter Set Files.

Will do. Thank you, @hfattahi and @bhawkins !

PR #53, opened 2025-07-03
Review the C++ implementation of the binary search function

gshiroma:
This PR sets an upper limit on the number of iterations for the binary search in ISCE3. It addresses an issue reported [here](https://github.com/opera-adt/RTC/issues/94#issuecomment-2948290065), where Topo processing runs indefinitely. The root cause was traced to the presence of NaNs in the DEM, which prevented ISCE3’s binary search from converging to a solution. To prevent such cases from causing infinite loops, this PR adds a maximum iteration threshold to the while loop in the binary search.

(UPDATE) This PR now reviews the binary search function `binarySearch()` entirely and adds unit tests to verify its correctness.

### DIFF
--- a/cxx/isce3/core/Utilities.h
+++ b/cxx/isce3/core/Utilities.h
@@ -357,7 +357,6 @@ namespace isce3 { namespace core {
         // The loop is guaranteed to converge with correct input, but we include
         // this check as a safe measure for unexpected edge cases.
         std::size_t max_iter = std::ceil(std::log2(array.size())) + 1;
-        // std::cout << "binary search" << std::endl;
         while (left + 1 < right) {
             const int middle = static_cast<int>(std::round(0.5 * (left + right)));
             const auto middle_value = array[middle];
@@ -373,7 +372,6 @@ namespace isce3 { namespace core {
                 throw std::runtime_error(
                     "Binary search failed to converge within the allowed iterations.");
             }
-            // std::cout << "count:" << count << std::endl;
         }
 
         if (std::isnan(array[left]) || std::isnan(array[right])) {
--- a/cxx/isce3/core/Utilities.h
+++ b/cxx/isce3/core/Utilities.h
@@ -334,38 +334,52 @@ namespace isce3 { namespace core {
         }
     }
     
-    /** Searches array for index closest to provided value */   
+    /** Searches array for index closest to provided value
+     *
+     *  Assumes the array is sorted in ascending order and contains no NaNs.
+    */
     inline int binarySearch(const std::valarray<double> & array, double value) {
-   
+
+        if (array.size() == 0) {
+                throw std::invalid_argument("input array must contain at least 1 element");
+        }
+        if (array.size() == 1) {
+                return 0;
+        }
+
         // Do the binary search 
         int left = 0;
         int right = array.size() - 1;
-        int index;
 
         std::size_t count = 0;
-        while (left <= right) {
+
+        // `max_iter` provides a safeguard against unexpected behavior.
+        // The loop is guaranteed to converge with correct input, but we include
+        // this check as a safe measure for unexpected edge cases.
+        std::size_t max_iter = std::ceil(std::log2(array.size())) + 1;
+        while (left + 1 < right) {
             const int middle = static_cast<int>(std::round(0.5 * (left + right)));
             const auto middle_value = array[middle];
             if (std::isnan(middle_value)) {
                 throw std::invalid_argument("input array may not contain NaN values");
             }
-            if (left == (right - 1)) {
-                index = left;
-                return index;
-            }
-            if (middle_value <= value) {
+            if (middle_value < value) {
                 left = middle;
             } else {
                 right = middle;
             }
-            count++;
-            if (count > array.size()) {
+            if (++count > max_iter) {
                 throw std::runtime_error(
                     "Binary search failed to converge within the allowed iterations.");
             }
         }
-        index = left;
-        return index;
+
+        // Return the closest of left and right
+        if (std::abs(array[left] - value) <= std::abs(array[right] - value)) {
+            return left;
+        }
+
+        return right;
     }
 
     /** Clip a number between an upper and lower range (implements std::clamp for older GCC) */
--- a/cxx/isce3/core/Utilities.h
+++ b/cxx/isce3/core/Utilities.h
@@ -357,6 +357,7 @@ namespace isce3 { namespace core {
         // The loop is guaranteed to converge with correct input, but we include
         // this check as a safe measure for unexpected edge cases.
         std::size_t max_iter = std::ceil(std::log2(array.size())) + 1;
+        // std::cout << "binary search" << std::endl;
         while (left + 1 < right) {
             const int middle = static_cast<int>(std::round(0.5 * (left + right)));
             const auto middle_value = array[middle];
@@ -372,6 +373,11 @@ namespace isce3 { namespace core {
                 throw std::runtime_error(
                     "Binary search failed to converge within the allowed iterations.");
             }
+            // std::cout << "count:" << count << std::endl;
+        }
+
+        if (std::isnan(array[left]) || std::isnan(array[right])) {
+            throw std::invalid_argument("input array may not contain NaN values");
         }
 
         // Return the closest of left and right
--- a/cxx/isce3/core/Utilities.h
+++ b/cxx/isce3/core/Utilities.h
@@ -337,6 +337,13 @@ namespace isce3 { namespace core {
     /** Searches array for index closest to provided value
      *
      *  Assumes the array is sorted in ascending order and contains no NaNs.
+     * @param array The input array, which must be sorted in ascending order and contain no NaNs.
+     * @param value The value to search for.
+     * @param always_pick_left If true, and `value` is strictly between two array elements,
+     *                         the function will return the lower index (left). If false, the
+     *                         function returns the closest index (left or right).
+     *
+     * @return The index of the array element closest to `value`.
     */
     inline int binarySearch(const std::valarray<double> & array, double value,
                             const bool always_pick_left = false) {
@@ -348,16 +355,24 @@ namespace isce3 { namespace core {
                 return 0;
         }
 
-        // Do the binary search 
         int left = 0;
         int right = array.size() - 1;
+
+        // Test extreme values
+        if (value <= array[left]) {
+            return left;
+        }
+        if (array[right] <= value) {
+            return right;
+        }
 
         std::size_t count = 0;
 
         // `max_iter` provides a safeguard against unexpected behavior.
         // The loop is guaranteed to converge with correct input, but we include
         // this check as a safe measure for unexpected edge cases.
-        std::size_t max_iter = std::ceil(std::log2(array.size())) + 1;
+        constexpr std::size_t padding = 2;
+        std::size_t max_iter = std::ceil(std::log2(array.size())) + padding;
         while (left + 1 < right) {
             const int middle = left + (right - left) / 2;
             const auto middle_value = array[middle];
@@ -375,12 +390,14 @@ namespace isce3 { namespace core {
             }
         }
 
-        if (always_pick_left) {
-            return left;
-        }
-
+        // Test for NaNs
         if (std::isnan(array[left]) || std::isnan(array[right])) {
             throw std::invalid_argument("input array may not contain NaN values");
+        }
+
+        // If always pick left and the `right` index is not pointing at value
+        if (always_pick_left && value < array[right]) {
+            return left;
         }
 
         // Return the closest of left and right
--- a/cxx/isce3/core/Utilities.h
+++ b/cxx/isce3/core/Utilities.h
@@ -340,9 +340,9 @@ namespace isce3 { namespace core {
         int right = array.size() - 1;
         int index;
 
-        long long counter = 0;
+        long long count = 0;
         long long MAX_NUM_ITERATIONS = 1e11;
-        while (left <= right and counter < MAX_NUM_ITERATIONS) {
+        while (left <= right and count < MAX_NUM_ITERATIONS) {
             const int middle = static_cast<int>(std::round(0.5 * (left + right)));
             if (left == (right - 1)) {
                 index = left;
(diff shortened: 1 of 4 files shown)


cxx/isce3/core/Utilities.h
```diff
@@ -339,7 +339,10 @@ namespace isce3 { namespace core {
         int left = 0;
         int right = array.size() - 1;
         int index;
-        while (left <= right) {
+
+        long long count = 0;
+        long long MAX_NUM_ITERATIONS = 1e11;
```
gmgunter: Wow, nice job tracking down this issue! That's a subtle one.

Not sure I understand the logic behind this fix, though. Previously, if `array` contained NaN values, the code could enter an infinite loop. Now, if `array` contains NaN values, the code might spin for 100 billion useless iterations and then return the wrong answer? Is that really the behavior that we want?

Since NaN values in `array` are the underlying issue, why don't we just add a check for NaNs instead?

```c++
#include <cmath>
#include <stdexcept>
```

```diff
+            const auto x = array[middle];
+            if (std::isnan(x)) {
+                raise std::invalid_argument("input array may not contain NaN values");
+            }
-            if (array[middle] <= value) {
+            if (x <= value) {
                 left = middle;
-            } else if (array[middle] > value) {
+            } else if (x > value) {
                 right = middle;
             }
```

gshiroma: I think that even if we handle the NaN case, it’s still important for this loop to have a fixed upper bound. But I like your suggestion and will include it. Thanks!

gshiroma: I incorporated your suggestion in commit https://github.com/isce-framework/isce3/pull/53/commits/e19052c137568469e15ee9cbaa01ac9fd96e371b. Thank you, @gmgunter !

gmgunter: >it’s still important for this loop to have a fixed upper bound

Why?

gshiroma: I think it's a good practice. It's also the second rule in [The Power of 10: Rules for Developing Safety-Critical Code](
https://en.wikipedia.org/wiki/The_Power_of_10:_Rules_for_Developing_Safety-Critical_Code).

gmgunter: Thanks for the link.

Leaving aside for the moment the question of whether or not it's desirable for isce3 to try to follow rules designed for safety-critical contexts, I'm not sure that spinning for 100 billion iterations and then returning the wrong answer would meet the rules' author's definition of safe code. The rule seems to encourage raising an error when the maximum iteration count is reached.

<p align="center">
  
</p>

Also, at the risk of muddying the waters a bit, I think the maximum number of iterations could be much smaller. Binary search should require at most $\log_{2} N$ iterations, where $N$ is the maximum size of the input array. So you could define the maximum number of iterations like this, I think:

```c++
#include <climits>  // for `CHAR_BIT`

constexpr int maxiter = sizeof(std::size_t) * CHAR_BIT;
```

But overall my preference would be to keep the NaN checking and remove the upper bound on the number of iterations.

cxx/isce3/core/Utilities.h
```diff
@@ -339,17 +339,25 @@ namespace isce3 { namespace core {
...
+        while (left <= right and count < MAX_NUM_ITERATIONS) {
             const int middle = static_cast<int>(std::round(0.5 * (left + right)));
+            const auto middle_value = array[middle];
+            if (std::isnan(middle_value)) {
+                throw std::invalid_argument("input array may not contain NaN values");
```
gmgunter, line 382: [`std::isnan`](https://en.cppreference.com/w/cpp/numeric/math/isnan.html) is defined in the `cmath` header and [`std::invalid_argument`](https://en.cppreference.com/w/cpp/error/invalid_argument.html) is defined in the `stdexcept` header, so we should add

```c++
#include <cmath>
#include <stdexcept>
```

at the top.

cxx/isce3/core/Utilities.h
```diff
@@ -339,17 +339,25 @@ namespace isce3 { namespace core {
...
+            }
             if (left == (right - 1)) {
                 index = left;
                 return index;
             }
-            if (array[middle] <= value) {
+            if (middle_value <= value) {
                 left = middle;
-            } else if (array[middle] > value) {
+            } else if (middle_value > value) {
```
gmgunter: Technically we don't need this `if` condition, I think.

If `middle_value` is not NaN and `middle_value` is not `<= value` (checked above on line 355), then `middle_value > value` must be true (for `double`, at least).

```suggestion
            } else {
```

cxx/isce3/core/Utilities.h
gshiroma: All great points! By the way, instead of using the formula above, why not just use `array.size()`? 

I’m still not sure why you don't like the idea of having an upper bound. If the code always converges, the exception will never be triggered. But if, for any reason, it doesn’t, raising an exception is much better than letting the process run indefinitely.

cxx/isce3/core/Utilities.h
gshiroma: Thank you, @gmgunter , this was updated in commit https://github.com/isce-framework/isce3/pull/53/commits/15d38d30b18a931f4268728887839b2790940af1.

cxx/isce3/core/Utilities.h
gshiroma: Sounds good. I wonder if that would make any difference in the compiled code. Anyways, I updated it in the commit https://github.com/isce-framework/isce3/pull/53/commits/feef376d97df4485b7a3bb9b3cb81b8b4497a4d4. Thanks!

cxx/isce3/core/Utilities.h
gshiroma: What do you think if the changes in the commit https://github.com/isce-framework/isce3/pull/53/commits/1b2e7e9a4c7e43dc561e10870d2d2472dfc23c9d?

gmgunter: >instead of using the formula above, why not just use `array.size()`?

The number of iterations should be log2 of the array size, so you could do this instead of the approach I used above:

```c++
const auto maxiter = static_cast<int>(std::ceil(std::log2(array.size())));
```

That seems fine too. I just thought my previous approach was a bit simpler, and it's nice that it's a compile-time constant so you don't have to do any additional math at runtime.

If we compute `maxiter` the way I described in the previous comment, it should always be equal to 64 on 64-bit machines, which isn't a very large number of iterations to try before giving up, so I figured it'd be reasonable even if it's a bit looser than the maximum number of iterations that are strictly necessary for a given array size.

>I’m still not sure why you don't like the idea of having an upper bound.

I just think it's an unnecessary pessimization of the algorithm. Generally we want to minimize the number of instructions executed in tight loops like this. This change is adding an additional increment and comparison operation to each iteration. If the function is free of bugs, these extra operations shouldn't be necessary since the number of iterations will never exceed the specified maximum.

Looking at the implementations of binary search in [libc++](https://github.com/llvm/llvm-project/blob/bc8aa974c05b35e8895086d499abb7b1b36b88fb/libcxx/include/__algorithm/lower_bound.h#L29-L48) and [NumPy](https://github.com/numpy/numpy/blob/46c0bffa8b137762dee19cbe1541790bd1108fb0/numpy/_core/src/npysort/binsearch.cpp#L171-L212), neither one seems to implement an explicit check that the loop doesn't exceed some maximum number of iterations. Presumably these are high-quality, mature implementations that we should seek to emulate.

That said, I'm much happier with it now that the code raises an exception upon hitting the max number of iterations instead of just returning the wrong answer (using an `assert` would also be fine in this case). Pipelining and branch prediction will probably hide most of the latency, and the way the function is implemented is already probably pretty non-optimal anyway.

So, while I would still prefer to remove the max iterations check, I think my concerns are non-blocking if the team would like to include this check.

cxx/isce3/core/Utilities.h
```diff
@@ -339,17 +341,28 @@ namespace isce3 { namespace core {
...
             }
-            if (array[middle] <= value) {
+            if (middle_value <= value) {
                 left = middle;
-            } else if (array[middle] > value) {
+            } else {
                 right = middle;
             }
+            count++;
+            if (count >= array.size()) {
```
gmgunter: This is comparing values of two different types. The type of `count` is `long long`. The type of `array.size()` is `std::size_t`. So every comparison will require casting `count` to `std::size_t`, which is pretty inefficient.

gmgunter: This will fail if the number of iterations is exactly equal to the array size. For example, if `array.size()` is 1, an exception will always be raised here.

```suggestion
            if (count > array.size()) {
```

cxx/isce3/core/Utilities.h
gshiroma: > That seems fine too. I just thought my previous approach was a bit simpler, and it's nice that it's a compile-time constant so you don't have to do any additional math at runtime.
> 
> If we compute `maxiter` the way I described in the previous comment, it should always be equal to 64 on 64-bit machines, which isn't a very large number of iterations to try before giving up, so I figured it'd be reasonable even if it's a bit looser than the maximum number of iterations that are strictly necessary for a given array size.
> 

I also used `array.size()` instead of `std::ceil(std::log2(array.size()))` to be a bit looser on the maximum number of iterations. My goal with this PR was just to detect cases in which the binary search is running unexpectedly long and we'd catch it. I don't feel very comfortable setting the maximum number of iterations to `std::ceil(std::log2(array.size()))`. 

> > I’m still not sure why you don't like the idea of having an upper bound.
> 
> I just think it's an unnecessary pessimization of the algorithm. Generally we want to minimize the number of instructions executed in tight loops like this. This change is adding an additional increment and comparison operation to each iteration. If the function is free of bugs, these extra operations shouldn't be necessary since the number of iterations will never exceed the specified maximum.
> 
> Looking at the implementations of binary search in [libc++](https://github.com/llvm/llvm-project/blob/bc8aa974c05b35e8895086d499abb7b1b36b88fb/libcxx/include/__algorithm/lower_bound.h#L29-L48) and [NumPy](https://github.com/numpy/numpy/blob/46c0bffa8b137762dee19cbe1541790bd1108fb0/numpy/_core/src/npysort/binsearch.cpp#L171-L212), neither one seems to implement an explicit check that the loop doesn't exceed some maximum number of iterations. Presumably these are high-quality, mature implementations that we should seek to emulate.

Well, I’m also sure their code went through much more scrutiny than ours. We don't expect our code to have bugs until we find them =). And thanks for sharing these other implementations!

> 
> That said, I'm much happier with it now that the code raises an exception upon hitting the max number of iterations instead of just returning the wrong answer (using an `assert` would also be fine in this case). Pipelining and branch prediction will probably hide most of the latency, and the way the function is implemented is already probably pretty non-optimal anyway.
> 
> So, while I would still prefer to remove the max iterations check, I think my concerns are non-blocking if the team would like to include this check.

Thank you, @gmgunter , I just don't want to use `std::ceil(std::log2(array.size()))` because I think it's too strict. Do you think we can use `array.size()` or do you still prefer your suggestion with `constexpr int maxiter = sizeof(std::size_t) * CHAR_BIT;`?

gmgunter: I think we agree that, at least in theory, the binary search implementation should take at most $\log_{2} N$ iterations. It seems like your concern is that the code may be buggy, causing it to perform additional iterations in some circumstances. Is that why you prefer to use `array.size()` instead of `std::ceil(std::log2(array.size()))`?

If we use the stricter upper bound, and the code does have a bug that causes it to be exceeded, then that will just raise an exception that alerts us to the bug so we can fix it. I'm not sure why that'd be more risky than using a looser tolerance that allows such bugs to go undetected.

Beyond that, my concern is that using `array.size()` instead of `std::ceil(std::log2(array.size()))` would be misleading for someone reading the code. If I was reading this function body for the first time, I think I would be very confused about why the authors expected binary search to take $N$ iterations instead of $\log_2 N$.

gshiroma: Yeah, I have my own concerns about this function too. It was not my intent to review its functionality since we've been using it for a while. I'd avoid being too strict in these edge cases since we are so close to the NISAR launch, but it's probably a good idea to take a deeper look into it. Things that I found:

- I don't fully understand why we need to keep in the loop if `left == right` in:

```
while (left <= right) {
...
}
```

- Also, if `left` is equal to `right -1`, we always return the `left` in the next skip condition, even if `right` is closer to `value`.

```
            if (left == (right - 1)) {
                index = left;
                return index;
}
```
The conditions look a bit strange to me. I tried to improve and make the function more intuitive in commits https://github.com/isce-framework/isce3/pull/53/commits/96915ae696d3035b09de10c4eb802822fbc1ad3a and https://github.com/isce-framework/isce3/pull/53/commits/8893ba8ba9088203e4db9086d04960c23a470599. I also added some unit tests in commit https://github.com/isce-framework/isce3/pull/53/commits/0da209aced4de20fdb953956475b8ee37ed6ee02 .

Now, I feel more confident in using `std::ceil(std::log2(array.size())`. I also added `+1` just to avoid being too strict.

I hope the changes make sense. Please let me know your thoughts.

gshiroma: It turns out that `Topo.cpp` doesn't want the closest point (index) to `value` in the binary search. It wants the left index because of the linear interpolation that happens a few lines below (see code [here](https://github-fn.jpl.nasa.gov/isce-3/isce/blob/b38758ba6abb80c17726941e775d0657a0599ad4/cxx/isce3/geometry/Topo.cpp#L823)).  So, I added an argument to `binarySearch()` to force picking the `left` index (`always_pick_left`), and updated the unit test accordingly. Without these updates the computation of the layover-shadow mask fails in some points. See updates in the commit https://github.com/isce-framework/isce3/pull/53/commits/563fea9e31e65418f30e9c75e65aeb97d2b4b79b.

hfattahi:
My understanding is that this issue could be avoided if the user would pass correct data (DEM without NaN) to the module. I can imagine that downstream workflows would check the inputs before passing to this core module. Anyways, for now I removed R4.1 milestone as this does not seem NISAR or OPERA critical and requires MCR approval to be added to the milestone of the next delivery.

PR #54, opened 2025-07-07
Fix function `normalize_geocode_output_mode()`

gshiroma:
This PR fixes the function `normalize_geocode_output_mode()`. The interpolation algorithm are selected in the GCOV runconfig through the name of the algorithm ("sinc", "bilinear", "bicubic", "nearest", or "biquintic"), not "interp".

### DIFF
--- a/python/packages/isce3/geocode/geocode_cov.py
+++ b/python/packages/isce3/geocode/geocode_cov.py
@@ -25,7 +25,7 @@ def normalize_geocode_output_mode(mode: str | GeocodeOutputMode) -> GeocodeOutpu
 
     if mode == "area_projection":
         return GeocodeOutputMode.AREA_PROJECTION
-    if mode == "interp":
+    if mode in ["sinc", "bilinear", "bicubic", "nearest", "biquintic"]:
         return GeocodeOutputMode.INTERP
 
     raise ValueError(f"unexpected GeocodeOutputMode argument: {mode!r}")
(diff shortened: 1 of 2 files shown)


hfattahi (approved):
LGTM. The change seems logical and should not break anything. Only place this function is called is [here](https://github.com/isce-framework/isce3/blob/6116cdaca709c861839804567cca406667887d8b/python/packages/nisar/workflows/gcov_runconfig.py#L57) which takes the input from run config. And the run config schema has clearly documented the different options [here](https://github.com/isce-framework/isce3/blob/develop/share/nisar/schemas/gcov.yaml#L280) which are consistent with the change in this PR.

gshiroma:
> LGTM. The change seems logical and should not break anything. Only place this function is called is [here](https://github.com/isce-framework/isce3/blob/6116cdaca709c861839804567cca406667887d8b/python/packages/nisar/workflows/gcov_runconfig.py#L57) which takes the input from run config. And the run config schema has clearly documented the different options [here](https://github.com/isce-framework/isce3/blob/develop/share/nisar/schemas/gcov.yaml#L280) which are consistent with the change in this PR.

That was fast! Thank you, @hfattahi !

gmgunter (approved):
Thanks for the fix! LGTM :+1:

gshiroma:
> Thanks for the fix! LGTM 👍

Thank you, @gmgunter !

PR #56, opened 2025-07-15
Mitigate Scalloping in DM2-to-Science Workflow

rad-eng-59:
This PR focuses on two updates for the DM2-to-Science L0B workflow.
1. Mitigate scalloping caused by RX antenna pattern in mosaicking multi-channel DM2 into science L0B via 1-tap DBF.
2. Update respective WD/WL datasets in L0B to represent 1-tap DBF process.

Here is an example of scalloping and its mitigation in the respective RSLC pairs generated from DM2 product of REE Congo Sim.
dm2_rslc_congo_comparison_onetap_dbf_rx_antpat_corr


hfattahi:
> This PR focuses on two updates for the DM2-to-Science L0B workflow.
> 
> 1. Mitigate scalloping caused by RX antenna pattern in mosaicking multi-channel DM2 into science L0B via 1-tap DBF.
> 2. Update respective WD/WL datasets in L0B to represent 1-tap DBF process.
> 
> Here is an example of scalloping and its mitigation in the respective RSLC pairs generated from DM2 product of REE Congo Sim. dm2_rslc_congo_comparison_onetap_dbf_rx_antpat_corr

Thanks @rad-eng-59 for updating the plot. Very impressive what can be achieved with 1 tap.

python/packages/nisar/workflows/nisar_l0b_dm2_to_dbf.py
```diff
@@ -484,22 +578,49 @@ def nisar_l0b_dm2_to_dbf(args):
                 # mid AZ time at the center of the AZ block
                 azt_mid = azt_raw[rgl_slice].mean()
 
+                # Compute RX ANT Power Pattern calibration if requested
+                if args.rx_antpat_calib:
+                    if plot:
+                        plot_name = plot_name = out_path.joinpath(
```
bhawkins:
```suggestion
                        plot_name = out_path.joinpath(
```

python/packages/nisar/workflows/nisar_l0b_dm2_to_dbf.py
```diff
@@ -484,22 +578,49 @@ def nisar_l0b_dm2_to_dbf(args):
...
+                        # prior to echo multiplication to include pulse
+                        # extension!
+                        # Alternatively, apply ANT correction after range comp
+                        # as part of RSLC workflow (preferred)!
+                        inv_rxpat_1w[...] = fftconvolve(
+                            inv_rxpat_1w**2,
+                            (1 / nrgb_pw) * np.ones(nrgb_pw),
+                            mode='full'
+                            )[:sr.size]
+                        echo_dbf *= np.sqrt(inv_rxpat_1w)
```
bhawkins: This seems a little strange to me. Wouldn't you want to smooth the pattern before taking the inverse? It's also confusing to change the units of a variable. I'm thinking something like
```suggestion
                        inv_rxpat_1w[...] = np.sqrt(1.0 / fftconvolve(
                            1.0 / inv_rxpat_1w**2,
                            (1 / nrgb_pw) * np.ones(nrgb_pw),
                            mode='full'
                            ))[:sr.size]
                        echo_dbf *= inv_rxpat_1w
```

python/packages/nisar/workflows/nisar_l0b_dm2_to_dbf.py
```diff
@@ -544,6 +665,19 @@ def nisar_l0b_dm2_to_dbf(args):
                 dset_prod_out.write_direct(echo_dbf.astype(
                     'uint16').view(dset.dtype_storage), dest_sel=rgl_slice)
 
+                # compute WD/WL values to be updated per single-tap DBF
+                # use mid range line values
```
bhawkins, line 671: I'm having trouble understanding why this PR adds a bunch of range bookkeeping logic. I'd guess the idea is that the antenna pattern may not fully cover the swath, so we make the swath smaller and update all the metadata. But it seems like we're computing the pattern over all range bins, so I must be missing something.

python/packages/nisar/workflows/nisar_l0b_dm2_to_dbf.py
```diff
@@ -621,5 +755,184 @@ def _adjust_delays_in_place(dset, sample_delays):
...
+        # interpolate EL pattern for three points of EL within a beam
+        # and get its power in dB.
+        el_gain_point3 = _amp2db(
+            np.interp(el_points[i_slice],
+                      el_cuts.angle,
+                      el_cuts.copol_pattern[cc])
+        )
+        # perform 2-order polyfit of power (dB) as a function sr (km)
+        pf_coef_db_km = np.polyfit(
+            sr_points_km[i_slice], el_gain_point3, deg=2)
```
bhawkins, line 866: We fit like this to keep the algorithm fast, I guess? While in principle we could compute for every range bin (or tabulate like in focus.py), this is probably good enough for improving the radiometry of DM2 images.

python/packages/nisar/workflows/nisar_l0b_dm2_to_dbf.py
```diff
@@ -621,5 +755,184 @@ def _adjust_delays_in_place(dset, sample_delays):
...
+            'One-tap DBFed/Mosaicked Peak-Normalized RX EL Power Pattern')
+        plt.grid(True)
+        plt.savefig(plot_name)
+        plt.close()
+    # convert to linear scale check the min value to be non-zero
+    magpat1w[:] = 10 ** (magpat1w / 20)
+    min_mag = magpat1w.min()
+    if np.isclose(min_mag, 0):
+        raise ValueError(
+            'The one-way RX EL power pattern contain zero value(s)!')
```
bhawkins:
```suggestion
            'The one-way RX EL power pattern contain zero value(s)!  '
            'Consider using --max-p2p-ant option to limit dynamic range.'
            )
```

bhawkins (commented):
This is a nice feature that'll definitely increase our chances of getting a reasonable-looking image early in the mission. I've just got a couple of questions, which hopefully we can discuss in today's focusing/cal meeting.

python/packages/nisar/workflows/nisar_l0b_dm2_to_dbf.py
rad-eng-59: The initial range gating parameters `WD/WL` set by instrument for DM2 represent two active RX channels at a time while in this 1-tap DBFed/mosaicked science version of DM2 L0B there is simply a single active RX channel at a time. We need to update these DBFed-related parameters in case we want to turn on EAP in RSLC workflow to compensate for both RX+TX pattern in EL in the presence of proper instrument HDF5 product as an alternative way to mitigate scalloping. Basically, provide an opportunity with the proper inputs such as RD/WD/WL for RSLC to correct for two-way antenna pattern in place of `--rx-antpat-calib` when instrument HDF5 is available. Besides, we want our data widow position to properly represent contribution of each channel in this 1-tap DBFed science product of its DM2 version.

python/packages/nisar/workflows/nisar_l0b_dm2_to_dbf.py
rad-eng-59: Yes, but given the inverse is already generated, the convolution is lazily and directly applied to the inverse. I can change it but I have not noticed any remarkable difference in the respective RSLCs using either ones except for an overall scalar.

python/packages/nisar/workflows/nisar_l0b_dm2_to_dbf.py
bhawkins: That makes sense. Thanks!

bhawkins (approved):
LGTM

hfattahi (approved):
LGTM. Thanks @rad-eng-59 !

PR #57, opened 2025-07-15
Fix annotations compatibility with python < 3.10

rtburns-jpl:
This patch is already in our conda-forge feedstock: https://github.com/conda-forge/isce3-feedstock/commit/0726b64694bcb3e936d32c38a19ae09e32df0e48

### DIFF
--- a/python/packages/nisar/noise/noise_estimation_from_raw.py
+++ b/python/packages/nisar/noise/noise_estimation_from_raw.py
@@ -1,6 +1,7 @@
 """
 Functions and classes for noise power estimation from Raw data
 """
+from __future__ import annotations
 from warnings import warn
 from dataclasses import dataclass
 from collections.abc import Iterator


gmgunter (approved):
Thanks for the fix!

piyushrpt:
python 3.9 is EOL in Oct 2025. Maybe best to stop supporting it.

rad-eng-59 (approved):
Thanks for the fix @rtburns-jpl .
Out of curiosity, I wonder if this addition is necessary. The only annotation used in that script is `collections.abc.Iterator[slice]` which is supported by python `3.9`. Is not the current min version of python `3.9` in ISCE3?

rtburns-jpl:
The incompatible syntax is `datetime_utc: str | np.ndarray`, the `|` operator for type annotations was added in python 3.10.

rad-eng-59:
Totally missed it. Yes, that union operator is a problem. Thanks

PR #58, opened 2025-07-15
Pin all docker dependencies

rtburns-jpl:
* Replace all yum dependencies with locked conda deps
* Remove all yum commands from docker images

We were previously installing cuda and rpm-build dependencies from yum, but these are available from conda-forge, allowing us to exactly specify all of our dependencies in our conda spec files to make our images fully reproducible.

In the runtime image, we only need:
* cuda-cudart
* libcufft

In the dev image, we also need:
* cuda-nvcc
* cuda-cudart-dev
* libcufft-dev
* rpm-tools

I installed cuda dependencies from `nvidia/label/cuda-11.8.0` to match the previous versions used with the yum installation method.


gmgunter (approved):
Thanks!

Reviewed and approved this PR on the internal GitHub server: https://github-fn.jpl.nasa.gov/isce-3/isce/pull/2246

PR #59, opened 2025-07-17
Remove processing type "Undefined" from the GCOV & GSLC writer

gshiroma:
This PR removes processing type "Undefined" from the GCOV & GSLC writer. With the update, if a process type is not "PR" or "UR", it is considered as a "Custom" processing.

PR #60, opened 2025-07-17
Update NISAR InSAR product metadata to latest product specifications

vbrancat:
This PR addresses some inconsistencies between the metadata of NISAR InSAR HDF5 data products and the latest updates to the InSAR product specifications. 

Specifically, this PR reconciles the following inconsistencies:

1. Rephrase the description of /science/LSAR/identification/productType to match the latest updates
2. Reconcile inconsistencies in the description of the Line-Of-Sight (LOS) and along-track unit vectors
3. Fix inconsistent usage of ticks and backticks in the description of several InSAR datasets
4. Fix erroneous capitalization of referenceTerrainHeight dataset description

### DIFF
--- a/python/packages/nisar/products/insar/InSAR_base_writer.py
+++ b/python/packages/nisar/products/insar/InSAR_base_writer.py
@@ -928,8 +928,6 @@ class InSARBaseWriter(h5py.File):
             processing_type = np.bytes_('Urgent')
         elif processing_type == 'OD':
             processing_type = np.bytes_('Custom')
-        else:
-            processing_type = np.bytes_('Undefined')
 
         # Adopt same logic as RSLC, GSLC, GCOV
         # If no condition is met, assign string from runconfig
@@ -1158,7 +1156,10 @@ class InSARBaseWriter(h5py.File):
             DatasetParams(
                 "processingType",
                 processing_type,
-                "Nominal (or) Urgent (or) Custom (or) Undefined",
+                'Processing pipeline used to generate this granule. ' \
+                '"Nominal": standard production system; "Urgent": time-sensitive ' \
+                'processing in response to urgent response events; ' \
+                '"Custom": user-initiated processing outside the nominal production system',
             ),
             DatasetParams(
                 "radarBand", radar_band_name,
--- a/python/packages/nisar/products/insar/InSAR_base_writer.py
+++ b/python/packages/nisar/products/insar/InSAR_base_writer.py
@@ -911,7 +911,6 @@ class InSARBaseWriter(h5py.File):
         """
         Add the identification group to the product
         """
-        info_channel = journal.info('InSAR_base_writer.add_identification_to_hdf5')
         warning_channel = journal.warning('InSAR_base_writer.add_identification_to_hdf5')
         radar_band_name = self._get_band_name()
         primary_exec_cfg = self.cfg["primary_executable"]
--- a/python/packages/nisar/products/insar/InSAR_base_writer.py
+++ b/python/packages/nisar/products/insar/InSAR_base_writer.py
@@ -4,6 +4,7 @@ from itertools import product
 from typing import Any, Optional, Union
 
 import h5py
+import journal
 import numpy as np
 from isce3.core import crop_external_orbit
 from isce3.core.types import complex32, to_complex32
@@ -910,6 +911,7 @@ class InSARBaseWriter(h5py.File):
         """
         Add the identification group to the product
         """
+        info_channel = journal.info('InSAR_base_writer.add_identification_to_hdf5')
         radar_band_name = self._get_band_name()
         primary_exec_cfg = self.cfg["primary_executable"]
 
@@ -926,8 +928,11 @@ class InSARBaseWriter(h5py.File):
             processing_type = np.bytes_('Nominal')
         elif processing_type == 'UR':
             processing_type = np.bytes_('Urgent')
-        elif processing_type == 'OD':
+        else:
             processing_type = np.bytes_('Custom')
+            if processing_type != 'OD':
+                info_channel.log(
+                    f"Processing type is {processing_type}. Setting to default of 'Custom'")
 
         # Adopt same logic as RSLC, GSLC, GCOV
         # If no condition is met, assign string from runconfig
--- a/python/packages/nisar/products/insar/InSAR_base_writer.py
+++ b/python/packages/nisar/products/insar/InSAR_base_writer.py
@@ -380,7 +380,7 @@ class InSARBaseWriter(h5py.File):
 
         reference_terrain_height = "referenceTerrainHeight"
         reference_terrain_height_description = \
-            f"Reference Terrain Height as a function of time for {rslc_name} RSLC"
+            f"Reference terrain height as a function of time for {rslc_name} RSLC"
         if reference_terrain_height in src_param_group:
             src_param_group.copy(reference_terrain_height, dst_param_group)
             dst_param_group[reference_terrain_height].attrs['description'] = \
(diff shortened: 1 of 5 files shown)


gshiroma (approved):
LGTM. Thank you, @vbrancat

python/packages/nisar/products/insar/InSAR_base_writer.py
```diff
@@ -928,8 +928,6 @@ def add_identification_to_hdf5(self):
             processing_type = np.bytes_('Urgent')
         elif processing_type == 'OD':
             processing_type = np.bytes_('Custom')
-        else:
-            processing_type = np.bytes_('Undefined')
```
nemo794: Thank you for updating this conditional block to remove the "Undefined" option!

Hmm.. what happens if the user leaves that parameter blank (null) in the runconfig? Or provides a different option to this function? ([In the runconfig, the parameter is not required](https://github-fn.jpl.nasa.gov/isce-3/isce/blob/b38758ba6abb80c17726941e775d0657a0599ad4/share/nisar/schemas/insar.yaml#L76)). Then the `processing_type` variable in this function never gets sets. 🐛 

Consider updating the third condition (`elif processing_type == 'OD':`) to simply be the catch-all `else:`, and also include a log message to be explicit about what the workflow is doing. Maybe something like this:

Side Note: I don't know the logging function call for InSAR. 😬 

```python
        else:
            if processing_type != 'OD':
                <log>(f"{processing_type=}, which is not one of 'PR', 'UR', or 'OD'. Setting to default of 'Custom'.")
            processing_type = np.bytes_('Custom')
```

nemo794 (commented):
The updated descriptions match what's in PIX. Thank you!

One comment to avoid introducing a bug...

python/packages/nisar/products/insar/InSAR_base_writer.py
nemo794: Hmm... GSLC and GCOV implement a similar conditional block [here](https://github-fn.jpl.nasa.gov/isce-3/isce/blob/dcd4870bcaf0a3ef256fed6afddb1fc400aac13e/python/packages/nisar/products/writers/BaseWriterSingleInput.py#L663-L670).

@gshiroma , what are your thoughts on this update to the InSAR workflow? And can GSLC/GCOV workflow be updated to match what is decided here?

Thanks!

cc: @bhawkins , who will eventually need to implement this update for focus.py.

gshiroma: Hi @nemo794 , yes, we've put "Custom" in the else condition. See PR [here](https://github.com/isce-framework/isce3/pull/59).

nemo794: Ah, got it.

@gshiroma Thoughts on the suggestion to add a log message or something, so that the "OD" case and the workflow's behavior is more explicit for maintainability?

gshiroma: I think the idea was that "OD" and `None` would both map to "Custom", right? I don't see a need for a log message. Up to you.

nemo794: > I think the idea was that "OD" and None would both map to "Custom", right?

Yes, that is correct.

But for maintainability, our goals should be:
* Make it clear to future readers what all expected conditions are
* Alert users to changes that the workflow is doing to their inputs
* (Best effort) Proactively catch bugs if the user tries to do something incorrectly

Here:
*  The [expected conditions per the schema are `"PR"`, `"UR"`, `"OD"`, `None`](https://github-fn.jpl.nasa.gov/isce-3/isce/blob/b38758ba6abb80c17726941e775d0657a0599ad4/share/nisar/schemas/insar.yaml#L76).
    - Is there also a need for a generic catch-all for anything else the user might provide, e.g. if `self.cfg` is updated manually before it is used by this function?
*  [In the version proposed in the GSLC/GCOV PR](https://github.com/isce-framework/isce3/blob/5a09604a64c1c231e399a12c820cfa05aa5ea4cf/python/packages/nisar/products/writers/BaseWriterSingleInput.py#L663-L668), only the `"PR"` and `"UR"` cases are explicit to the reader. "OD" and `None` are not made explicit, and so there's no way for a casual reader to see the expected options.
   - **This lack of clarity is hard to maintain.**

There are many ways to achieve these goals; log messages or assert statements are some of those ways. I suggested using a log message in case we need a generic catch-all, but perhaps there's a preferred way?

In similar scenarios in QA, @gmgunter has enforced certain coding patterns to address these concerns. Some examples:

1) In `igram.py`, [there is a conditional block which with three expected conditions: "rifg", "runw", or "gunw"](https://github.com/isce-framework/nisarqa/blob/cd16089ae436be7815c4a95d00d36351858b658b/src/nisarqa/products/igram.py#L87-L93). We could have simply used the "else:" condition bare, but [during code review @gmgunter  requested the use of an assert statement](https://github-fn.jpl.nasa.gov/NISAR-ADT/QualityAssurance/pull/103#discussion_r18956) to 1) inform the first-time reader of the third expected condition, and 2) proactively prevent/catch bugs.
    - Note: An `assert` was used because it will only be triggered by a developer coding bug. (The assert would **not** be triggered by an issue with the user-provided inputs.)

2) When @gmgunter wrote `caltools.py` and [there were exactly two expected options](https://github.com/isce-framework/nisarqa/blob/cd16089ae436be7815c4a95d00d36351858b658b/src/nisarqa/products/caltools.py#L718-L724), the catch-all "else" block contained an assert to inform the reader of exactly what should be expected, and also raise an error if the user tried to provide something unexpected.
    - Note: An `assert` was used because it will only be triggered by a developer coding bug. (The assert would **not** be triggered by an issue with the user-provided inputs.)

3) In a different location in `igram.py`, [only RIFG, RUNW, and GUNW types were expected. Each were explicitly handled, and then an exception raised in the catch-all `else:`.](https://github.com/isce-framework/nisarqa/blob/cd16089ae436be7815c4a95d00d36351858b658b/src/nisarqa/products/igram.py#L42-L52) If the user failed to provide a valid input.
    - Note: The use of an `Exception` (and not an `assert`) was chosen to because the error would be caused by user-provided inputs, not from a developer coding bug.

Our case for these PRs is slightly different, but the needs for future maintainability are the same.

@gshiroma , Taking all of the above considerations into account, how would you propose address these needs/goals in this conditional block for processing type?

vbrancat: @nemo794 I agree with you. An "else" catch-all statement is the right way to go. I also like the suggestion of the logging. I have incorporated them both in my latest commit.

xhuang-jpl (approved):
LGTM, thank you

python/packages/nisar/products/insar/InSAR_base_writer.py
```diff
@@ -926,10 +928,11 @@ def add_identification_to_hdf5(self):
             processing_type = np.bytes_('Nominal')
         elif processing_type == 'UR':
             processing_type = np.bytes_('Urgent')
-        elif processing_type == 'OD':
-            processing_type = np.bytes_('Custom')
         else:
-            processing_type = np.bytes_('Undefined')
+            processing_type = np.bytes_('Custom')
+            if processing_type != 'OD':
```
nemo794, line 933: Please take a look at the log message @gshiroma used in https://github.com/isce-framework/isce3/pull/59 . It's a bit more informative that this message, but this simpler message is also functional. Your choice if you'd like to adjust it or not. :)

python/packages/nisar/products/insar/InSAR_base_writer.py
```diff
@@ -926,10 +928,11 @@ def add_identification_to_hdf5(self):
...
+                info_channel.log(
+                    f"Processing type is {processing_type}. Setting to default of 'Custom'")
```
nemo794: Please take a look at the log message @gshiroma used in https://github.com/isce-framework/isce3/pull/59 . It's a bit more informative that this message, but this simpler message is also functional. Your choice if you'd like to adjust it or not. :)

gshiroma (approved):
Changes LGTM. Thank you, @vbrancat .

PR #64, opened 2025-07-21
Update the InSAR writer orbit object when compute the radar grid nominal ground spacing

xhuang-jpl:
This PR is to update the RSLC orbit object when compute the radar grid nominal ground spacing in the InSAR writer. The change is to take into account the external orbit file, while the` ref_rslc.getOrbit()` only gets the real time orbit state vectors.


vbrancat (approved):
LGTM, thanks

xhuang-jpl:
@hfattahi would you please reviewing this PR? it is rather simple, and I have addressed your comments in our internal repo. Thank you

image

hfattahi (approved):
LGTM!

PR #68, opened 2025-07-22
Populate `dryTroposphericGeolocationCorrectionApplied` using the RSLC runconfig (MCR-93801)

gshiroma:
This PR updates the GCOV and GSLC writer (`BaseL2WriterSingleInput`) to populate the metadata field `dryTroposphericGeolocationCorrectionApplied` using runconfig field `dry_troposphere_model` parsed from the input RSLC H5 dataset `.../metadata/processingInformation/parameters/runConfigurationContents`. If the input RSLC product does not contain the H5 dataset or if the runconfig does not contain the field `dry_troposphere_model`, it is assumed that the dry tropospheric geolocation correction has been applied (default behavior) and therefore `dryTroposphericGeolocationCorrectionApplied` is set to `True`.


hfattahi (approved):
The only caveat is the intrinsic assumption here that if the run config does not have the dry_troposphere_model, then we assume it has been applied by RSLC which is the current default behavior. If by any chance RSLC will change its default behavior or if you make GCOV/GSLC from a reformatted ALOS RSLC to NISAR format, this assumption will not hold and the flag will be invalid. 
I think this is not an issue for NISAR production, since the run config will be fully populated. So it is ok to go forward. Perhaps adding the field to RSLC would be the proper approach over long run.

gshiroma:
> The only caveat is the intrinsic assumption here that if the run config does not have the dry_troposphere_model, then we assume it has been applied by RSLC which is the current default behavior. If by any chance RSLC will change its default behavior or if you make GCOV/GSLC from a reformatted ALOS RSLC to NISAR format, this assumption will not hold and the flag will be invalid. I think this is not an issue for NISAR production, since the run config will be fully populated. So it is ok to go forward. Perhaps adding the field to RSLC would be the proper approach over long run.

@hfattahi , you brought up a great point. I initially thought that it would better to set the flag `dryTroposphericGeolocationCorrectionApplied` to `True` if `dry_troposphere_model` is undetermined (i.e., missing from the runconfig), but now I'm thinking that this won't happen if the image is focused by ISCE3. If the field is missing from the user runconfig, it will be populated by the [default runconfig](https://github.com/isce-framework/isce3/blob/8160dc07f37688f8078a43d8fcd5f64aea73da5b/share/nisar/defaults/focus.yaml#L377).

But, as you said, this may happen on repackaged RSLC products. So, not I'm inclined to either set it to `False` or simply not adding the flag `dryTroposphericGeolocationCorrectionApplied` to the GCOV product. I think I prefer the latter. Notice that this case will not happen in production, so it will not affect the standard GCOV product. What do you think?

gshiroma:
@hfattahi , FYI, I updated the code following our discussion yesterday. If it's unclear whether the dry tropospheric correction has been applied or not, we set `dryTroposphericGeolocationCorrectionApplied` to '(NOT SPECIFIED)'

python/packages/nisar/products/writers/BaseWriterSingleInput.py
```diff
@@ -904,6 +904,55 @@ def copy_from_runconfig(self, h5_field,
...
+                               '/metadata/processingInformation/parameters/'
+                               'runConfigurationContents')
+
+        if input_h5_field_path not in self.input_hdf5_obj:
+            return
+
+        input_h5_dataset_obj = self.input_hdf5_obj[input_h5_field_path]
+
+        # check if dataset contains a string. If so, read it using method
+        # `asstr()``
```
seongsujeong:
```suggestion
        # `asstr()`
```
Just a nit pick

python/packages/nisar/products/writers/BaseWriterSingleInput.py
```diff
@@ -904,6 +904,55 @@ def copy_from_runconfig(self, h5_field,
 
         return self.set_value(h5_field, data=data, *args, **kwargs)
 
+    def get_value_from_input_runconfig(self, field_name):
```
seongsujeong, line 907: I understand this function to extract the value that corresponds to `field_name`. How about extracting the (dumped) runconfig values from HDF5, parse that using `yaml.save_load()`, and get the field value you want?

seongsujeong (commented):
LGTM overall. I have a question as below:

python/packages/nisar/products/writers/BaseWriterSingleInput.py
gshiroma: Thank you, @seongsujeong , I'm afraid a yaml parser will be stricter and start raising exceptions in fields that have nothing to do with `field_name`. Also, we probably would need to do some work to make the H5 dataset `runConfigurationContents` to be parsable by the yaml parser.

seongsujeong (approved):
LGTM

gshiroma:
Thank you, @seongsujeong !

PR #74, opened 2025-07-23
Reorganize populating CRS metadata into a separate function

gmgunter:
Currently, the [`set_get_geo_info()`](https://github.com/isce-framework/isce3/blob/d806ba7b9aaf151bbf9d1b2159195f795d4e5a94/python/packages/nisar/workflows/h5_prep.py#L666) function in `h5_prep.py` is responsible for populating a bunch of datasets and attributes in NISAR L2 products.

Among its responsibilities is adding attributes to the `projection` dataset that describe the coordinate reference system (CRS) of the product's raster grid(s). I'd like to reuse that functionality in the Static Layers workflow without inheriting all of the other behavior of `set_get_geo_info()`.

This PR is pulling out that functionality into a separate function that can be reused in the Static Layers workflow. The new function `build_projection_dataset_attrs_dict()` takes an input EPSG code and returns a dict of attributes that can then be added to the `projection` datasets in NISAR L2 products.

This PR is not intended to affect the functionality of existing NISAR SAS workflows -- it's just intended to modularize/reorganize the existing code to improve reusability.

I also added a few unit tests to validate the behavior for a few EPSG codes.

### DIFF
--- a/python/packages/isce3/core/__init__.py
+++ b/python/packages/isce3/core/__init__.py
@@ -16,5 +16,6 @@ from .orbit import (
 from .poly2d import fit_bivariate_polynomial
 from . import rdr_geo_block_generator
 from .block_param_generator import BlockParam
+from .projections import is_utm
 from .serialization import load_orbit_from_h5_group
 from . import types
--- a/python/packages/isce3/core/projections.py
+++ b/python/packages/isce3/core/projections.py
@@ -1,0 +1,16 @@
+def is_utm(epsg: int) -> bool:
+    """
+    Check if the input EPSG code represents a UTM zone.
+
+    Parameters
+    ----------
+    epsg : int
+        The input EPSG code.
+
+    Returns
+    -------
+    bool
+        True if the coordinate reference system represented by `epsg` is a Universal
+        Transverse Mercator (UTM) projection; otherwise False.
+    """
+    return (32600 < epsg <= 32660) or (32700 < epsg <= 32760)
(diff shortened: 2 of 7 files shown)


vbrancat:
@gmgunter I believe this change will break the InSAR workflow. We use this functionality in `h5_prep.py ` in parts of our product writers, but I am all for moving it elsewhere (it was a long due change)

gshiroma (approved):
LGTM. Thank you for organizing it, @gmgunter !

gmgunter:
@vbrancat I tried to implement the changes in a way that wouldn't affect the behavior of `h5_prep.py`. Are you concerned that this refactor may introduce some bugs?

If you're not comfortable with this update, I could revert the change to `h5_prep.py`, but then there will be two redundant implementations of this functionality -- one used by Static Layers and another used by the other L2 workflows.

Let me know your preference.

vbrancat:
@gmgunter I would like all the L2 workflows to use the functionalities that you have refactored and get rid of (the repugnant) `h5_prep.py`, which is deprecated. However, this is not your responsibility :P 

To be 100% conservative, can you do a couple of tests and make sure that the InSAR SAS is not broken?

vbrancat (approved):
@gmgunter I ended up exercising the InSAR workflow using this PR and there is no impact on the InSAR SAS. Thanks

xhuang-jpl (approved):
same with @vbrancat , my test is also good without any issue. Thanks @gmgunter

gmgunter:
Thank you @vbrancat and @xhuang-jpl for testing this out!

PR #77, opened 2025-07-25
Add license to staged WaterMask VRT

oberonia78:
This PR introduced a small function (parse_readme_fields) that scans the `LICENSE.txt` from the released water mask in s3 bucket, and copy them into staged vrt file's metadata. The collected information includes items such as the product name, version, processing date, notes, and source datasets.

If the README/Licence file is absent (v0.4 or v0.3), the code now issues a warnings.warn(...) message and continues, instead of raising an exception.

### DIFF
--- a/python/packages/nisar/workflows/stage_watermask.py
+++ b/python/packages/nisar/workflows/stage_watermask.py
@@ -407,43 +407,6 @@ def download_watermask(polys, epsgs, outfile, version):
         raise ValueError(errmsg)
 
 
-def extract_dem_description(in_readme_path):
-    """Extract DEM description from README.txt on nisar-dem
-       s3 bucket
-
-    Parameters
-    ----------
-    in_readme_path: str
-        Path to the README.txt on the nisar-dem
-        s3 bucket
-
-    Returns
-    -------
-    dem_descr: str
-        String containing the "dem description"
-        extracted from the README.txt
-    """
-    pattern = r'^\s*- Short description: (.+)$'
-
-    # JPL internal s3 buckets are not accessible via
-    # https addresses due to cybersecurity concerns. This
-    # excludes using "requests". Using boto3 and its AWS s3
-    # API would add another unnecessary dependency to ISCE3.
-    # Therefore, we use GDAL to read a remote text file.
-    fp = gdal.VSIFOpenL(in_readme_path, "rb")
-    text = gdal.VSIFReadL(1, 100000, fp).decode()
-    gdal.VSIFCloseL(fp)
-
-    match = re.search(pattern, text, re.MULTILINE)
-    if match:
-        dem_descr = match.group(1)
-    else:
-        err_str = 'Line with "Short Description" not found in README.txt'
-        raise ValueError(err_str)
-
-    return dem_descr
-
-
 def parse_readme_fields(
         readme_path: str,
         encoding: str = "utf-8"
--- a/python/packages/nisar/workflows/stage_watermask.py
+++ b/python/packages/nisar/workflows/stage_watermask.py
@@ -386,18 +386,19 @@ def download_watermask(polys, epsgs, outfile, version):
         vrt_dataset = gdal.BuildVRT(outfile, watermask_list)
 
         # Get the WATER description from the LICENSE.txt file using GDAL
-        in_license_path = vrt_filename.replace(f'EPSG{epsg}.vrt', f'EPSG{epsg}/LICENSE.txt')
+        in_license_path = vrt_filename.replace(f'EPSG{epsg}.vrt',
+                                               f'EPSG{epsg}/LICENSE.txt')
         license_exists = gdal.VSIStatL(in_license_path) is not None
 
         if license_exists:
             try:
-                water_descr = parse_license_fields(in_license_path)
-                for key, val in water_descr.items():
-                    vrt_dataset.SetMetadataItem(key, val)
+                water_descr = parse_license_notes(in_license_path)
+                vrt_dataset.SetMetadataItem("water_mask_description",
+                                            f'{water_descr}')
 
             except Exception as e:
                 warnings.warn(
-                    f"Found {license_path} but could not parse it ({e}). "
+                    f"Found {in_license_path} but could not parse it ({e}). "
                     "Proceeding without metadata."
                 )
         else:
@@ -405,17 +406,18 @@ def download_watermask(polys, epsgs, outfile, version):
                 f"No LICENSE.txt found at {in_license_path} "
                 f"(expected for WATERMASK v{version}). Proceeding without metadata."
             )
-    except Exception:
+    except Exception as e:
         errmsg = f'Failed to download NISAR WATERMASK {version} from s3 bucket. ' \
                  f'Maybe {version} is not currently supported.'
         raise ValueError(errmsg) from e
 
 
-def parse_license_fields(license_path: str,
-                         encoding: str = "utf-8") -> dict[str, str]:
+def parse_license_notes(license_path: str,
+                        encoding: str = "utf-8") -> str | None:
     """
-    Fetch a LICENSE.txt living in an S3 bucket via GDAL and return all
-    bullet style fields in a dictionary.
+    Extract the multi-line text of the “Notes” bullet from
+    a LICENSE.txt in an S3 bucket via GDAL and returning the
+    value without the leading "- Notes:" header.
 
     Parameters
     ----------
@@ -426,65 +428,49 @@ def parse_license_fields(license_path: str,
 
     Returns
     -------
-    fields : dict[str, str]
-        Mapping from the bullet header (e.g. "Short description")
-        to its full text (potentially multi line).
+    str | None
+        The extracted Notes text (possibly multi-line) with surrounding
+        whitespace stripped, or None if no “Notes” bullet is found.
     """
     stat = gdal.VSIStatL(license_path)
     if stat is None:
         raise ValueError(f"No LICENSE.txt found at {license_path}")
 
-    MAX_SIZE = 10_000_000 # 10 MB sanity limit
-    if stat and stat.size > MAX_SIZE:
-        warnings.warn(
-            f"{license_path} is {stat.size/1_000_000:.1f} MB—"
-            "skipping LICENSE parsing."
-        )
+    MAX_SIZE = 10_000_000  # 10 MB
 
-    # Pull the file into memory
+    if stat.size and stat.size > MAX_SIZE:
+        warnings.warn(
+            f"{license_path} is {stat.size/1_000_000:.1f} MB — "
+            f"only the first {MAX_SIZE/1_000_000:.0f} MB will be parsed."
+        )
     fp = gdal.VSIFOpenL(license_path, "rb")
     if fp is None:
         raise IOError(f"Could not open {license_path} via GDAL")
-
     try:
-        data = gdal.VSIFReadL(1, MAX_SIZE, fp)      # read up to 10 MB
+        data = gdal.VSIFReadL(1, min(stat.size, MAX_SIZE), fp)
     finally:
         gdal.VSIFCloseL(fp)
 
-    if data is None or len(data) == 0:
+    if not data:
         raise IOError(f"Failed to read any data from {license_path}")
 
     text = data.decode(encoding, errors="replace")
 
-    # Parse bullet fields
-    fields = {}
-    curr_key, curr_val_lines = None, []
+    if "notes:" not in text.lower():
+        warnings.warn(f'No "Notes" field found in {license_path}')
+        return None
 
-    regex = re.compile(r"^\s*-\s+(.*?):\s*(.*)$")
-    for line in text.splitlines():
-        bullet_match = regex.match(line)
-        if bullet_match:
-            # Save the previous field before starting a new one
-            if curr_key is not None:
-                fields[curr_key] = "\n".join(curr_val_lines).strip()
+    NEXT_BULLET = r"(?=^\s*-\s+[^:\n]+:\s*|\Z)"
+    NOTES_RE = re.compile(
+        rf"^\s*-\s*Notes:\s*(.*?){NEXT_BULLET}",
+        flags=re.IGNORECASE | re.MULTILINE | re.DOTALL)
 
-            # Start a new field
-            curr_key = bullet_match.group(1).strip()
-            first_val_segment = bullet_match.group(2).rstrip()
-            curr_val_lines = [first_val_segment] if first_val_segment else []
-        else:
-            # Continuation of current bullet (indentation or empty line)
-            if curr_key is not None:
-                curr_val_lines.append(line.rstrip())
+    m = NOTES_RE.search(text)
+    if not m:
+        warnings.warn(f'No "Notes" field found in {license_path}')
+        return None
 
-    # Capture the final field
-    if curr_key is not None:
-        fields[curr_key] = "\n".join(curr_val_lines).strip()
-
-    if not fields:
-        raise ValueError(f"No bullet-style fields detected in {license_path}")
-
-    return fields
+    return m.group(1).strip()
 
 
 def transform_polygon_coords(polys, epsgs):
--- a/python/packages/nisar/workflows/stage_watermask.py
+++ b/python/packages/nisar/workflows/stage_watermask.py
@@ -4,8 +4,10 @@
 
 import argparse
 import os
-import backoff
+import re
+import warnings
 
+import backoff
 import numpy as np
 import shapely.ops
 import shapely.wkt
@@ -25,7 +27,7 @@ WATER_MASK_VSIS3_PATH = f'/vsis3/{STATIC_REPO}/WATER_MASK'
 
 def cmdLineParse():
     """
-     Command line parser
+    Command line parser
     """
     parser = argparse.ArgumentParser(description="""
                                      Stage and verify WATERMASK for processing. """,
@@ -68,7 +70,7 @@ def check_dateline(poly):
 
     xmin, _, xmax, _ = poly.bounds
     # Check dateline crossing
-    if (xmax - xmin) > 180.0:
+    if ((xmax - xmin > 180.0) or (xmin <= 180.0 <= xmax)):
         dateline = shapely.wkt.loads('LINESTRING( 180.0 -90.0, 180.0 90.0)')
 
         # build new polygon with all longitudes between 0 and 360
@@ -98,7 +100,7 @@ def check_dateline(poly):
             x_wrapped_minus_360 = np.asarray(x) - 360
             polys[polygon_count] = Polygon(zip(x_wrapped_minus_360, y))
 
-        assert len(polys) == 2
+        assert (len(polys) == 2)
     else:
         # If dateline is not crossed, treat input poly as list
         polys = [poly]
@@ -110,7 +112,7 @@ def determine_polygon(ref_slc, bbox=None):
     """Determine bounding polygon using RSLC radar grid/orbit
     or user-defined bounding box
 
-    Parameters:
+    Parameters
     ----------
     ref_slc: str
         Filepath to reference RSLC product
@@ -118,7 +120,7 @@ def determine_polygon(ref_slc, bbox=None):
         Bounding box with lat/lon coordinates (decimal degrees)
         in the form of [West, South, East, North]
 
-    Returns:
+    Returns
     -------
     poly: shapely.Geometry.Polygon
         Bounding polygon corresponding to RSLC perimeter
@@ -137,14 +139,14 @@ def determine_polygon(ref_slc, bbox=None):
 def point2epsg(lon, lat):
     """Return EPSG code based on point lat/lon
 
-    Parameters:
+    Parameters
     ----------
     lat: float
         Latitude coordinate of the point
     lon: float
         Longitude coordinate of the point
 
-    Returns:
+    Returns
     -------
     epsg code corresponding to the point lat/lon coordinates
     """
@@ -167,8 +169,8 @@ def get_geo_polygon(ref_slc, min_height=-500.,
                     max_height=9000., pts_per_edge=5):
     """Create polygon (EPSG:4326) using RSLC radar grid and orbits
 
-    Parameters:
-    -----------
+    Parameters
+    ----------
     ref_slc: str
         Path to RSLC product to stage the water mask for
     min_height: float
@@ -178,7 +180,7 @@ def get_geo_polygon(ref_slc, min_height=-500.,
     pts_per_edge: float
         Number of points per edge for min/max bounding box computation
 
-    Returns:
+    Returns
     -------
     poly: shapely.Geometry.Polygon
         Bounding polygon corresponding to RSLC perimeter on the ground
@@ -220,13 +222,13 @@ def determine_projection(polys):
     EPSG is computed for a regular list of points. EPSG
     is assigned based on a majority criteria.
 
-    Parameters:
-    -----------
+    Parameters
+    ----------
     polys: shapely.Geometry.Polygon
         List of shapely Polygons
 
-    Returns:
-    --------
+    Returns
+    -------
     epsg:
         List of EPSG codes corresponding to elements in polys
     """
@@ -268,7 +270,7 @@ def translate_watermask(vrt_filename, outpath, x_min, x_max, y_min, y_max):
        throttling (see "Query throttling" section at
        https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html).
 
-    Parameters:
+    Parameters
     ----------
     vrt_filename: str
         Path to the input VRT file
@@ -289,6 +291,20 @@ def translate_watermask(vrt_filename, outpath, x_min, x_max, y_min, y_max):
     input_x_min, xres, _, input_y_max, _, yres = ds.GetGeoTransform()
     length = ds.GetRasterBand(1).YSize
     width = ds.GetRasterBand(1).XSize
+
+    # Declare lambda function to snap min/max X and Y
+    # coordinates over the water grid
+    snap_coord = lambda val, snap, offset, round_func: round_func(  # noqa: E731
+        float(val - offset) / snap) * snap + offset
+
+    # Snap edge coordinates using the DEM pixel spacing
+    # and starting coordinates. Max values are rounded
+    # using np.ceil and min values are rounded with np.floor
+    x_min = snap_coord(x_min, xres, input_x_min, np.floor)
+    x_max = snap_coord(x_max, xres, input_x_min, np.ceil)
+    y_min = snap_coord(y_min, yres, input_y_max, np.floor)
+    y_max = snap_coord(y_max, yres, input_y_max, np.ceil)
+
     input_y_min = input_y_max + (length * yres)
     input_x_max = input_x_min + (width * xres)
 
@@ -324,13 +340,13 @@ def translate_watermask(vrt_filename, outpath, x_min, x_max, y_min, y_max):
 def download_watermask(polys, epsgs, outfile, version):
     """Download water mask from nisar-WATERMASK bucket
 
-    Parameters:
+    Parameters
     ----------
     polys: shapely.geometry.Polygon
         List of shapely polygons
     epsg: str, list
         List of EPSG codes corresponding to polys
-    outfile:
+    outfile: str
         Path to the output WATERMASK file to be staged
     version: str
         Water mask version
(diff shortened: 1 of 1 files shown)


python/packages/nisar/workflows/stage_watermask.py
```diff
@@ -68,7 +70,7 @@ def check_dateline(poly):
 
     xmin, _, xmax, _ = poly.bounds
     # Check dateline crossing
-    if (xmax - xmin) > 180.0:
+    if ((xmax - xmin > 180.0) or (xmin <= 180.0 <= xmax)):
```
gmgunter: What about this case?

```python
(xmin < -180 <= xmax)
```

Isn't that basically the same as the case you're adding here? Wouldn't that also indicate a dateline crossing?

python/packages/nisar/workflows/stage_watermask.py
```diff
@@ -289,6 +291,20 @@ def translate_watermask(vrt_filename, outpath, x_min, x_max, y_min, y_max):
     input_x_min, xres, _, input_y_max, _, yres = ds.GetGeoTransform()
     length = ds.GetRasterBand(1).YSize
     width = ds.GetRasterBand(1).XSize
+
+    # Declare lambda function to snap min/max X and Y
+    # coordinates over the water grid
+    snap_coord = lambda val, snap, offset, round_func: round_func(  # noqa: E731
+        float(val - offset) / snap) * snap + offset
```
gmgunter: Instead of adding the `# noqa: E731` comment to disable the [E731](https://docs.astral.sh/ruff/rules/lambda-assignment/) check, we could just implement its suggestion :shrug:. It has the extra benefit of also being more readable (IMO).

```suggestion
    def snap_coord(val, snap, offset, round_func):
        return round_func(float(val - offset) / snap) * snap + offset
```

gmgunter: No need to convert to `float` -- the "true division" operator (`/`) in Python3 always performs floating-point division.

image

```suggestion
    snap_coord = lambda val, snap, offset, round_func: round_func(  # noqa: E731
        (val - offset) / snap) * snap + offset
```

python/packages/nisar/workflows/stage_watermask.py
```diff
@@ -289,6 +291,20 @@ def translate_watermask(vrt_filename, outpath, x_min, x_max, y_min, y_max):
...
+
+    # Snap edge coordinates using the DEM pixel spacing
```
gmgunter: There's no DEM here.

```suggestion
    # Snap edge coordinates using the raster pixel spacing
```

python/packages/nisar/workflows/stage_watermask.py
```diff
@@ -366,19 +382,99 @@ def download_watermask(polys, epsgs, outfile, version):
...
+            water_descr = parse_readme_fields(in_readme_path)
+
+            # Build vrt with downloaded watermasks
+            vrt_dataset = gdal.BuildVRT(outfile, watermask_list)
+            # vrt_dataset.SetMetadataItem("dem_description", f'{dem_descr}')
+            for key, val in water_descr.items():
+                # # convert the bullet header to a tidy metadata key
+                # clean_key = key.lower().replace(" ", "_")
+                # vrt_dataset.SetMetadataItem(clean_key, val)
+                vrt_dataset.SetMetadataItem(key, val)
```
gmgunter: Let's remove this commented-out code.

```suggestion
            # Build vrt with downloaded watermasks
            vrt_dataset = gdal.BuildVRT(outfile, watermask_list)
            for key, val in water_descr.items():
                vrt_dataset.SetMetadataItem(key, val)
```

python/packages/nisar/workflows/stage_watermask.py
```diff
@@ -366,19 +382,99 @@ def download_watermask(polys, epsgs, outfile, version):
...
+        except Exception as e:
+            warnings.warn(
+                f"Could not read or parse README.txt at {in_readme_path} "
+                f"({e}). Proceeding without metadata."
+            )
```
gmgunter: The call to `gdal.BuildVRT()` shouldn't be inside this `try:` block. If an exception is raised by `gdal.BuildVRT()`, it will be caught and treated as a warning -- we don't want that.

In general, it's good practice to minimize the contents of `try:` blocks to just the line(s) of code that you expect to raise an exception that you want to handle.

python/packages/nisar/workflows/stage_watermask.py
```diff
@@ -366,19 +382,99 @@ def download_watermask(polys, epsgs, outfile, version):
...
+
     except Exception:
         errmsg = f'Failed to download NISAR WATERMASK {version} from s3 bucket. ' \
                  f'Maybe {version} is not currently supported.'
         raise ValueError(errmsg)
```
gmgunter: Let's [chain](https://docs.python.org/3/tutorial/errors.html#exception-chaining) these exceptions so that the original exception message isn't lost, since the new error message doesn't give any useful information about what actually went wrong.

```suggestion
    except Exception as e:
        errmsg = f'Failed to download NISAR WATERMASK {version} from s3 bucket. ' \
                 f'Maybe {version} is not currently supported.'
        raise ValueError(errmsg) from e
```

python/packages/nisar/workflows/stage_watermask.py
```diff
@@ -366,19 +382,99 @@ def download_watermask(polys, epsgs, outfile, version):
...
 
 
+def parse_readme_fields(
+        readme_path: str,
+        encoding: str = "utf-8"
+        ) -> dict[str, str]:
```
gmgunter: Weird indentation style. This doesn't seem to match the style used elsewhere in this script, e.g.:

https://github.com/isce-framework/isce3/blob/607825a8ae7d88dcc6571e1b05d262bf4e2e949a/python/packages/nisar/workflows/stage_watermask.py#L168-L169

python/packages/nisar/workflows/stage_watermask.py
```diff
@@ -366,19 +382,99 @@ def download_watermask(polys, epsgs, outfile, version):
...
+    encoding : str, default "utf-8"
+        Text encoding to decode the bytes.
+
+    Returns
+    -------
+    fields : dict[str, str]
+        Mapping from the bullet header (e.g. "Short description")
+        to its full text (potentially multi line).
+    """
+    # ---- 1. Pull the file into memory ----------------------------------
```
gmgunter: Why are we introducing this new commenting style here? I haven't seen this used elsewhere in the file (or in isce3 for that matter). Let's just stick with normal comments and maintain stylistic consistency.

```suggestion
    # Pull the file into memory
```

python/packages/nisar/workflows/stage_watermask.py
```diff
@@ -366,19 +382,99 @@ def download_watermask(polys, epsgs, outfile, version):
...
+        raise IOError(f"Could not open {readme_path} via GDAL")
+
+    try:
+        data = gdal.VSIFReadL(1, 10_000_000, fp)      # read up to 10 MB
+    finally:
+        gdal.VSIFCloseL(fp)
+
+    text = data.decode(encoding, errors="replace")
+
+    # ---- 2. Parse bullet fields ----------------------------------------
```
gmgunter:
```suggestion
    # Parse bullet fields
```

python/packages/nisar/workflows/stage_watermask.py
```diff
@@ -366,19 +382,99 @@ def download_watermask(polys, epsgs, outfile, version):
...
+        Mapping from the bullet header (e.g. "Short description")
+        to its full text (potentially multi line).
+    """
+    # ---- 1. Pull the file into memory ----------------------------------
+    fp = gdal.VSIFOpenL(readme_path, "rb")
+    if fp is None:
+        raise IOError(f"Could not open {readme_path} via GDAL")
+
+    try:
+        data = gdal.VSIFReadL(1, 10_000_000, fp)      # read up to 10 MB
```
gmgunter: Should we add a check for the case where the file is >10MB in size?

python/packages/nisar/workflows/stage_watermask.py
```diff
@@ -366,19 +382,99 @@ def download_watermask(polys, epsgs, outfile, version):
...
+        gdal.VSIFCloseL(fp)
+
+    text = data.decode(encoding, errors="replace")
+
+    # ---- 2. Parse bullet fields ----------------------------------------
+    fields = {}
+    curr_key, curr_val_lines = None, []
+
+    for line in text.splitlines():
+        bullet_match = re.match(r"^\s*-\s+(.*?):\s*(.*)$", line)
```
gmgunter: This is recompiling the regex in each iteration of the `for` loop. It'd be more efficient to compile the regex before entering the loop.

```suggestion
    regex = re.compile(r"^\s*-\s+(.*?):\s*(.*)$")
    for line in text.splitlines():
        bullet_match = regex.match(line)
```

python/packages/nisar/workflows/stage_watermask.py
```diff
@@ -366,19 +382,99 @@ def download_watermask(polys, epsgs, outfile, version):
             xmin, ymin, xmax, ymax = poly.bounds
             translate_watermask(vrt_filename, outpath, xmin, xmax, ymin, ymax)
 
-        # Build vrt with downloaded watermasks
-        gdal.BuildVRT(outfile, watermask_list)
+        # Get the WATER description from the README.txt file using GDAL
+        in_readme_path = vrt_filename.replace(f'EPSG{epsg}.vrt', f'EPSG{epsg}/LICENSE.txt')
```
gmgunter: The comment says we're reading the `README.txt` file and the variable name mentions "readme", but we're actually reading the `LICENSE.txt` file.

```suggestion
        # Get the WATER description from the LICENSE.txt file using GDAL
        in_license_path = vrt_filename.replace(f'EPSG{epsg}.vrt', f'EPSG{epsg}/LICENSE.txt')
```

python/packages/nisar/workflows/stage_watermask.py
```diff
@@ -366,19 +382,99 @@ def download_watermask(polys, epsgs, outfile, version):
...
+        encoding: str = "utf-8"
+        ) -> dict[str, str]:
+    """
+    Fetch a README.txt living in an S3 bucket via GDAL and return all
+    bullet style fields in a dictionary.
+
+    Parameters
+    ----------
+    readme_path : str
+        `/vsis3/.../README.txt` or any other GDAL compatible URI.
```
gmgunter: The function name and docstring mention the "README.txt" file, but actually this function is only able to properly parse the "LICENSE.txt" file.

The license file contents are a bulleted list, e.g.

```
- Name: NISAR water mask (EPSG 4326)
- Short description: Water Mask for the NASA-ISRO NISAR mission version 0.5, derived through analysis and processing of multiple input datasets, referenced to the WGS84 ellipsoid.
- EPSG code: 4326
- Version: 0.5
- Processing date: 2025-06-11
- Notes:
Water mask derived at the Jet Propulsion Laboratory, California Institute of Technology, under contract with NASA, using multiple open datasets: JRC Global Surface Water v1.3 (Pekel et al 2016, Nature 540, 418-422, © EC JRC/Google); ESA WorldCover 10 m 2020 (© ESA WorldCover project 2020 / Contains modified Copernicus Sentinel data (2020) processed by ESA WorldCover consortium); OpenStreetMap Water Polygons (© OpenStreetMap contributors, ODbL 1.0); MEaSUREs GIMP Mask v1 (Howat 2017, doi:10.5067/B8X58MQBFUPA; Howat et al 2014). All trademarks and rights remain with their respective owners; NISAR mission organizations assume no liability for use of this product.
- Source Data:
JRC Global Surface Water v1.3 (doi:10.1038/nature20584) © EC JRC/Google; ESA WorldCover 10 m 2020 v100 (doi:10.5281/zenodo.5571936) © ESA WorldCover project 2020 / Contains modified Copernicus Sentinel data (2020) processed by ESA WorldCover consortium; OpenStreetMap Water Polygons (2023-10-01) © OpenStreetMap contributors, ODbL 1.0; MEaSUREs GIMP Land Ice & Ocean Classification Mask v1 (doi:10.5067/B8X58MQBFUPA) © NASA NSIDC.
```

whereas the readme file contents have a different structure that's not handled by this function, e.g.

```
Water Mask for the NASA-ISRO NISAR mission version 0.5, derived through analysis and processing of multiple input datasets, referenced to the WGS84 ellipsoid.
distance-to-water classification
0 : non-water / background
1-100 : ocean-shoreline distance in 1 km bins value
    1 → 0-1 km from coastline value
    2 → 1-2 km from coastline
    …
    value 100 → 99-100 km from coastline
101-200 : inland-water distance (lake & river shore) in 1 km bins value
    101 → 0-1 km from lake/river shoreline value
    102 → 1-2 km from lake/river shoreline
    …
    value 200 → 99-100 km from lake/river shoreline
201-254 : reserved (currently unused)
255 : NoData (transparent)

Distances are Euclidean in the raster’s projected CRS.
```

Let's rename the function and update the docstring to make clear that it's intended for parsing the license file -- not the readme file.

python/packages/nisar/workflows/stage_watermask.py
gmgunter: It looks like the `LICENSE.txt` file was first introduced in v0.4? Should we add a comment here noting that this file is not expected in versions below 0.4?

python/packages/nisar/workflows/stage_watermask.py
```diff
@@ -501,6 +597,9 @@ def apply_margin_polygon(polygon, margin_in_km=5):
     lat_margin = margin_km_to_deg(margin_in_km)
     lon_margin = margin_km_to_longitude_deg(margin_in_km, lat=lat_worst_case)
 
+    if lon_max - lon_min > 180:
+        lon_min, lon_max = lon_max, lon_min
```
gmgunter, line 604: :eyes: What's happening here?

python/packages/nisar/workflows/stage_watermask.py
```diff
@@ -587,7 +686,7 @@ def main(opts):
     else:
         # Check connection to AWS s3 nisar-WATERMASK  ucket
```
gmgunter:
```suggestion
        # Check connection to AWS s3 nisar-WATERMASK bucket
```

python/packages/nisar/workflows/stage_watermask.py
```diff
@@ -455,7 +551,7 @@ def check_watermask_overlap(watermaskFilepath, polys):
     return perc_area
 
 
-def check_aws_connection():
+def check_aws_connection(version='0.5'):
```
gmgunter: WDYT about removing the default value? That way, if we release new versions of the Water Mask Product, we won't need to remember to update this.

```suggestion
def check_aws_connection(version):
```

python/packages/nisar/workflows/stage_watermask.py
```diff
@@ -587,7 +686,7 @@ def main(opts):
     else:
         # Check connection to AWS s3 nisar-WATERMASK  ucket
         try:
-            check_aws_connection()
+            check_aws_connection(opts.version)
```
gmgunter, line 692: Should we update the default value of `opts.version` to v0.5? Currently it defaults to v0.3.

https://github.com/isce-framework/isce3/blob/607825a8ae7d88dcc6571e1b05d262bf4e2e949a/python/packages/nisar/workflows/stage_watermask.py#L46-L48

python/packages/nisar/workflows/stage_watermask.py
```diff
@@ -366,19 +382,99 @@ def download_watermask(polys, epsgs, outfile, version):
...
+            # Continuation of current bullet (indentation or empty line)
+            if curr_key is not None:
+                curr_val_lines.append(line.rstrip())
+
+    # Capture the final field
+    if curr_key is not None:
+        fields[curr_key] = "\n".join(curr_val_lines).strip()
+
+    if not fields:
+        raise ValueError("No bullet‑style fields detected in README.txt")
```
gmgunter:
```suggestion
        raise ValueError(f"No bullet‑style fields detected in {readme_path}")
```

gmgunter (commented):
Thanks!

gmgunter:
Ultimately, what I think we want is a single text blurb that gives a basic description of the Water Mask and its license info, similar to the description found in the DEM product:

```
The NISAR DEM has been prepared at the Jet Propulsion Laboratory,
California Institute of Technology, under contract with the National
Aeronautics and Space Administration, using the Copernicus DEM 30m
COP-DEM_GLO-30-DGED/2023_1 model provided by the European Space
Agency. The Copernicus DEM 30m was produced using Copernicus
WorldDEM-30 © DLR e.V. 2010-2014 and © Airbus Defence and Space GmbH
2014-2018 provided under COPERNICUS by the European Union and ESA;
all rights reserved. The organizations in charge of the NISAR
mission by law or by delegation do not incur any liability for any
use of this DEM. The organizations in charge of the Copernicus
programme by law or by delegation do not incur any liability for any
use of the Copernicus WorldDEM-30. Users, including those who
redistribute, adapt, modify, or combine NISAR DEM data, must comply
with the terms of the Copernicus DEM 30m License Agreement. For
additional information, please refer to:
https://www.doi.org/2010.5067/NIDEM-1 and
https://doi.org/10.5270/ESA-c5d3d65.
```

That way, the InSAR and Static Layers workflows could extract this blurb from the input Water Mask file and copy it into the `waterMaskSource` dataset in their respective output GUNW and Static Layers HDF5 products.

In the case of `stage_dem.py`, there is a single metadata item that contains the full description of the product and its license.

https://github.com/isce-framework/isce3/blob/607825a8ae7d88dcc6571e1b05d262bf4e2e949a/python/packages/nisar/workflows/stage_dem.py#L386

However, in the case of `stage_watermask.py`, it looks like we're adding multiple metadata items to the VRT file.

https://github.com/isce-framework/isce3/blob/607825a8ae7d88dcc6571e1b05d262bf4e2e949a/python/packages/nisar/workflows/stage_watermask.py#L393-L397

Is there some way for the InSAR and Static Layers workflows to extract these metadata items from the VRT and convert them into a single string that we can place in the HDF5 product?

It might be easier for these downstream workflows if there was just a single metadata item that they could read from the VRT file, similar to the `dem_description` in the DEM product.

oberonia78:
@gmgunter TThank you for the detailed feedback. I agree that storing the license text as a single string in the HDF5 product is the right approach. 
To make sure I understand your recommendation correctly, I ran `stage_dem.py` and noticed that the staged VRT includes this metadata:
```
  dem_description=Digital Elevation Model (DEM) for the NASA-ISRO NISAR mission, version 1.2, based on the Copernicus DEM 30-m (2023_1), referenced to the WGS84 ellipsoid.
```
in metadata field. This string is much shorter than the original license text and differs from what you suggested. Could you confirm which DEM product and license statement you’d like us to include?

Regarding  `stage_watermask.py`, the script currently copies all info from `LICENSE.txt`. 
```
  Name=NISAR water mask (EPSG 4326)
  Notes=Water mask derived at the Jet Propulsion Laboratory, California Institute of Technology, under contract with NASA, using multiple open datasets: JRC Global Surface Water v1.3 (Pekel et al 2016, Nature 540, 418-422, © EC JRC/Google); ESA WorldCover 10 m 2020 (© ESA WorldCover project 2020 / Contains modified Copernicus Sentinel data (2020) processed by ESA WorldCover consortium); OpenStreetMap Water Polygons (© OpenStreetMap contributors, ODbL 1.0); MEaSUREs GIMP Mask v1 (Howat 2017, doi:10.5067/B8X58MQBFUPA; Howat et al 2014). All trademarks and rights remain with their respective owners; NISAR mission organizations assume no liability for use of this product.
  Processing date=2025-06-11
  Short description=Water Mask for the NASA-ISRO NISAR mission version 0.5, derived through analysis and processing of multiple input datasets, referenced to the WGS84 ellipsoid.
  Source Data=JRC Global Surface Water v1.3 (doi:10.1038/nature20584) © EC JRC/Google; ESA WorldCover 10 m 2020 v100 (doi:10.5281/zenodo.5571936) © ESA WorldCover project 2020 / Contains modified Copernicus Sentinel data (2020) processed by ESA WorldCover consortium; OpenStreetMap Water Polygons (2023-10-01) © OpenStreetMap contributors, ODbL 1.0; MEaSUREs GIMP Land Ice & Ocean Classification Mask v1 (doi:10.5067/B8X58MQBFUPA) © NASA NSIDC.
```
I’m leaning toward using the `Notes` line as a concise license description. Do you think that would be sufficient, or would you prefer more detail?

gmgunter:
> I ran `stage_dem.py` and noticed that the staged VRT includes this metadata:
> ```
>   dem_description=Digital Elevation Model (DEM) for the NASA-ISRO NISAR mission, version 1.2, based on the Copernicus DEM 30-m (2023_1), referenced to the WGS84 ellipsoid.
> ```
> in metadata field. This string is much shorter than the original license text and differs from what you suggested. Could you confirm which DEM product and license statement you’d like us to include?

Thanks for testing that, @oberonia78.

@gshiroma is this what you would expect the `dem_description` in the DEM VRT file to look like?

Based on our offline discussion in the ISCE3 meeting last week, my understanding was that PGE/PCM will extract this string from the VRT file and pass it to the SAS via the `dem_file_description` field in the runconfig.

The short DEM description above doesn't include any license information, so if PGE/PCM passes this string to the SAS workflows, then downstream NISAR products will also be missing this license info.

Here's an example README.txt from the 'nisar-dem' S3 bucket:

```
- Name: Copernicus DEM for NISAR v1.2 (EPSG 4326)
- Short description: Digital Elevation Model (DEM) for the NASA-ISRO NISAR mission, version 1.2, based on the Copernicus DEM 30-m (2023_1), referenced to the WGS84 ellipsoid.
- DOI: 10.5067/NIDEM-1
- EPSG code: 4326
- Vertical reference level: WGS84 ellipsoid
- Version: 1.2
- Processing date: 2024-10-30
- Notes: This digital elevation model (DEM) was prepared at the Jet Propulsion Laboratory, California Institute of Technology, under contract with the National Aeronautics and Space Administration, using the Copernicus DEM 30-m COP-DEM_GLO-30-DGED/2023_1 model, provided by the European Space Agency (ESA). The Copernicus DEM 30-m was produced using Copernicus WorldDEM-30 © DLR e.V. 2010-2014 and © Airbus Defence and Space GmbH 2014-2018 provided under COPERNICUS by the European Union and ESA; all rights reserved. The organizations in charge of the NISAR mission or the Copernicus programme by law or by delegation do not incur any liability for any use of the Copernicus WorldDEM-30. Users, including those who redistribute, adapt, modify, or combine the data contained in this DEM, must comply with the terms of the Copernicus DEM 30-m License Agreement. For additional information, please refer to https://doi.org/10.5067/NIDEM-1 and https://doi.org/10.5270/ESA-c5d3d65.
- Source Data: Copernicus DEM 30-m GLO-30 (COP-DEM_GLO-30-DGED/2023_1). DOI: https://doi.org/10.5270/ESA-c5d3d65. © DLR e.V. 2010-2014 and © Airbus Defence and Space GmbH 2014-2018 provided under COPERNICUS by the European Union and ESA; all rights reserved.
- Copernicus DEM License: LICENSE_COP-DEM-GLO_DGED__2023_1_eula_F.pdf [Local Relative Path]. Available online at https://doi.org/10.5270/ESA-c5d3d65.
- Other Processing Information: The vertical reference level of all Copernicus DEM tiles was updated from the EGM2008 geoid to the WGS84 ellipsoid. Non-existing Copernicus DEM 30-m tiles (e.g., over ocean) were filled with the EGM2008 geoid model, resulting in a DEM with complete global coverage. A hierarchical VRT (Virtual Raster) structure was created to enable fast and unified access to the entire DEM. The top-level file, EPSG4326.vrt, points to a second tier of VRTs organized by latitude bands. These, in turn, reference a third level of VRT files that map to the individual GeoTIFF DEM tiles.
```

It looks like `stage_dem.py` is copying the "Short description" section into the VRT. Maybe it'd be better to copy the "Notes" section instead?

https://github.com/isce-framework/isce3/blob/70b4a3faf9137c804ced2182c9574f6d684bb3bd/python/packages/nisar/workflows/stage_dem.py#L409

gmgunter:
>I’m leaning toward using the `Notes` line as a concise license description. Do you think that would be sufficient, or would you prefer more detail?

That sounds good to me :+1:

python/packages/nisar/workflows/stage_watermask.py
```diff
@@ -382,43 +382,45 @@ def download_watermask(polys, epsgs, outfile, version):
...
+                )
+        else:
             warnings.warn(
-                f"Could not read or parse README.txt at {in_readme_path} "
-                f"({e}). Proceeding without metadata."
+                f"No LICENSE.txt found at {in_license_path} "
+                f"(expected for WATERMASK v{version}). Proceeding without metadata."
             )
-
     except Exception:
```
gmgunter:
```suggestion
    except Exception as e:
```

Needed for this line below:

https://github.com/isce-framework/isce3/blob/60d9072af860638651365bd47974d5fad8b40166/python/packages/nisar/workflows/stage_watermask.py#L411

python/packages/nisar/workflows/stage_watermask.py
```diff
@@ -428,24 +430,39 @@ def parse_readme_fields(
...
-    fp = gdal.VSIFOpenL(readme_path, "rb")
+    stat = gdal.VSIStatL(license_path)
+    if stat is None:
+        raise ValueError(f"No LICENSE.txt found at {license_path}")
+
+    MAX_SIZE = 10_000_000 # 10 MB sanity limit
+    if stat and stat.size > MAX_SIZE:
+        warnings.warn(
+            f"{license_path} is {stat.size/1_000_000:.1f} MB—"
+            "skipping LICENSE parsing."
```
gmgunter: What do you mean by "skipping LICENSE parsing"? This seems misleading. It doesn't look like we skip parsing the file in this case. We just parse the first 10,000,000 bytes.

```suggestion
            f"{license_path} is {stat.size/1_000_000:.1f} MB—"
            " only the first 10 MB will be parsed."
```

gshiroma:
> > I ran `stage_dem.py` and noticed that the staged VRT includes this metadata:
> > ```
> >   dem_description=Digital Elevation Model (DEM) for the NASA-ISRO NISAR mission, version 1.2, based on the Copernicus DEM 30-m (2023_1), referenced to the WGS84 ellipsoid.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > in metadata field. This string is much shorter than the original license text and differs from what you suggested. Could you confirm which DEM product and license statement you’d like us to include?
> 
> Thanks for testing that, @oberonia78.
> 
> @gshiroma is this what you would expect the `dem_description` in the DEM VRT file to look like?
> 
> Based on our offline discussion in the ISCE3 meeting last week, my understanding was that PGE/PCM will extract this string from the VRT file and pass it to the SAS via the `dem_file_description` field in the runconfig.
> 
> The short DEM description above doesn't include any license information, so if PGE/PCM passes this string to the SAS workflows, then downstream NISAR products will also be missing this license info.
> 
> Here's an example README.txt from the 'nisar-dem' S3 bucket:
> 
> ```
> - Name: Copernicus DEM for NISAR v1.2 (EPSG 4326)
> - Short description: Digital Elevation Model (DEM) for the NASA-ISRO NISAR mission, version 1.2, based on the Copernicus DEM 30-m (2023_1), referenced to the WGS84 ellipsoid.
> - DOI: 10.5067/NIDEM-1
> - EPSG code: 4326
> - Vertical reference level: WGS84 ellipsoid
> - Version: 1.2
> - Processing date: 2024-10-30
> - Notes: This digital elevation model (DEM) was prepared at the Jet Propulsion Laboratory, California Institute of Technology, under contract with the National Aeronautics and Space Administration, using the Copernicus DEM 30-m COP-DEM_GLO-30-DGED/2023_1 model, provided by the European Space Agency (ESA). The Copernicus DEM 30-m was produced using Copernicus WorldDEM-30 © DLR e.V. 2010-2014 and © Airbus Defence and Space GmbH 2014-2018 provided under COPERNICUS by the European Union and ESA; all rights reserved. The organizations in charge of the NISAR mission or the Copernicus programme by law or by delegation do not incur any liability for any use of the Copernicus WorldDEM-30. Users, including those who redistribute, adapt, modify, or combine the data contained in this DEM, must comply with the terms of the Copernicus DEM 30-m License Agreement. For additional information, please refer to https://doi.org/10.5067/NIDEM-1 and https://doi.org/10.5270/ESA-c5d3d65.
> - Source Data: Copernicus DEM 30-m GLO-30 (COP-DEM_GLO-30-DGED/2023_1). DOI: https://doi.org/10.5270/ESA-c5d3d65. © DLR e.V. 2010-2014 and © Airbus Defence and Space GmbH 2014-2018 provided under COPERNICUS by the European Union and ESA; all rights reserved.
> - Copernicus DEM License: LICENSE_COP-DEM-GLO_DGED__2023_1_eula_F.pdf [Local Relative Path]. Available online at https://doi.org/10.5270/ESA-c5d3d65.
> - Other Processing Information: The vertical reference level of all Copernicus DEM tiles was updated from the EGM2008 geoid to the WGS84 ellipsoid. Non-existing Copernicus DEM 30-m tiles (e.g., over ocean) were filled with the EGM2008 geoid model, resulting in a DEM with complete global coverage. A hierarchical VRT (Virtual Raster) structure was created to enable fast and unified access to the entire DEM. The top-level file, EPSG4326.vrt, points to a second tier of VRTs organized by latitude bands. These, in turn, reference a third level of VRT files that map to the individual GeoTIFF DEM tiles.
> ```
> 
> It looks like `stage_dem.py` is copying the "Short description" section into the VRT. Maybe it'd be better to copy the "Notes" section instead?
> 
> https://github.com/isce-framework/isce3/blob/70b4a3faf9137c804ced2182c9574f6d684bb3bd/python/packages/nisar/workflows/stage_dem.py#L409

@gmgunter ,yes, this was meant to be just a description of the DEM, not a license, since we're not redistributing the DEM in these products (which is different than STATIC produts). However, we've since learned that we should include statements clarifying that the Copernicus DEM was used to generate the product. So, we’ll likely need to update `stage_dem.py` to include those statements. Thank you for bringing this up.

Including @hfattahi  and @vbrancat in this discussion.

gmgunter (commented):
Few more small suggestions

gmgunter:
>this was meant to be just a description of the DEM, not a license, since we're not redistributing the DEM in these products (which is different than STATIC produts).

Thanks, @gshiroma.

Based on our offline discussion about the `demSource` and `waterMaskSource` datasets in the Static Layers product (summarized [here](https://github-fn.jpl.nasa.gov/NISAR-ADT/NISAR_PIX/pull/283#discussion_r27395)), I thought our intention was that the `demSource` dataset contents would be the same in all NISAR products and would include both a description of the input product and the license info.

I think it's probably feasible for this dataset to be different in the Static Layers product vs the other products (I can just tell the On-Demand team exactly what to put in the runconfig instead of telling them to extract this string from the DEM VRT file). But I wonder if it'd be better if all the products had the same (longer) string that includes the license info? That way, the same logic will be consistent for both the DEM and the Water Mask across all products.

vbrancat (approved):
Thank you @jungkyoJung, LGTM!

vbrancat:
@gmgunter and @jungkyoJung can you work together and get this PR merged? It seems that there are no major points to address. Thanks

python/packages/nisar/workflows/stage_watermask.py
oberonia78: I copied these lines from `stage_dem.py`.

gmgunter: Fair enough -- I agree we should keep the behavior consistent with `stage_dem.py`.

I'll open an Issue about this. Maybe we can add a comment eventually explaining what's going on, but let's not worry about it in this PR.

gmgunter:
Let's update the behavior to concatenate the `Short description` + `Notes` sections, instead of just using the `Notes` section. This will be consistent with the behavior implemented for `stage_dem.py` here: https://github.com/isce-framework/isce3/pull/88. It will ensure that the VRT includes information about the version number of the water mask as well as the license info.

python/packages/nisar/workflows/stage_watermask.py
```diff
@@ -367,18 +383,101 @@ def download_watermask(polys, epsgs, outfile, version):
...
+        # Get the WATER description from the LICENSE.txt file using GDAL
+        in_license_path = vrt_filename.replace(f'EPSG{epsg}.vrt',
+                                               f'EPSG{epsg}/LICENSE.txt')
+        license_exists = gdal.VSIStatL(in_license_path) is not None
+
+        if license_exists:
+            try:
+                water_descr = parse_license_notes(in_license_path)
+                vrt_dataset.SetMetadataItem("water_mask_description",
+                                            f'{water_descr}')
```
gmgunter: It should already be a string, so no need for this

```suggestion
                                            water_descr)
```

python/packages/nisar/workflows/stage_watermask.py
```diff
@@ -367,18 +383,101 @@ def download_watermask(polys, epsgs, outfile, version):
...
+
+            except Exception as e:
+                warnings.warn(
+                    f"Found {in_license_path} but could not parse it ({e}). "
+                    "Proceeding without metadata."
+                )
+        else:
+            warnings.warn(
+                f"No LICENSE.txt found at {in_license_path} "
+                f"(expected for WATERMASK v{version}). Proceeding without metadata."
```
gmgunter:
```suggestion
                f"(expected for WATERMASK v0.4+). Proceeding without metadata."
```

python/packages/nisar/workflows/stage_watermask.py
```diff
@@ -367,18 +383,101 @@ def download_watermask(polys, epsgs, outfile, version):
...
                  f'Maybe {version} is not currently supported.'
-        raise ValueError(errmsg)
+        raise ValueError(errmsg) from e
+
+
+def parse_license_notes(license_path: str,
+                        encoding: str = "utf-8") -> str | None:
+    """
+    Extract the multi-line text of the “Notes” bullet from
+    a LICENSE.txt in an S3 bucket via GDAL and returning the
```
gmgunter:
```suggestion
    a LICENSE.txt in an S3 bucket via GDAL and return the
```

python/packages/nisar/workflows/stage_watermask.py
```diff
@@ -367,18 +383,101 @@ def download_watermask(polys, epsgs, outfile, version):
...
+    license_path : str
+        `/vsis3/.../LICENSE.txt` or any other GDAL compatible URI.
+    encoding : str, default "utf-8"
+        Text encoding to decode the bytes.
+
+    Returns
+    -------
+    str | None
+        The extracted Notes text (possibly multi-line) with surrounding
+        whitespace stripped, or None if no “Notes” bullet is found.
```
gmgunter, line 433: Let's raise an exception if no "Notes" bullet was found, instead of returning `None`. In that case, the calling code will catch the exception, emit a warning, and proceed without metadata.

python/packages/nisar/workflows/stage_watermask.py
```diff
@@ -367,18 +383,101 @@ def download_watermask(polys, epsgs, outfile, version):
...
+    if stat.size and stat.size > MAX_SIZE:
+        warnings.warn(
+            f"{license_path} is {stat.size/1_000_000:.1f} MB — "
+            f"only the first {MAX_SIZE/1_000_000:.0f} MB will be parsed."
+        )
+    fp = gdal.VSIFOpenL(license_path, "rb")
+    if fp is None:
+        raise IOError(f"Could not open {license_path} via GDAL")
+    try:
+        data = gdal.VSIFReadL(1, min(stat.size, MAX_SIZE), fp)
```
gmgunter: Looks like we could simplify this and just read `stat.size` bytes from the file -- no need for `MAX_SIZE` anymore.

```suggestion
    fp = gdal.VSIFOpenL(license_path, "rb")
    if fp is None:
        raise IOError(f"Could not open {license_path} via GDAL")
    try:
        data = gdal.VSIFReadL(1, stat.size, fp)
```

python/packages/nisar/workflows/stage_watermask.py
```diff
@@ -367,18 +383,101 @@ def download_watermask(polys, epsgs, outfile, version):
...
+        warnings.warn(f'No "Notes" field found in {license_path}')
+        return None
+
+    NEXT_BULLET = r"(?=^\s*-\s+[^:\n]+:\s*|\Z)"
+    NOTES_RE = re.compile(
+        rf"^\s*-\s*Notes:\s*(.*?){NEXT_BULLET}",
+        flags=re.IGNORECASE | re.MULTILINE | re.DOTALL)
+
+    m = NOTES_RE.search(text)
+    if not m:
```
gmgunter:
```suggestion
    if m is None:
```

gmgunter (commented):
Reviewed in PR Fridays 2025-08-29

oberonia78:
Hi @gmgunter, Thank you for reviewing the PR. Would you be able to review one more time? I just finished revising this based on your comments.

oberonia78:
Updated code concatenates `short description` and `note`.
```
  water_mask_description=Water Mask for the NASA-ISRO NISAR mission version 0.5, derived through analysis and processing of multiple input datasets, referenced to the WGS84 ellipsoid. Water mask derived at the Jet Propulsion Laboratory, California Institute of Technology, under contract with NASA, using multiple open datasets: JRC Global Surface Water v1.3 (Pekel et al 2016, Nature 540, 418-422, © EC JRC/Google); ESA WorldCover 10 m 2020 (© ESA WorldCover project 2020 / Contains modified Copernicus Sentinel data (2020) processed by ESA WorldCover consortium); OpenStreetMap Water Polygons (© OpenStreetMap contributors, ODbL 1.0); MEaSUREs GIMP Mask v1 (Howat 2017, doi:10.5067/B8X58MQBFUPA; Howat et al 2014). All trademarks and rights remain with their respective owners; NISAR mission organizations assume no liability for use of this product.
```

python/packages/nisar/workflows/stage_watermask.py
```diff
@@ -461,16 +454,23 @@ def parse_license_notes(license_path: str,
...
-    if not m:
-        warnings.warn(f'No "Notes" field found in {license_path}')
-        return None
+    def _grab(header: str) -> str | None:
+        pat = re.compile(
+            rf"^\s*-\s*{re.escape(header)}\s*:\s*(.*?){NEXT_BULLET}",
+            flags=re.IGNORECASE | re.MULTILINE | re.DOTALL,
+        )
+        m = pat.search(text)
+        return m.group(1).strip() if m else None
```
gmgunter:
```suggestion
        return m.group(1).strip() if (m is not None) else None
```

gmgunter (approved):
One small nitpick -- otherwise, LGTM. Nice work! Thank you!

PR #78, opened 2025-07-29
Initialize maxval to lowest float for correct peak detection consistenc…

### LINKED ISSUE #62: Abnormal offset in CPU Ampcor.
**Description:**

During InSAR processing using the ISCE3 stack, we encountered a strange issue in the layers that they have different coverages cross the layers. 

Image

In the unwrapped phase, the bottom region unexpectedly contains zero values.
The coherence layer shows NaN values in the same region.
The mask layer, however, labels these regions as valid, suggesting they should be processed normally.
Initial suspicion was non-overlapping SLCs, but this was ruled out after visualizing the coregistered SLCs.

**Investigation and Findings:**
Image

The issue appears to originate from dense offset estimation using CPU Ampcor:
Extremely large offset values (on the order of 10⁶) are observed at the top of the image in both range and azimuth directions.
These extreme values propagate through interpolation, affecting regions like ocean areas that should be decorrelated.

In contrast, GPU Ampcor produces reasonable offset values:
For the same region, GPU Ampcor outputs consistent values (e.g., -16), and final coregistration via rubbersheeting works correctly.

Due to this issue, we are suspecting that the resampled image has zero values over the ocean area, unexpectedly. 

Image

**Runconfigs:**

`/scratch2/jungkyoj/iscale/hawaii/manual/runconfig_set4.yaml`

`/scratch2/jungkyoj/iscale/hawaii/manual/runconfig_set5.yaml`

at dev 5 machine.

vbrancat:
Closes #62 

**Background**
The goal of incoherent cross-correlation, also known as Ampcor, is to identify the peak value and the location of the cross-correlation surface. This information is propagated downstream to estimate the offset between time-separated SAR image pairs.

**Description of the Issue**
In the CPU implementation of incoherent cross-correlation, the variable used to track the maximum value of the correlation surface was incorrectly initialized to the smallest positive representable floating-point number. This causes problems when the correlation surface contains uniformly low values (e.g., exp(-38)). In such cases, none of the surface values are recognized as higher than the initialized `my_maxval`, which means `my_maxloc` i.e., the location of the peak of the correlation surface, either remains uninitialized or is incorrectly set.
This leads to extremely large and incorrect offset estimates. For example, Figure 1 shows this behavior for a 46-day ALOS/PALSAR RSLC image pair:

Peak

Figure 1a displays the peak of a correlation surface, where the upper region contains values around 1.18e-38. These low values, when incorrectly handled, result in offset estimates on the order of 1e38 (top-part of Figure 1b).

**Solution**
This PR corrects the initialization of variable holding the cross-correlation peak in the CPU code. This variable is initialized as: 

`float my_maxval = -std::numeric_limits<float>::max();`

This change ensures that any valid correlation surface value—regardless of how small—is correctly compared during peak detection. It also brings the CPU implementation in line with the CUDA version, which uses `-FLT_MAX `for the same purpose.

**Testing**

- Executed existing unit tests on a variety of correlation surfaces, including those with extremely low peak values.
- Verified that peak detection in the CPU implementation is now robust and produces accurate locations even for nearly flat surfaces.
- Confirmed that results now match the CUDA implementation for the same input data.

### DIFF
--- a/cxx/isce3/matchtemplate/pycuampcor/cuOffset.cpp
+++ b/cxx/isce3/matchtemplate/pycuampcor/cuOffset.cpp
@@ -21,7 +21,7 @@ void  cudaKernel_maxloc2D(const float* const images, int2* maxloc, float* maxval
     const int imageSize = imageNX * imageNY;
 
     for (int bid = 0; bid < nImages; bid++) {
-        float my_maxval = -std::numeric_limits<float>::max();
+        float my_maxval = std::numeric_limits<float>::lowest();
         int2 my_maxloc;
         const float* image = &images[bid * imageSize];
         for (int i = 0; i < imageSize; i++) {
--- a/cxx/isce3/matchtemplate/pycuampcor/cuOffset.cpp
+++ b/cxx/isce3/matchtemplate/pycuampcor/cuOffset.cpp
@@ -21,7 +21,7 @@ void  cudaKernel_maxloc2D(const float* const images, int2* maxloc, float* maxval
     const int imageSize = imageNX * imageNY;
 
     for (int bid = 0; bid < nImages; bid++) {
-        float my_maxval = std::numeric_limits<float>::min();
+        float my_maxval = -std::numeric_limits<float>::max();
         int2 my_maxloc;
         const float* image = &images[bid * imageSize];
         for (int i = 0; i < imageSize; i++) {


cxx/isce3/matchtemplate/pycuampcor/cuOffset.cpp
```diff
@@ -21,7 +21,7 @@ void  cudaKernel_maxloc2D(const float* const images, int2* maxloc, float* maxval
     const int imageSize = imageNX * imageNY;
 
     for (int bid = 0; bid < nImages; bid++) {
-        float my_maxval = std::numeric_limits<float>::min();
+        float my_maxval = -std::numeric_limits<float>::max();
```
rtburns-jpl:
```suggestion
        float my_maxval = std::numeric_limits<float>::lowest();
```

I don't think the negative of the max value is guaranteed to be a valid float value, so let's instead explicitly request the lowest possible float.

rtburns-jpl (approved):
Great work, thanks for investigating this. Changes look good!

gmgunter (approved):
Nice catch, thanks!

PR #79, opened 2025-08-01
Add product spec version to `ProductSpec` XML parser (MCR NOT NEEDED)

gmgunter:
I realized that the `ProductSpec` XML parser class is missing a way to get the product spec version. This is useful since each NISAR product includes the product spec version as a dataset in the product -- it's convenient to get this info directly from the XML file instead of manually updating the product spec version each time we update the XML file.

This PR uses a regex to parse the product spec version from a comment in the XML file.

### DIFF
--- a/python/packages/nisar/products/product_spec.py
+++ b/python/packages/nisar/products/product_spec.py
@@ -316,12 +316,16 @@ class ProductSpec:
         """
         Create a new `ProductSpec` object.
 
+        This constructor typically shouldn't be called directly. Instead, the
+        `from_file` and `from_string` methods provide convenient ways to create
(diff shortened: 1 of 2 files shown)


gmgunter:
@nemo794 @Tyler-g-hudson  this is a copy of a PR that you approved on the internal GitHub server: https://github-fn.jpl.nasa.gov/isce-3/isce/pull/2185

Tyler-g-hudson (approved):
LGTM!

nemo794 (approved):
Thank you for adding the edge case tests! LGTM. Thanks, @gmgunter !

PR #83, opened 2025-08-07
Allow Partitioning of DM-to-Science L0B

rad-eng-59:
This PR simply adds optional L0B partitioning capability to DM2-to-science workflow to speed up generation of large mosaicked (1-tap DBF) L0B product via embarrassingly parallel I/O over smaller chunks of a large DM2 L0B product (way beyond `240-km` along track).
Due to similarity to DM2 counterpart, the DM1-to-science workflow has also being updated in similar fashion.
Note that trivial instrument-related high-rate telemetry fields of input L0B product are not chunked but rather repeated in all sub-products.
Moreover, the bounding polygon and the original filename covering entire input L0B are preserved in all sub-products for the sake of convenient tracking of original product and DEM generation.
Simply start/end zero Doppler date-time associated with each part is reflected in both L0B identification as well as an extension to the original filename to distinguish various parts/chunks at the output properly.
I've got almost double speed-up of entire runtime by simply breaking up the `42-sec` DM2 L0B product w/ `RCID=34` into two `21-sec` science ones via two background jobs running on a single machine.

### DIFF
--- a/python/packages/nisar/workflows/nisar_l0b_dm2_to_dbf.py
+++ b/python/packages/nisar/workflows/nisar_l0b_dm2_to_dbf.py
@@ -7,6 +7,7 @@ from pathlib import Path
 import os
 import time
 import tempfile
+from copy import copy
 
 import numpy as np
 from scipy.signal import fftconvolve
@@ -25,14 +26,17 @@ from isce3.io import Raster
 from nisar.log import set_logger
 from isce3.signal import dbf_onetap_from_dm2, dbf_onetap_from_dm2_seamless
 from isce3.focus import fill_gaps, form_linear_chirp
+from isce3.core import TimeDelta
 from nisar.workflows.helpers import build_uniform_quantizer_lut_l0b, slice_gen
 from isce3.antenna import ant2rgdop
 
 
-def copy_swath_except_echo_h5(fid_in, fid_out, swath_path, frq_pol):
+def copy_swath_except_echo_h5(
+        fid_in, fid_out, swath_path, frq_pol, tx_keys=None, rx_keys=None):
     """
     Copy all groups and datasets under swath from input HDF5 to output
-    HDF5 except for echo products.
+    HDF5 except for echo and optionally for desired datasets under tx and
+    rx group listed by tx_keys and rx_keys.
 
     Parameters
     ----------
@@ -45,8 +49,21 @@ def copy_swath_except_echo_h5(fid_in, fid_out, swath_path, frq_pol):
     frq_pol : dict
         A dict of all frequency bands as keys and list of TxRx
         polarization as values.
+    tx_keys : sequence of str, optional
+        Sequence of dataset names to be excluded from tx group in `fid_out`.
+    rx_keys : sequence of str, optional
+        Sequence of dataset names to be excluded from rx group in `fid_out`.
+        Echo datasets are always excluded from RX group!
 
     """
+    # TX/RX dataset names to be excluded over all frequency bands and pols!
+    tx_names = []
+    if tx_keys is not None:
+        tx_names.extend(tx_keys)
+    rx_names = []
+    if rx_keys is not None:
+        rx_names.extend(rx_keys)
+
     for freq_band in frq_pol:
         # build band path
         band_path = os.path.join(swath_path, f'frequency{freq_band}/')
@@ -54,6 +71,10 @@ def copy_swath_except_echo_h5(fid_in, fid_out, swath_path, frq_pol):
         grp_band = fid_out.require_group(band_path)
         # list of all TxRx products
         txrx_pols = frq_pol[freq_band]
+        # form rx datset names plus echo for all pols to be
+        # excluded from output
+        rx_names_echo = copy(txrx_pols)
+        rx_names_echo.extend(rx_names)
         # list of group names txp where "p" is all TX pols
         txp = {f'tx{p[0]}' for p in txrx_pols}
         # list of group names rxp where "p" is all RX pols
@@ -71,11 +92,13 @@ def copy_swath_except_echo_h5(fid_in, fid_out, swath_path, frq_pol):
                         # create RX group for output product
                         grp_rx = fid_out.require_group(rx_path)
                         for rx_item in fid_in[rx_path]:
-                            if rx_item not in txrx_pols:
-                                # copy all RX fields except echo product
+                            if rx_item not in rx_names_echo:
+                                # copy all RX fields except
+                                # echo + rx_keys
                                 fid_in.copy(rx_path + rx_item, grp_rx)
-
-                    else:  # copy the rest of TX as it is
+                    # copy the rest of TX that is not in tx_keys
+                    elif ((tx_item not in tx_names) and
+                          ('validSamplesSubSwath' not in tx_item)):
                         fid_in.copy(tx_path + tx_item, grp_tx)
 
             else:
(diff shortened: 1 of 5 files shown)


python/packages/nisar/workflows/nisar_l0b_dm2_to_dbf.py
```diff
@@ -45,15 +51,32 @@ def copy_swath_except_echo_h5(fid_in, fid_out, swath_path, frq_pol):
...
+        Echo datasets are always excluded from RX group!
 
     """
+    # TX/RX dataset names to be excluded over all frequency bands and pols!
+    tx_names = []
+    if tx_keys is not None:
+        tx_names.extend(tx_keys)
+    rx_names = []
+    if rx_keys is not None:
+        rx_names.extend(rx_keys)
```
bhawkins, line 75:
```suggestion
    tx_names = [f"vadidSamplesSubSwath{n}" for n in "12345"]
    if tx_keys is not None:
        tx_names.extend(tx_keys)
    rx_names = [f"{tx}{rx}" for tx in "HVLR" for rx in "HV"]
    if rx_keys is not None:
        rx_names.extend(rx_keys)
```
No big deal, but we could simplify some of the logic below by just initializing the exclude lists.

python/packages/nisar/workflows/nisar_l0b_dm2_to_dbf.py
```diff
@@ -132,6 +157,137 @@ def create_echo_dataset_h5(fid_in, fid_out, band_path, txrx_pol,
...
+        of the new echo product
+    pulse_slice : slice
+        Desired pulse slice for input HDF5 file.
+    logger: logging.Logger
+    keys: sequence of str
+        Desired keys under RX group to be copied as long as
+        they exist in input hdf5.
+
+    """
+    path = os.path.join(band_path, f'tx{pol[0]}/')
```
bhawkins: Broken on windows, but I guess we do this elsewhere.

python/packages/nisar/workflows/nisar_l0b_dm2_to_dbf.py
```diff
@@ -132,6 +157,137 @@ def create_echo_dataset_h5(fid_in, fid_out, band_path, txrx_pol,
...
+        try:
+            dset_in = fid_in[path + name]
+        except KeyError as err:
+            logger.warning(
+                f'Missing dataset "{path + name}" in input L0B!'
+                f' Detailed Error -> "{err}"'
+            )
+            continue
+        else:
+            data_in = dset_in[pulse_slice.start:pulse_slice.stop]
```
bhawkins:
```suggestion
            data_in = dset_in[pulse_slice]
```
Simpler syntax and doesn't ignore possible stride.

python/packages/nisar/workflows/nisar_l0b_dm2_to_dbf.py
```diff
@@ -132,6 +157,137 @@ def create_echo_dataset_h5(fid_in, fid_out, band_path, txrx_pol,
...
+                name, shape=data_in.shape, dtype=data_in.dtype, data=data_in)
+            # copy attributes for the product from input
+            for a_name, a_val in dset_in.attrs.items():
+                dset_out.attrs[a_name] = a_val
+
+
+def copy_rx_datasets_truncated(
+        fid_in, fid_out, band_path, txrx_pol, pulse_slice, logger,
+        keys=('caltone', 'RD', 'WD', 'WL', 'basebandPhaseCorrection',
+              'TRMDataWindow', 'attenuation')
```
bhawkins, line 223: `TRMDataWindow` was removed from the spec a long time ago ([internal link](https://github-fn.jpl.nasa.gov/NISAR-ADT/NISAR_PIX/pull/80)), but I guess it doesn't hurt to have it here.

python/packages/nisar/workflows/nisar_l0b_dm2_to_dbf.py
```diff
@@ -236,11 +394,25 @@ def cmd_line_parser():
                            'pattern coverage of swath. A positive value used '
                            'only if `rx-antpat-calib`.')
                      )
+    prs.add_argument('--num-pulses-max', type=int,
+                     help=('Max number of pulses, >=1, starting from '
+                           '`start-pulse` to be processed. Default is all.')
+                     )
+    prs.add_argument('--start-pulse', type=int, default=1,
+                     help='Start pulse number, >=1, to be processed.'
+                     )
```
bhawkins: I'd prefer to start counting at zero. Maybe to avoid confusion we could call this argument `--pulse-offset` instead (with default=0).

bhawkins (approved):
LGTM. I don't consider any of the following comments as blockers.

I do think it'd be nice to crop the bounding polygon in order to simplify DEM queries. It's not too much code to implement, e.g.,
https://github.com/isce-framework/isce3/blob/7d15ed2f8dc52210bf0e7d1c274205cf3f9e37e4/share/nisar/examples/alos_to_nisar_l0b.py#L596-L607
The only tricky part is computing the Doppler LUT, but zero Doppler or a hard-coded 950 Hz would be better than nothing IMO.

python/packages/nisar/workflows/nisar_l0b_dm2_to_dbf.py
hfattahi: Thanks @rad-eng-59 for the quick turn around. Would you please remind me why don't we crop based on azimuth time instead of line index?

rad-eng-59:
Note that the last two commits captures new but similar updates in *DM1-to-science* workflow. Given the similarity of these two workflows, both DM2/DM1 updates are included in this PR to save time. 
If there is a strong objection, I can rebase the branch and/or remove those DM1-related updates.

python/packages/nisar/workflows/nisar_l0b_dm2_to_dbf.py
rad-eng-59: To preserve continuity of non-overlap range lines of several parts generated from a single L0B in both DM2 and DM1.
Using time units, requires exact floating point time stamps separated by **1 / PRF**. 
Instead, using range lines both facilitates the partition process and avoids any missing or duplicated range lines in multiple partitions to be eventually merged into a single continuous RSLC.

rad-eng-59:
> LGTM. I don't consider any of the following comments as blockers.
> 
> I do think it'd be nice to crop the bounding polygon in order to simplify DEM queries. It's not too much code to implement, e.g.,
> 
> https://github.com/isce-framework/isce3/blob/7d15ed2f8dc52210bf0e7d1c274205cf3f9e37e4/share/nisar/examples/alos_to_nisar_l0b.py#L596-L607
> 
> The only tricky part is computing the Doppler LUT, but zero Doppler or a hard-coded 950 Hz would be better than nothing IMO.

I did that on purpose. It's not like I couldn't do it. 
Actually one doesn't need Doppler LUT nor needs to hardcode any Doppler. The mechanical pointing and eletrcial boresight plus footprint all can be obtained from radar pointing and antenna pattern HDF5 product. Let's talk about it in the CAL meeting. Thanks.

python/packages/nisar/workflows/nisar_l0b_dm2_to_dbf.py
hfattahi: I see. sounds like your concern would be possibility for missing or double counting one rangeline. However, cropping with time would be easier from scaling perspective. Its not a big deal for this PR as we are handling the processing outside PCM, but for workflows within PCM we may want to use time.

hfattahi:
> > LGTM. I don't consider any of the following comments as blockers.
> > I do think it'd be nice to crop the bounding polygon in order to simplify DEM queries. It's not too much code to implement, e.g.,
> > https://github.com/isce-framework/isce3/blob/7d15ed2f8dc52210bf0e7d1c274205cf3f9e37e4/share/nisar/examples/alos_to_nisar_l0b.py#L596-L607
> > 
> > The only tricky part is computing the Doppler LUT, but zero Doppler or a hard-coded 950 Hz would be better than nothing IMO.
> 
> I did that on purpose. It's not like I couldn't do it. Actually one doesn't need Doppler LUT nor needs to hardcode any Doppler. The mechanical pointing and eletrcial boresight plus footprint all can be obtained from radar pointing and antenna pattern HDF5 product. Let's talk about it in the CAL meeting. Thanks.

If by any chance we will end up with a very long swath that also covers high latitudes, then the large polygon will lead to a very big DEM. I agree with @bhawkins that it would be nice to recompute the polygon unless if you have a good reason not to do so.

python/packages/nisar/workflows/nisar_l0b_dm2_to_dbf.py
Tyler-g-hudson: I agree that it would be simpler to handle this via azimuth time since this is the unit that users of this workflow are likely to be aware of and concerned about, and because it doesn't require potentially buggy arithmetic and remote file access for scaling users who want to determine where to make the cut. From the outside looking in, it's not clear to me how many range lines are present in a file or how much time a given number of them corresponds to. It's preferable to me that users would have an interface option for azimuth time, and would simply be expected to add a fraction of a second of margin to their times.

hfattahi: @bhawkins if multiple L0Bs fed to focusing module have overlap in azimuth, how would RSLC handle that? (Asking because @Tyler-g-hudson mentioned margin)

hfattahi: We discussed how RSLC handles L0B with overlaps and the answer is the data will be duplicated. i.e., we must not have overlap in L0Bs.

hfattahi (approved):
LGTM. Thanks for changing to time instead of range index. Makes life easier.

bhawkins (approved):
LGTM

PR #84, opened 2025-08-08
Replace raise Error with info_channel in remove_small_components

oberonia78:
An ionosphere script `ionosphere_filter.py` includes a function `remove_small_components`. This excludes the clusters with smaller number of pixels than threshold to minimize the contaminations to ionosphere. It raised an error, when all total number of valid pixel are smaller than the threshold because it will results in empty array as an output. 

However, this may be issues in some examples such as ocean dominated scene. Also, in ionosphere, even though the interferogram has noise we need to produce the results without raising issue.

This PR replaces raisingError with info_channel to avoid this issue.


python/packages/isce3/atmosphere/ionosphere_filter.py
```diff
@@ -952,13 +955,13 @@ def remove_small_components(image, min_cluster_pixels):
     n_valid = int(valid_mask.sum())
 
     if min_cluster_pixels == 0:
-        print("[remove_small_components] min_cluster_pixels == 0 → "
-              "no components will be removed.")
+        info_channel.log("[remove_small_components] min_cluster_pixels == 0 → "
+                         "no components will be removed.")
```
vbrancat: It is more appropriate to give a warning to the user instead of logging an info.
Please, use the `warning` channel.

python/packages/isce3/atmosphere/ionosphere_filter.py
```diff
@@ -952,13 +955,13 @@ def remove_small_components(image, min_cluster_pixels):
...
         return image.copy()
 
     if min_cluster_pixels >= n_valid:
-        raise ValueError(
-            f"min_cluster_pixels ({min_cluster_pixels}) must be smaller than "
+        info_channel.log(
+            f"min_cluster_pixels ({min_cluster_pixels}) are lager than "
             f"the total number of valid pixels ({n_valid})."
         )
```
vbrancat:
```suggestion
        info_channel.log(
            f"min_cluster_pixels ({min_cluster_pixels}) are larger than "
            f"the total number of valid pixels ({n_valid})."
        )
```

Typo. Also here, it is more `appropriate` to use a `warning` channel.

share/nisar/defaults/insar.yaml
```diff
@@ -297,7 +297,7 @@ runconfig:
                     # whose size is strictly smaller than `min_cluster_pixels`.
                     # When `min_cluster_pixels` is 0, this pre-removal step is
                     # skipped entirely.
-                    min_cluster_pixels: 5
+                    min_cluster_pixels: 2
```
vbrancat, line 305: How did you choose this threshold?

vbrancat (commented):
Thank you, @jungkyoJung. Just some minor comments

xhuang-jpl (approved):
As per discussed with @jungkyoJung  in the InSAR meeting, this PR looks good to me.

python/packages/isce3/atmosphere/ionosphere_filter.py
oberonia78: @vbrancat I agree with you. `warning` is better than `info` in this sentence. and thank you for picking the typo here :)

python/packages/isce3/atmosphere/ionosphere_filter.py
oberonia78: @vbrancat I think the info channel is fine for this line. Users can intentionally set `min_cluster_pixel` so that no pixels are removed.

share/nisar/defaults/insar.yaml
oberonia78: `min_cluster_pixels` determines which clusters are masked out during the ionosphere filter. Before applying the filter, invalid regions are masked from the unwrapped interferogram based on coherence or water, depending on the input. For example, if the invalid region is defined by a coherence threshold of 0.15, then all unwrapped phase pixels have coherence above 0.15. In this case, clusters smaller than min_cluster_pixels will be removed to reduce potential phase jumps.

By default, the ionosphere filter uses “coherence” and “water” as masks. The choice of min_cluster_pixels is related to coherence bias: if high bias is expected, this value should be increased. Since bias is often stronger in low-coherence areas, the parameter is closely tied to coherence level, though accounting for all factors can be tricky. For now, I'm setting this value based on my experience, but may need to change after looking actual interferograms...

oberonia78:
@vbrancat Would you be able to review one more time? This is a simple PR and I corrected as you request and answer your question. Please let me know if you don't agree with my answer. Thanks @vbrancat

vbrancat (approved):
LGTM. Thanks

PR #85, opened 2025-08-10
Crop Coverage in Pointing DNC Workflows

rad-eng-59:
This PR simply adds a capability to process a smaller portion of a large L0B product by cropping in along track for all three pointing DNC workflows, that is EL Null (DM2), EL rising edge (Science) and Doppler centroid estimation (DM2+Science).
This update leads to adding two new optional argument `time-start` and `time-dur-max`, both in seconds wrt first range line of the L0B. 
The updates help select the decent chunks of a long DM2/science product best suited for correct pointing estimation from viewpoint of both scene backscatter and topography variation.
This also helps with total runtime by avoiding processing of poor and unfit data that can lead to biased pointing estimation.
**Note that a bunch of REE simulated sample products is temporarily added back to the repo  to allow full testing of these new additions. They will be removed prior to merging!**


python/packages/nisar/pointing/doppler_lut_from_raw.py
```diff
@@ -417,9 +422,26 @@ def time_dop_est(echo, prf, lag=1, axis=None):
...
+                tot_pulses - 1)
+        if rangeline_slice.stop is not None:
+            rgl_start_stop[1] = max(
+                min(rangeline_slice.stop, tot_pulses),
+                rgl_start_stop[0] + 1)
+    logger.info('[start, stop] range line index to be processed -> '
+                f'{rgl_start_stop}')
+    # update number of range lines to be processed
+    num_rgls = rgl_start_stop[1] - rgl_start_stop[0]
+    logger.info(f'Number of range lines to be processed -> {num_rgls}')
```
bhawkins, line 440: Need to throw an exception if `rangeline_slice.step not in (None, 1)`.

python/packages/nisar/pointing/el_rising_edge_from_raw_ant.py
```diff
@@ -223,9 +227,25 @@ def el_rising_edge_from_raw_ant(raw, ant, *, dem_interp=None,
...
+    rgl_start_stop = [0, num_rgls_tot]
+    if rangeline_slice is not None:
+        if rangeline_slice.start is not None:
+            rgl_start_stop[0] = min(
+                max(rangeline_slice.start, 0),
+                num_rgls_tot - 1)
+        if rangeline_slice.stop is not None:
+            rgl_start_stop[1] = max(
+                min(rangeline_slice.stop, num_rgls_tot),
+                rgl_start_stop[0] + 1)
```
bhawkins: This logic is unintuitive to me. Python slicing is pretty flexible, and if you have something like `slice(1, 0)` it'll just resolve to an empty selection. However, this logic means `slice(1, 0)` becomes equivalent to `slice(1, 2)` which is pretty different behavior. I guess that's okay if documented.

bhawkins (commented):
Thanks for this handy feature. The use of a Python `slice` object with different slicing semantics causes me a bit of heartburn. Ideally stuff `slice(1, 0)` or `slice(0, 4, 5)` would either work as they do elsewhere or cause an exception for use cases we don't want these methods to support.

rad-eng-59:
> Thanks for this handy feature. The use of a Python `slice` object with different slicing semantics causes me a bit of heartburn. Ideally stuff `slice(1, 0)` or `slice(0, 4, 5)` would either work as they do elsewhere or cause an exception for use cases we don't want these methods to support.

I had a feeling using slice in place of tuple for limiting range lines would raise concern like this, FYI, I was supposed to use a tuple (start,stop) to limit range lines in all three workflows but got lazy and use a more compact but way more extensible version via slice object by ignoring its step arg hoping no once would complain. My bad. Lesson learned.
I'll change them all to a tuple to avoid all these headaches.  Sorry about that.

bhawkins (approved):
LGTM, just need to remove the big files.

python/packages/nisar/pointing/doppler_lut_from_raw.py
```diff
@@ -77,6 +78,10 @@ def doppler_lut_from_raw(raw, *, freq_band='A', txrx_pol=None,
     az_block_dur : float, default=4.0
         Azimuth block duration in seconds defining time-domain correlator
         length used in Doppler estimator.
+    rangeline_limit : tuple[int | None, int | None], optional
+        0-based range line [start, stop] indices to limit echo range lines to
+        be processed. Default is all range lines. The start/stop will be
+        limited to within [0, total rangelines]!
```
gmgunter: Let's update this description to make clear that the interval is not inclusive of the right endpoint.

```suggestion
        0-based range line [start, stop) indices to limit echo range lines to
        be processed. Default is all range lines. The start/stop will be
        limited to within [0, total rangelines)!
```

python/packages/nisar/pointing/doppler_lut_from_raw.py
```diff
@@ -417,9 +422,26 @@ def time_dop_est(echo, prf, lag=1, axis=None):
...
+    if rangeline_limit is not None:
+        if rangeline_limit[0] is not None:
+            rgl_start_stop[0] = min(
+                max(rangeline_limit[0], 0),
+                tot_pulses - 1)
+        if rangeline_limit[1] is not None:
+            rgl_start_stop[1] = max(
+                min(rangeline_limit[1], tot_pulses),
+                rgl_start_stop[0] + 1)
+    logger.info('[start, stop] range line index to be processed -> '
```
gmgunter:
```suggestion
    logger.info('[start, stop) range line index to be processed -> '
```

python/packages/nisar/pointing/doppler_lut_from_raw.py
gmgunter: (Please make this change in the other scripts as well, thanks!)

python/packages/nisar/workflows/gen_el_null_range_product.py
```diff
@@ -209,6 +218,29 @@ def gen_el_null_range_product(args):
         raw_obj, args.freq_band, args.txrx_pol)
     logger.info(f'List of selected frequency bands and TxRx Pols -> {frq_pol}')
 
+    # determine [start, stop] range lines of echo to be processed.
+    freq_band = list(frq_pol.keys())[0]
+    txrx_pol = frq_pol[freq_band][0]
+    epoch, tm = raw_obj.getPulseTimes(freq_band, txrx_pol[0])
```
gmgunter: `epoch` seems to be unused

```suggestion
    _, tm = raw_obj.getPulseTimes(freq_band, txrx_pol[0])
```

gmgunter (approved):
LGTM, thanks!

We reviewed this in our PR Fridays meeting 2025-08-29

gmgunter (approved):
Thanks, @rad-eng-59!

PR #44, opened 2025-06-30
Save timing correction LUTs into GCOV products

gshiroma:
This PR updates the GCOV writer to save the azimuth and range timing correction LUTs into GCOV products. These LUTs are geocoded using the same geogrid utilized to save other `processingInformation` LUTs and are stored within the H5 Group `science/LSAR/GCOV/metadata/processingInformation/timingCorrections` as the H5 datasets `azimuthIonosphere` and `slantRangeIonosphere` describing the azimuth and slant-range ionospheric shifts respectively.

PR #86, opened 2025-08-11
Subdivide GCOV timing correction LUTs into frequencies A and B

gshiroma:
This PR follows #44 and updates the GCOV writer to save the GCOV timing correction LUTs into subgroups associated with frequencies A and B. So, instead of saving these LUTs into:

`/science/LSAR/GCOV/metadata/processingInformation/timingCorrections/`

They'll be saved into:

`/science/LSAR/GCOV/metadata/processingInformation/timingCorrections/frequency{A|B}` .


gmgunter:
We still expect the LUT in frequency A and B to be the same, though, right?

The frequency sub-band just determines the wavelength, look direction, and reference epoch of the LUT's radar grid. The grids should be zero-Doppler so the wavelength presumably doesn't affect geocoding, and the other two parameters should be the same in both sub-bands.

In that case, would it make more sense to just have a single LUT for both frequency sub-bands, instead of splitting it up into `frequencyA` and `frequencyB` groups?

gshiroma:
> We still expect the LUT in frequency A and B to be the same, though, right?
> 
> The frequency sub-band just determines the wavelength, look direction, and reference epoch of the LUT's radar grid. The grids should be zero-Doppler so the wavelength presumably doesn't affect geocoding, and the other two parameters should be the same in both sub-bands.
> 
> In that case, would it make more sense to just have a single LUT for both frequency sub-bands, instead of splitting it up into `frequencyA` and `frequencyB` groups?

Hi @gmgunter , yes, the center frequency (which depends on the frequency band) is used to compute the timing correction LUTs due to ionosphere in both the [range direction](https://github.com/isce-framework/isce3/blob/7d15ed2f8dc52210bf0e7d1c274205cf3f9e37e4/python/packages/isce3/atmosphere/tec_product.py#L105) and [azimuth direction](https://github.com/isce-framework/isce3/blob/7d15ed2f8dc52210bf0e7d1c274205cf3f9e37e4/python/packages/isce3/atmosphere/tec_product.py#L181). Thanks for checking!

gmgunter:
Ah, I see, thanks. Good catch :+1:

gmgunter (approved):
LGTM, thanks.

Note: this update is consistent with the proposed XML spec updates in PIX: https://github-fn.jpl.nasa.gov/NISAR-ADT/NISAR_PIX/pull/308

gshiroma:
> LGTM, thanks.
> 
> Note: this update is consistent with the proposed XML spec updates in PIX: https://github-fn.jpl.nasa.gov/NISAR-ADT/NISAR_PIX/pull/308

Thank you, @gmgunter !

hfattahi (approved):
LGTM!

gshiroma:
> LGTM!

Thank you, @hfattahi !

PR #87, opened 2025-08-11
Update the DEM Attributes for the InSAR L1 Products

xhuang-jpl:
This PR is to response to this [request](https://github-fn.jpl.nasa.gov/NISAR-ADT/NISAR_PIX/pull/328) from the PIX repo where the description of the DEM needs to be updated and a new attribute `disclaimer` needs to be added.


python/packages/nisar/products/insar/InSAR_L1_writer.py
```diff
@@ -446,13 +446,11 @@ def add_pixel_offsets_to_swaths_group(self):
...
-                                                 " Copyright DLR e.V. 2010-2014 and Copyright Airbus Defence and"
-                                                 " Space GmbH 2014-2018 provided under COPERNICUS by the European Union and ESA;"
-                                                 " all rights reserved. This dataset is generated by referencing the"
-                                                 " Copernicus DEM elevations to the WGS84 ellipsoid and"
-                                                 " projecting them onto a range/Doppler grid"),
+                                                 " This dataset is generated by referencing the input DEM elevations"
+                                                 " to the WGS84 ellipsoid and projecting them onto a range/Doppler grid"
+                                                 ),
                                     units=Units.meter)
+            offset_group['digitalElevationModel'].attrs['disclaimer'] = np.bytes_('Disclaimer for DEM')
```
gmgunter: Thanks, @xhuang-jpl!

Sorry for the vague instructions -- I think our intention was that this `disclaimer` attribute should have the same contents as the `demSource` dataset. That way, it'll be clear to users that the DEM layer contained within this product is subject to the same license terms as the input DEM raster.

https://github.com/isce-framework/isce3/blob/ab000f4c17d76af1746731955ba3234c5b49677e/python/packages/nisar/products/insar/InSAR_base_writer.py#L791-L795

gmgunter: Also, note that the DEM description + license string is expected to contain Unicode (non-ASCII) characters, so converting it to a bytestring using `np.bytes_()` won't work as expected.

```python
>>> import numpy as np
>>> s = "The Copernicus DEM 30m was produced using Copernicus WorldDEM-30 © DLR e.V. 2010-2014 and © Airbus Defence and Space GmbH 2014-2018 provided under COPERNICUS by the European Union and ESA; all rights reserved."
>>> np.bytes_(s)
  File "<stdin>", line 1, in <module>
UnicodeEncodeError: 'ascii' codec can't encode character '\xa9' in position 65: ordinal not in range(128)
```

Instead, we could use something like `np.char.encode(..., encoding="utf-8")`.

```python
>>> np.char.encode(s, encoding="utf-8")
array(b'The Copernicus DEM 30m was produced using Copernicus WorldDEM-30 \xc2\xa9 DLR e.V. 2010-2014 and \xc2\xa9 Airbus Defence and Space GmbH 2014-2018 provided under COPERNICUS by the European Union and ESA; all rights reserved.',
      dtype='|S211')
```

Perhaps we should add a function somewhere in the `isce3` or `nisar` packages to do this conversion so that other NISAR products can make use of it as well.

```python
import numpy as np
from numpy.typing import ArrayLike

def to_bytes(s: str | ArrayLike) -> np.ndarray:
    """
    Convert a Unicode string or array of strings into bytes.

    Parameters
    ----------
    s : str or array_like
        A Unicode string or array of Unicode strings.

    Returns
    -------
    numpy.ndarray
        The input string(s) converted to bytestrings with 'utf-8' encoding.
    """
    return np.char.encode(s, encoding="utf-8")
```

xhuang-jpl:
[like]  Huang, Xiaodong (US 334K) reacted to your message:
________________________________
From: Geoffrey Gunter ***@***.***>
Sent: Monday, August 11, 2025 8:22:04 PM
To: isce-framework/isce3 ***@***.***>
Cc: Huang, Xiaodong (US 334K) ***@***.***>; Mention ***@***.***>
Subject: [EXTERNAL] Re: [isce-framework/isce3] Update the DEM Attributes for the InSAR L1 Products (PR #87)


@gmgunter commented on this pull request.

________________________________

In python/packages/nisar/products/insar/InSAR_L1_writer.py<https://urldefense.us/v3/__https://github.com/isce-framework/isce3/pull/87*discussion_r2267905017__;Iw!!PvBDto6Hs4WbVuu7!LgKPoPGMvu6jVh6Klw2vshtLkHDtiFqGe4fGd52YU1p6080woybtoBIRuHsyxvDliYEja61GQ8JtJR0vXueP4l3J-v-2oSv5Ew$>:

>                                      units=Units.meter)
+            offset_group['digitalElevationModel'].attrs['disclaimer'] = np.bytes_('Disclaimer for DEM')


Thanks, @xhuang-jpl<https://urldefense.us/v3/__https://github.com/xhuang-jpl__;!!PvBDto6Hs4WbVuu7!LgKPoPGMvu6jVh6Klw2vshtLkHDtiFqGe4fGd52YU1p6080woybtoBIRuHsyxvDliYEja61GQ8JtJR0vXueP4l3J-v90yEmIuA$>!

Sorry for the vague instructions -- I think our intention was that this disclaimer attribute should have the same contents as the demSource dataset. That way, it'll be clear to users that the DEM layer contained within this product is subject to the same license terms as the input DEM raster.

https://github.com/isce-framework/isce3/blob/ab000f4c17d76af1746731955ba3234c5b49677e/python/packages/nisar/products/insar/InSAR_base_writer.py#L791-L795<https://urldefense.us/v3/__https://github.com/isce-framework/isce3/blob/ab000f4c17d76af1746731955ba3234c5b49677e/python/packages/nisar/products/insar/InSAR_base_writer.py*L791-L795__;Iw!!PvBDto6Hs4WbVuu7!LgKPoPGMvu6jVh6Klw2vshtLkHDtiFqGe4fGd52YU1p6080woybtoBIRuHsyxvDliYEja61GQ8JtJR0vXueP4l3J-v_bLVot9g$>

________________________________

In python/packages/nisar/products/insar/InSAR_L1_writer.py<https://urldefense.us/v3/__https://github.com/isce-framework/isce3/pull/87*discussion_r2267935624__;Iw!!PvBDto6Hs4WbVuu7!LgKPoPGMvu6jVh6Klw2vshtLkHDtiFqGe4fGd52YU1p6080woybtoBIRuHsyxvDliYEja61GQ8JtJR0vXueP4l3J-v_5lcrYYw$>:

>                                      units=Units.meter)
+            offset_group['digitalElevationModel'].attrs['disclaimer'] = np.bytes_('Disclaimer for DEM')


Also, note that the DEM description + license string is expected to contain Unicode (non-ASCII) characters, so converting it to a bytestring using np.bytes_() won't work as expected.

>>> import numpy as np
>>> s = "The Copernicus DEM 30m was produced using Copernicus WorldDEM-30 © DLR e.V. 2010-2014 and © Airbus Defence and Space GmbH 2014-2018 provided under COPERNICUS by the European Union and ESA; all rights reserved."
>>> np.bytes_(s)
  File "<stdin>", line 1, in <module>
UnicodeEncodeError: 'ascii' codec can't encode character '\xa9' in position 65: ordinal not in range(128)

Instead, we could use something like np.char.encode(..., encoding="utf-8").

>>> np.char.encode(s, encoding="utf-8")
array(b'The Copernicus DEM 30m was produced using Copernicus WorldDEM-30 \xc2\xa9 DLR e.V. 2010-2014 and \xc2\xa9 Airbus Defence and Space GmbH 2014-2018 provided under COPERNICUS by the European Union and ESA; all rights reserved.',
      dtype='|S211')

Perhaps we should add a function somewhere in the isce3 or nisar packages to do this conversion so that other NISAR products can make use of it as well.

import numpy as np
from numpy.typing import ArrayLike

def to_bytes(s: str | ArrayLike) -> np.ndarray:
    """
    Convert a Unicode string or array of strings into bytes.

    Parameters
    ----------
    s : str or array_like
        A Unicode string or array of Unicode strings.

    Returns
    -------
    numpy.ndarray
        The input string(s) converted to bytestrings with 'utf-8' encoding.
    """
    return np.char.encode(s, encoding="utf-8")

python/packages/nisar/products/insar/InSAR_L1_writer.py
nemo794: @gshiroma , have you already created a function like what @gmgunter requested above?

gshiroma: > @gshiroma , have you already created a function like what @gmgunter requested above?

I haven't, but I think it's a great idea. Good catch from @gmgunter in noticing that we wouldn't be able to use `np.bytes_()` directly for the DEM description + license.

gmgunter: I ended up adding the `to_bytes()` function in #88 in order to ensure that my changes to `stage_dem.py` there don't break any SAS workflows.

I'm happy to coordinate with @xhuang-jpl to avoid conflicts and work duplication, though. If it's convenient to merge this PR first and then rebase my PR (or vice versa), that's fine with me.

xhuang-jpl: Thanks a lot @gmgunter , I have made a significant change of the InSAR reader to replace the np.bytes_ with the to_bytes, and copy your unit test case too. If you have time, please take a look and review. Thanks again

gmgunter: Sounds good, @xhuang-jpl, thanks.

It looks like, as part of your recent changes, lots of variables that were previously stored as bytes are now stored as strings, and some variables that were previously stored as strings are now stored as bytes. I'm not very familiar with the InSAR writer code so it might take me a while to check whether the updated code is still consistent with the expected behavior, but I'll work on it when I have time.

xhuang-jpl: for the string, they are all bytes same with before, just replace the np.bytes_ with the to_bytes function.

gmgunter: Here's an instance where a variable that was bytes is now a string:

image

More instances here:

image

Presumably, these strings are converted to bytes somewhere later on before writing them to the HDF5 product. But, since I'm not familiar with the code, it'll just take me a little while to verify that.

xhuang-jpl: you are right, they are all converted to the bytes in the writer.

python/packages/nisar/products/insar/InSAR_base_writer.py
```diff
@@ -790,22 +788,22 @@ def add_inputs_to_procinfo_group(self):
...
-                np.bytes_([os.path.basename(self.sec_h5_slc_file)]),
+                to_bytes([os.path.basename(self.sec_h5_slc_file)]),
                 "List of input secondary L1 RSLC products used",
             ),
             DatasetParams(
                 "orbitFiles",
-                np.bytes_(orbit_file),
+                to_bytes(orbit_file),
                 "List of input orbit files used",
             ),
```
gmgunter, line 808: Elsewhere in the codebase, when we create `DatasetParams` objects, we don't do this `to_bytes()` conversion. We just pass in the string without converting it to bytes, e.g.

https://github.com/isce-framework/isce3/blob/2435b7fb99dc12830f1d5e482e580d3ed8ba38de/python/packages/nisar/products/insar/InSAR_base_writer.py#L579-L583

It looks like the conversion to bytes happens later on inside the `add_dataset_and_attrs()` function.

https://github.com/isce-framework/isce3/blob/2435b7fb99dc12830f1d5e482e580d3ed8ba38de/python/packages/nisar/products/insar/dataset_params.py#L50

But here, when constructing these 3 `DatasetParams` objects, we're performing the `to_bytes()` conversion before passing them to `add_dataset_and_attrs()`. I think the reason you're doing that is that these objects are *lists of strings* and `add_dataset_to_attrs()` is only smart enough to convert to bytes if the input was a *single string*.

Should we add a comment here explaining what's going on? Or else maybe make `add_dataset_and_attrs()` smarter, so that it can convert lists of strings to bytes? It seems to me that this has the potential to cause confusion later on if we don't explain what's going on.

gmgunter (approved):
One minor suggestion. Otherwise LGTM, thanks!

python/packages/nisar/products/insar/InSAR_base_writer.py
xhuang-jpl: Thank you, @gmgunter! I’ve updated the docstring to make it clearer and prevent any confusion.

xhuang-jpl:
@gshiroma @vbrancat @nemo794 could you review this PR? @gmgunter  has approved it. Thank you

vbrancat (approved):
LGTM, thanks!

Issue #80, opened 2025-08-06
How to get DEM description + license for Static Layers?


gmgunter:
Unlike other NISAR products, the Static Layers product includes a resampled DEM layer, so it must include license information in the product in accordance with the terms of the input DEM license.

It's undesirable to hard-code the license info in the SAS because users may want to run the Static Layers workflow with a custom DEM that's subject to different license terms.

Currently, the runconfigs for all NISAR workflows include a `runconfig.groups.dynamic_ancillary_file_group.dem_file_description` field that could be used to supply this info. In the SDS, this field will be populated with a string extracted from the VRT file produced by `stage_dem.py`. This string contains a short description that does not include license info[^1].

We discussed several options for getting the DEM description + license in the Static Layers workflow offline this morning:

1. We could instruct the On-Demand team to populate this field in the runconfig with a pre-specified string that includes the license info, rather than extracting the string from the DEM VRT file.
2. We could add logic in the SAS to check whether the input DEM is a Copernicus DEM and, if so, use a hard-coded string with the license info.
3. We could update `stage_dem.py` to add an additional metadata attribute to the VRT file which includes the full description + license info. Then the SAS could extract this metadata from the VRT file.

I have a strong preference for option 3--

Option 1 requires coordinating with the On-Demand team. Option 2 requires implementing fuzzy logic to try to infer the source of the input DEM. Options 1 and 2 require hard-coding the license info in either the runconfig or the ISCE3 source code, which is less robust to potential future changes in the DEM license.

Option 3 is easy to implement and is most consistent with how we plan to handle passing the water mask license (see https://github.com/isce-framework/isce3/pull/77).

Let me know your thoughts, @gshiroma. I'm happy to implement the necessary changes to `stage_dem.py`.

cc @nemo794 @hfattahi @oberonia78 

[^1]: https://github.com/isce-framework/isce3/pull/77#issuecomment-3156232637

gshiroma:
@gmgunter , thank you for creating this issue and summarizing the problem. I was thinking about adding different statements for NISAR standard products and STATIC products. Something like:

NISAR standard products:
```
'This product has been generated using modified Copernicus DEM 30m data, provided by the European Space Agency. The Copernicus DEM 30m was produced using Copernicus WorldDEM-30 © DLR e.V. 2010-2014 and © Airbus Defence and Space GmbH 2014-2018 provided under COPERNICUS by the European Union and ESA; all rights reserved.
```

NISAR STATIC products:

```
[...]

PR #88, opened 2025-08-12
Update `stage_dem` to insert longer description (with license info) in output VRT

gmgunter:
Update `stage_dem.py` to extract both the "Short description" and "Notes" fields from the nisar-dem README file, concatenate them together, and insert them in the output VRT file as a metadata attribute.

Previously, only the "Short description" was inserted into the VRT. This string does not include license information, so we decided to also include the contents of the "Notes" field. (The "Notes" field doesn't contain the version number, so we want to include both.)

PCM will extract this info from the VRT file created with `stage_dem.py` and insert it into the SAS runconfig files. The SAS workflows then subsequently insert it in output NISAR HDF5 products in the `demSource` dataset.

See https://github.com/isce-framework/isce3/issues/80#issuecomment-3166138150 for additional context.

---

Following this change to `stage_dem.py`, the DEM description is now expected to contain non-ASCII characters. Converting such a string to bytes using `np.bytes_()` causes an exception to be raised.

```python
>>> import numpy as np
>>> s = "The Copernicus DEM 30m was produced using Copernicus WorldDEM-30 © DLR e.V. 2010-2014 and © Airbus Defence and Space GmbH 2014-2018 provided under COPERNICUS by the European Union and ESA; all rights reserved."
>>> np.bytes_(s)
  File "<stdin>", line 1, in <module>
UnicodeEncodeError: 'ascii' codec can't encode character '\xa9' in position 65: ordinal not in range(128)
```

This PR adds a helper function `nisar.products.utils.to_bytes()` that can be used instead to correctly encode any UTF-8 string (or array of UTF-8 strings) to bytes.

```python
>>> from nisar.products.utils import to_bytes
>>> to_bytes(s)
array(b'The Copernicus DEM 30m was produced using Copernicus WorldDEM-30 \xc2\xa9 DLR e.V. 2010-2014 and \xc2\xa9 Airbus Defence and Space GmbH 2014-2018 provided under COPERNICUS by the European Union and ESA; all rights reserved.',
      dtype='|S211')
```

Several instances of `np.bytes_(...)` were replaced with `nisar.products.utils.to_bytes(...)` in order to ensure that these updates to `stage_dem.py` don't cause NISAR workflows to break.


python/packages/nisar/workflows/stage_dem.py
```diff
@@ -378,77 +378,107 @@ def download_dem(polys, epsgs, outfile, version):
...
     # Build vrt with downloaded DEMs and add dem_descr in metadata
     vrt_dataset = gdal.BuildVRT(outfile, dem_list)
-    vrt_dataset.SetMetadataItem("dem_description", f'{dem_descr}')
+    vrt_dataset.SetMetadataItem("dem_description", full_descr)
 
     # Add license text to GeoTiff files
     for dem_file in dem_list:
-        add_dem_license_to_tiff(dem_file)
+        ds = gdal.Open(dem_file, gdal.GA_Update)
+        ds.SetMetadataItem("LICENSE", full_descr)
```
gshiroma: Just wondering why you are using the metadata field `"LICENSE" for the GeoTTIFs and "dem_description" for the VRT file? Maybe we can keep them consistent and use  "dem_description" for both?

gshiroma (approved):
LGTM! Thank you, @gmgunter , for updating the metadata the stage DEM with this new "dem_description" and the GCOV/GSLC/SLC product writers with this new `to_bytes()` function! I have a minor comment below, which is more of a question.

python/packages/nisar/workflows/stage_dem.py
gmgunter: I like the idea of using "dem_description" for both -- I just didn't want to break anything by renaming this metadata attribute from "LICENSE" to "dem_description".

Do you know if PGE/PCM is using this "LICENSE" attribute at all? Is it safe to rename it to "dem_description"?

gshiroma: Oh, that's true. This was implemented in the `add_dem_license_to_tiff()` function. Maybe @vbrancat knows if they use this field "LICENSE" in the GeoTIFFs?

xhuang-jpl:
Hi @gmgunter  JUST a reminder that this PR https://github.com/isce-framework/isce3/pull/87 was merged. Thanks again for the review.

gmgunter:
Thanks, @xhuang-jpl! I just rebased to incorporate your changes and addressed the merge conflicts.

rad-eng-59:
Since you are making changes to the `stage_dem` workflow, you may want to fix the type for margin from `int` to `float` in the [following line](https://github.com/gmgunter/isce3/blob/2ad875c0f26e111b50c0f5c63b1ebbc8cd716ab4/python/packages/nisar/workflows/stage_dem.py#L42) to avoid issue with a non-integer value provided as  margin in `km`. Thank you

python/packages/nisar/workflows/stage_dem.py
vbrancat: Yes, the license text was also added to the individual GeoTiffs downloaded by `stage_dem.py` as per the request of @gshiroma. I think the rationale was that the downloaded individual GeoTiffs may be redistributed.

gmgunter: Thanks @vbrancat. It sounds like the ground data system doesn't depend on this "LICENSE" attribute at all so I will go ahead and rename it to "dem_description" for consistency with the VRT file.

gmgunter:
@rad-eng-59 good suggestion. I'll make that change in a separate PR momentarily.

gmgunter:
FYI I just rebased to address a merge conflict, made a similar change to the one requested [here](https://github.com/isce-framework/isce3/pull/77#discussion_r2310779220) for `stage_watermask.py`, and pushed a fix for an issue with the unit test that I introduced in  f2dfeb5.

oberonia78 (approved):
LGTM. Tested locally. works as expected.

oberonia78:
Nit: When using v1.1 or v1.0, the first sentence of dem_description is missing a period, causing a run-on. The dem_description v1.2 looks fine.

gmgunter:
>Nit: When using v1.1 or v1.0, the first sentence of dem_description is missing a period, causing a run-on.

Nice catch! I pushed a small fix in d13eb85. Here's the "dem_description" for v1.1 after the fix:

>Digital Elevation Model (DEM) for the NASA-ISRO NISAR mission version 1.1 (v1.1) based on the Copernicus DEM 30-m and Copernicus 90-m referenced to the WGS84 ellipsoid. This digital elevation model (DEM) was processed at the Jet Propulsion Laboratory, California Institute of Technology, under contract with the National Aeronautics and Space Administration, using the Copernicus DEM 30-m and Copernicus DEM 90-m models provided by the European Space Agency. The Copernicus DEM 30-m and Copernicus DEM 90-m were produced using Copernicus WorldDEM-30 © DLR e.V. 2010-2014 and © Airbus Defence and Space GmbH 2014-2018 provided under COPERNICUS by the European Union and ESA; all rights reserved. The organizations in charge of the NISAR mission by law or by delegation do not incur any liability for any use of this DEM. The organisations in charge of the Copernicus programme by law or by delegation do not incur any liability for any use of the Copernicus WorldDEM-30.

PR #89, opened 2025-08-12
GCOV runconfig flag `fullcovariance` defaults to `True` if data is full-pol and `False` otherwise (MCR-93801)

gshiroma:
This PR sets the default value of the GCOV runconfig flag `fullcovariance` to `True` when the data to process is full-pol, and to `False` otherwise. 

To be full-pol, the dataset needs to have 3 or 4 polarizations. This is tested by verifying if the data to process includes the subset of polarizations [HH, VV, HV] or [HH, VV, VH], in frequency A or B. 

Notice that if the input RSLC if full-pol, but the user requests to process a dual-pol dataset, the flag `fullcovariance` should be set to `False`.


share/nisar/defaults/gcov.yaml
```diff
@@ -191,8 +191,10 @@ runconfig:
 
                 # Flag indicating whether the output GCOV product should include
                 # off-diagonal covariance elements. If False, only the diagonal
-                # elements are included.
-                fullcovariance:   False
+                # elements are included. If empty, `fullcovariance` is set to
+                # `True` any frequency to be processed includes a full-pol dataset
+                # (3 or 4 polarizations), and `False` otherwise
```
gmgunter: Suggest copying some of the description from the comments added in `gcov_runconfig.py` here to help explain the behavior to users.

One minor addition would be to try to include some of the verbiage from the XML file to better explain the receive-only case to users who might not be familiar with that mode.

https://github-fn.jpl.nasa.gov/NISAR-ADT/NISAR_PIX/blob/9f8c22fd4b18016b12d15c52ecf3956f5fc649ae/XML/L1/nisar_L1_RSLC.xml#L333

```suggestion
                # elements are included. If empty, `fullcovariance` is set to
                # `True` any frequency to be processed includes a full-pol dataset
                # (i.e. includes both co-pols and at least one cross-pol), and
                # `False` otherwise. 
```

share/nisar/defaults/gcov.yaml
```diff
@@ -191,8 +191,10 @@ runconfig:
...
+                fullcovariance: None
```
gmgunter: Let's just leave this empty instead of setting the default to the literal `"None"` so that the yaml parser treats it as the Python object `None` instead of the string `"None"`.

This is also consistent with how we treat other similar parameters in the runconfig that default to None.

```suggestion
                fullcovariance:
```

python/packages/nisar/workflows/gcov_runconfig.py
```diff
@@ -16,6 +18,94 @@ def load(self):
...
+        warning_channel = journal.warning('GCOVRunConfig.load()')
+
+        flag_fullcovariance = self.cfg['processing']['input_subset'][
+            'fullcovariance']
+
+        # Handle the case in which the `fullcovariance` flag field is empty.
+        # If so, the YAML parser assigns it the string value `"None"`.
+        if (flag_fullcovariance is None or
+                (isinstance(flag_fullcovariance, str) and
+                 flag_fullcovariance == 'None')):
```
gmgunter: I think we shouldn't need this if you adopt the previous comment

```suggestion
        if flag_fullcovariance is None:
```

python/packages/nisar/workflows/gcov_runconfig.py
```diff
@@ -16,6 +18,94 @@ def load(self):
...
+                        slc_pol_dataset = h5_obj[slc_pol_path]
+                        is_receive_only = False
+                        if 'isReceiveOnly' in slc_pol_dataset.attrs.keys():
+                            receive_only_attr = slc_pol_dataset.attrs[
+                                'isReceiveOnly']
+                            if not isinstance(receive_only_attr, str):
+                                receive_only_attr = \
+                                    receive_only_attr.tobytes().decode()
+                            is_receive_only = \
+                                receive_only_attr.title() == 'True'
```
gmgunter: Instead of opening the RSLC product as an `h5py.File`, let's update the `RSLC` product reader to add a method that checks whether a particular dataset in the product is receive-only.

```python
def is_receive_only(self, freq, pol) -> bool:
    '''
    '''
    dataset = self.getSlcDataset(freq, pol)
    if 'isReceiveOnly' in dataset.attrs.keys():
        receive_only_attr = slc_pol_dataset.attrs['isReceiveOnly'].decode()
        return receive_only_attr.title() == 'True'
    return False
```

gmgunter: Since many legacy NISAR products contain string-valued datasets (which should be populated as bytes), in a separate PR, we could add a helper function that checks whether the dataset is string-valued or bytes-valued and handles both appropriately. This could be used consistently throughout the codebase whenever we need to read a string/bytes-valued dataset.

python/packages/nisar/workflows/gcov_runconfig.py
```diff
@@ -16,6 +18,94 @@ def load(self):
...
+
+            # Finally, `fullcovariance` is set to `True` if any frequency
+            # to be processed includes a full-pol dataset
+            # (3 or 4 polarizations) and it does not contain a
+            # mix of "receive-only" and nominal channels; otherwise,
+            # `fullcovariance` is set to `False`.
+            flag_fullcovariance = False
+            for freq, pol_list in freq_pols_dict.items():
+                if has_mixed_nominal_and_receive_only[freq]:
+                    continue
```
gmgunter: Consider the case where frequency A is full-pol but frequency B contains a mix of nominal and receive-only channels. What do we want to do in that case? Should we populate the full covariance matrix, even though it will contain mixed nominal/receive-only terms for frequency B?

@hfattahi @bhawkins 

@gshiroma and I tentatively agree that the current behavior makes sense -- we should populate the full covariance matrix in this case, but want your opinion.

python/packages/nisar/workflows/gcov_runconfig.py
```diff
@@ -16,6 +18,94 @@ def load(self):
...
+
+                # Verify if frequency to process is full-pol
```
gmgunter:
```suggestion
                # Verify if frequency to process is full-pol. It's considered
                # full-pol if it contains both co-pols and at least one cross-pol.
```

python/packages/nisar/workflows/gcov_runconfig.py
```diff
@@ -16,6 +18,94 @@ def load(self):
...
+                if (all([pol in pol_list
+                        for pol in ['HH', 'VV', 'HV']]) or
+                    all([pol in pol_list
+                        for pol in ['HH', 'VV', 'VH']])):
```
gmgunter:
```suggestion
                if (('HH' in pol_list) and ('VV' in pol_list) and 
                   ('HV' in pol_list or 'VH' in pol_list)):
```

gmgunter (commented):
Reviewed this in the ISCE3 meeting 2025-08-21

python/packages/nisar/workflows/gcov_runconfig.py
gshiroma: Yes, just to add to @gmgunter 's comment, we don't want to lose "valid data" in one frequency (freq. A in @gmgunter 's example) because the other frequency (e.g., freq. B) has this mix of nominal/receive-only channels.

python/packages/nisar/workflows/gcov_runconfig.py
gshiroma: Hi @gmgunter , what do you think of the implementation I just pushed in the commit https://github.com/isce-framework/isce3/pull/89/commits/68adf7ff3bc2c899a092ea13d4126d36ee63646d.

gshiroma:
Hi @gmgunter , I believe I addressed all your comments. If you have a chance, would you mind taking another look at this PR, please? Thank you.

hfattahi:
Let's simplify the logic to create fullcov regardless of noise only or not in frequency B. And that is consistent with the existing MCR.

gshiroma:
> Let's simplify the logic to create fullcov regardless of noise only or not in frequency B. And that is consistent with the existing MCR.

Will do. Thank you, @hfattahi !

gshiroma:
> Let's simplify the logic to create fullcov regardless of noise only or not in frequency B. And that is consistent with the existing MCR.

Thank you again, @hfattahi , I simplified the code and remove the conditions that depended on the SLCs' attribute `isReceiveOnly` in commit [504b515](https://github.com/isce-framework/isce3/pull/89/commits/504b515a72a456bda49d6b8d5e992836462a0e95). 

cc.: @gmgunter @bhawkins

python/packages/nisar/products/readers/SLC/RSLC.py
```diff
@@ -84,6 +85,39 @@ def getSlcDatasetAsNativeComplex(
...
+        pol: str
+            The polarization term associated with the SLC dataset
+
+        Returns
+        -------
+        is_receive_only: bool or None
+            Flag indicating whether the SLC has been acquired in
+            receive-only mode, or None if undetermined.
+        '''
+        slc_pol_dataset = self.getSlcDatasetAsNativeComplex(
```
gmgunter: Here we should use the [`getSlcDataset()`](https://github.com/isce-framework/isce3/blob/675bde1e32611f856606d11f95d8654b70fe1ac1/python/packages/nisar/products/readers/SLC/SLCBase.py#L9) method that always returns an `h5py.Dataset` object.

The [`getSlcDatasetAsNativeComplex()`](https://github.com/isce-framework/isce3/blob/675bde1e32611f856606d11f95d8654b70fe1ac1/python/packages/nisar/products/readers/SLC/RSLC.py#L52-L56) method might return a [`ComplexFloat16Decoder`](https://github.com/isce-framework/isce3/blob/675bde1e32611f856606d11f95d8654b70fe1ac1/python/packages/isce3/core/types.py#L99) object instead if the RSLC product used half-precision. `ComplexFloat16Decoder` objects don't have `attrs` so we can't use them here.

```suggestion
        slc_pol_dataset = self.getSlcDataset(
```

python/packages/nisar/products/readers/SLC/RSLC.py
```diff
@@ -84,6 +85,39 @@ def getSlcDatasetAsNativeComplex(
...
+            frequency=frequency, polarization=polarization)
+
+        receive_only_attr = \
+            get_attribute_from_h5_dataset(slc_pol_dataset,
+                                          'isReceiveOnly',
+                                          format_function=str)
+        if receive_only_attr is None:
+            return
```
gmgunter: NBD, but this might be a good place to emit a warning saying that we were unable to determine whether the RSLC dataset was receive-only or not

python/packages/nisar/products/readers/Base/Base.py
```diff
@@ -6,6 +6,44 @@
...
+def get_attribute_from_h5_dataset(h5_dataset, attribute_name,
+                                  default=None,
+                                  format_function=None):
+    '''
+    Return attribute from an H5 dataset.
+
+    Parameters
+    ----------
+    h5_dataset: h5py.Dataset
+        H5 dataset containing the returning attribute
```
gmgunter: HDF5 groups can also have attributes. In theory, this function could be used to get the global attributes from the root group of NISAR products as well, such as the `"Conventions"` attribute.

gmgunter:
```suggestion
    Return attribute from an HDF5 dataset.

    Parameters
    ----------
    h5_dataset: h5py.Dataset
        HDF5 dataset containing the returning attribute
```

python/packages/nisar/products/readers/Base/Base.py
```diff
@@ -6,6 +6,44 @@
...
+    format_function: function, optional
+        Function to format attribute values
+
+    Returns
+    -------
+    h5_attribute: scalar
+        H5 attribute
+    '''
+    if attribute_name not in h5_dataset.attrs.keys():
+        return default
```
gmgunter: In my mind, this function is doing two things that I would rather separate into two separate functions:

1. It's checking if the attribute exists and falling back to a default value otherwise
2. It's applying `format_function` to the attribute

Coupling these two behaviors together in a single function might not be desirable in cases where we just want to do one of them but not the other.

In our group review last week, we discussed that this helper function could be used consistently throughout the codebase whenever we need to read a string/bytes-valued dataset/attribute, in order to ensure that the behavior is always the same when we parse this kind of metadata from NISAR products.

There are probably many datasets/attributes that we would want to access in NISAR products without providing a default value -- that is, we want the operation to fail if those datasets/attributes don't exist.

So I'd suggest separating this into two separate functions so we can reuse the functionality to convert bytes to str in cases where we don't want to provide a default value.

python/packages/nisar/products/readers/Base/Base.py
```diff
@@ -6,6 +6,44 @@
...
+
+    h5_attribute = h5_dataset.attrs[attribute_name]
+    if format_function is None:
+        return h5_attribute
+
+    if (format_function == str and
+            isinstance(h5_attribute, np.bytes_)):
+        h5_attribute = h5_attribute.decode()
```
gmgunter: Making this functionality flexible using `format_function` is a clever approach. It's nice to be able to apply generic transformations to the attributes that we're reading, rather than just being able to convert bytes to str.

However, I'm not a fan of adding this special case to handle the bytes -> str conversion.

This isn't the behavior that I would expect after reading the docstring. Applying the `str()` function to a `np.bytes_` object doesn't decode the bytes to a Unicode string:

```python
>>> import numpy as np
>>> data = np.bytes_("Hello world!")
>>> str(data)
"b'Hello world!'"
```

So if I saw someone passing `format_function=str` to this function, I would be confused if I didn't know about this special behavior that's not documented in the docstring.

In order to follow the [principle of least surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment), I think it'd be clearer and simpler to just implement a helper function that handles the conversion from `bytes` -> `str` (with a fallback if the data is already `str`) as we discussed last week.

```python
def to_str(s: bytes | np.bytes_ | str) -> str:
    """
    Decode the input byte string to a Unicode string.

    The input is returned unmodified if it is already a string.

    Parameters
    ----------
    s : bytes or numpy.bytes_ or str
        The input byte string or Unicode string.

    Returns
    -------
    str
        The input decoded to a Unicode string.
    """
    if isinstance(s, str):
        return s
    return s.decode()
```

Then we can use this function throughout the codebase everywhere we need to convert NISAR datasets and attributes to strings.

gmgunter (commented):
Thanks for addressing my comments, @gshiroma.

The updates to the GCOV runconfig logic LGTM. My remaining suggestions all relate to the functionality that checks whether an input RSLC dataset is receive-only or not. Since you no longer need this functionality here, maybe it'd be easiest to move these changes to a separate PR. Otherwise, I'm happy to keep iterating here if you prefer.

gshiroma:
> Thanks for addressing my comments, @gshiroma.
> 
> The updates to the GCOV runconfig logic LGTM. My remaining suggestions all relate to the functionality that checks whether an input RSLC dataset is receive-only or not. Since you no longer need this functionality here, maybe it'd be easiest to move these changes to a separate PR. Otherwise, I'm happy to keep iterating here if you prefer.

Good point, @gmgunter , since we no longer need that functionality, it’s better to remove it from this PR. I could open another PR to add it later if necessary, but at this point it may not be worth it, since there’s no longer a clear need and it would require an MCR =/ 

The changes to the `Base` and `RSLC` reader were reverted in the commit [8f89cfd](https://github.com/isce-framework/isce3/pull/89/commits/8f89cfd4d035db871dd44890049c6d2f15928fbc).

If I issue another PR, I'll make sure to address your comments (and also remove the argument `format_function`). Thank you for your review!

gmgunter (approved):
Thank you! This looks great!

gshiroma:
> Thank you! This looks great!

That was fast! Thanks again, @gmgunter .

bhawkins (approved):
LGTM. It ended up nice and simple!

python/packages/nisar/workflows/gcov_runconfig.py
```diff
@@ -16,6 +17,55 @@ def load(self):
...
+                # cross-pol.
+                if (('HH' in pol_list) and ('VV' in pol_list) and
+                   ('HV' in pol_list or 'VH' in pol_list)):
+                    warning_channel.log(
+                        'The `fullcovariance` field is not specified in the'
+                        'runconfig. By default, it is set to `True` if any'
+                        'processed frequency includes a full-pol dataset,'
+                        ' and `False` otherwise.'
+                        f'Since frequency {freq} contains polarizations'
+                        f'{pol_list}, `fullcovariance` is set to `True`.')
```
gmgunter: Looks like there are some spaces missing between these lines

```suggestion
                        'The `fullcovariance` field is not specified in the'
                        ' runconfig. By default, it is set to `True` if any'
                        ' processed frequency includes a full-pol dataset,'
                        ' and `False` otherwise.'
                        f' Since frequency {freq} contains polarizations'
                        f' {pol_list}, `fullcovariance` is set to `True`.')
```

python/packages/nisar/workflows/gcov_runconfig.py
```diff
@@ -16,6 +17,55 @@ def load(self):
...
+                    break
+            else:
+                warning_channel.log(
+                    'The `fullcovariance` field is not specified in the'
+                    'runconfig. By default, it is set to `True` if any'
+                    'processed frequency includes a full-pol dataset,'
+                    ' and `False` otherwise.'
+                    'Since the output product will not contain a full-'
+                    'polarimetric dataset, `fullcovariance` is set to'
+                    ' `False`.')
```
gmgunter:
```suggestion
                    'The `fullcovariance` field is not specified in the'
                    ' runconfig. By default, it is set to `True` if any'
                    ' processed frequency includes a full-pol dataset,'
                    ' and `False` otherwise.'
                    ' Since the output product will not contain a full-'
                    ' polarimetric dataset, `fullcovariance` is set to'
                    ' `False`.')
```

python/packages/nisar/workflows/gcov_runconfig.py
gshiroma: Yes! Thank you, @gmgunter !

gshiroma:
> LGTM. It ended up nice and simple!

Thank you, @bhawkins !

PR #90, opened 2025-08-12
Spectra Analysis Tool for Noise Only Data

hb6688:
This tool reads an L0B file and does the following:

- Read azimuth chunks of data
- Compute averaged range power spectra density through scipy.welch function
- Generate a simple frequency domain RFI threshold in dB/Hz based on time domain noise power estimate using Eigenvalue Decomposition.
- Count the occurrences of RFI for each frequency bin
- Count the max RFI hit streak for each frequency bin.
- Write the followings to an output HDF5 file:
output_screenshot3

### DIFF
--- a/share/nisar/examples/noise_spectra_analysis.py
+++ b/share/nisar/examples/noise_spectra_analysis.py
@@ -202,7 +202,7 @@ def update_rfi_hit_stats(
     # Update total hit count
     rfi_freq_bin_hit_count += np.sum(rfi_mask_blk, axis=0)
 
-    # Vectorized max streak computation per frequency bin
+    # Zero-pad start and end of RFI Mask
     diff = np.diff(np.pad(rfi_mask_blk.astype(int), ((1, 1), (0, 0)), mode='constant'), axis=0)
 
     # starts and ends are tuples: starts[0]: pulse idx; starts[1]: freq bin idx
@@ -285,7 +285,7 @@ def process_l0b_data(
         f_out.attrs['num_pulses_block'] = num_pulses_blk
         f_out.attrs['num_fft'] = num_fft
         f_out.attrs['psd_x_axis'] = 'range (axis=1)'
-        f_out.attrs['psd_y_axis'] = 'power (axis=0, dB)'
+        f_out.attrs['psd_y_axis'] = 'power spectra density (axis=0, dB/Hz)'
 
         for freq, pol_list in raw.polarizations.items(): # A, B
             for pol in pol_list: # HH, HV, VV, VH
@@ -336,7 +336,7 @@ def process_l0b_data(
                 dset_fc.attrs["units"] = "Hz"
                 dset_fc.attrs["dtype"] = str(dset_fc.dtype)
 
-                # Center Frequency
+                # Sampling Frequency
                 if "freqSampling" in out_grp:
                     del out_grp["freqSamping"]
                 dset_fs = out_grp.create_dataset('freqSampling', data=fs)
--- a/share/nisar/examples/noise_spectra_analysis.py
+++ b/share/nisar/examples/noise_spectra_analysis.py
@@ -11,7 +11,6 @@ from scipy.signal import welch
 from numpy import linalg as la
 from nisar.products.readers.Raw import Raw
 from isce3.signal.compute_evd_cpi import slice_gen
-import warnings
 import os
 
 
@@ -59,7 +58,7 @@ def cmd_line_parse():
         dest="num_pulses_blk",
         type=int,
         required=False,
-        default="10000",
+        default="4000",
         help="Number of pulses per batch to be averaged in Azimuth-Time. Default: 5000",
     )
     parser.add_argument(
@@ -68,7 +67,7 @@ def cmd_line_parse():
         dest="thresh_margin",
         type=float,
         required=False,
-        default="6",
+        default="14",
         help="Additional gain to be added to noise power estimate for RFI thresholding." 
              "Default: 6"
     )
@@ -195,7 +194,7 @@ def copy_group_from_input_hdf5(
             # Copy the group
             f_in.copy(input_group_path, f_out[output_group_path])
         else:
-            warnings.warn(f"'{input_group_path}' does NOT exist in the input file.")
+            raise KeyError(f"'{input_group_path}' does NOT exist in the input file.")
 
 def process_l0b_data(
     raw: Raw,
@@ -279,22 +278,24 @@ def process_l0b_data(
                 num_freq_bins = len(freqs)
 
                 # Create output group for output HDF5
-                base_path = f"/science/LSAR/QA/data/freq{freq}/{pol}"
+                base_path = f"/science/LSAR/QA/data/frequency{freq}/{pol}"
                 out_grp = f_out.require_group(base_path)
 
                 # Center Frequency
                 if "centerFrequency" in out_grp:
                     del out_grp["centerFrequency"]
-                dset_fc = out_grp.create_dataset('centerFrequency', data=fc)
+                dset_fc = out_grp.require_dataset('centerFrequency', (), np.float64)
+                dset_fc[()] = fc
                 dset_fc.attrs["description"] = (
                     f"Raw data center frequency"
                 )
                 dset_fc.attrs["units"] = np.bytes_("Hz")
 
                 # Sampling Frequency
-                if "samplingFrequency" in out_grp:
-                    del out_grp["freqSamping"]
-                dset_fs = out_grp.create_dataset('samplingFrequency', data=fs)
+                if "sampleRate" in out_grp:
+                    del out_grp["sampleRate"]
+                dset_fs = out_grp.create_dataset('sampleRate', (), np.float64)
+                dset_fs[()] = fs
                 dset_fs.attrs["description"] = (
                     f"Raw data sampling frequency"
                 )
--- a/share/nisar/examples/noise_spectra_analysis.py
+++ b/share/nisar/examples/noise_spectra_analysis.py
@@ -170,9 +170,8 @@ def update_rfi_hit_stats(
         start_idxs = starts[0][starts[1] == b]
         end_idxs = ends[0][ends[1] == b]
 
-        if len(start_idxs) != len(end_idxs):
-            # If a streak starts but never ends (shouldn't happen with padding)
-             continue
+        # Expected an end to every beginning
+        len(start_idxs) == len(end_idxs)
 
         streaks = end_idxs - start_idxs
         if len(streaks) > 0:
--- a/share/nisar/examples/noise_spectra_analysis.py
+++ b/share/nisar/examples/noise_spectra_analysis.py
@@ -283,24 +283,22 @@ def process_l0b_data(
                 out_grp = f_out.require_group(base_path)
 
                 # Center Frequency
-                if "freqCenter" in out_grp:
-                    del out_grp["freqCenter"]
-                dset_fc = out_grp.create_dataset('freqCenter', data=fc)
+                if "centerFrequency" in out_grp:
+                    del out_grp["centerFrequency"]
+                dset_fc = out_grp.create_dataset('centerFrequency', data=fc)
                 dset_fc.attrs["description"] = (
                     f"Raw data center frequency"
                 )
-                dset_fc.attrs["units"] = "Hz"
-                dset_fc.attrs["dtype"] = str(dset_fc.dtype)
+                dset_fc.attrs["units"] = np.bytes_("Hz")
 
                 # Sampling Frequency
-                if "freqSampling" in out_grp:
+                if "samplingFrequency" in out_grp:
                     del out_grp["freqSamping"]
-                dset_fs = out_grp.create_dataset('freqSampling', data=fs)
+                dset_fs = out_grp.create_dataset('samplingFrequency', data=fs)
                 dset_fs.attrs["description"] = (
                     f"Raw data sampling frequency"
                 )
-                dset_fs.attrs["units"] = "Hz"
-                dset_fs.attrs["dtype"] = str(dset_fs.dtype)
+                dset_fs.attrs["units"] = np.bytes_("Hz")
 
                 # Power Spectra Density
                 if "rangePowerSpectralDensity" in out_grp:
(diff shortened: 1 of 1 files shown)


share/nisar/examples/noise_spectra_analysis.py
```diff
@@ -0,0 +1,501 @@
...
+        type=int,
+        default="2048",
+        help="Number of samples per segment for Scipy Welch function." 
+             "It is equal to number of FFT to be performed on each segment."
+             "Default: 2048"
+    )
+
+    parser.add_argument(
+        "-b",
+        "--blk_size",
```
bhawkins:
```suggestion
        "--block-size",
```
I prefer we avoid mixing dashes and underscores in CLI options. Also we may as well spell out "block" since lazy typists can just use `-b` instead.

share/nisar/examples/noise_spectra_analysis.py
```diff
@@ -0,0 +1,501 @@
...
+        Decomposition
+    """
+
+    num_pulses = raw_data.shape[0]
+
+    if cpi_len > num_pulses:
+        raise ValueError(f"Requested cpi_len={cpi_len} exceeds number of pulses {num_pulses}")
+    
+    # Equally spaced indices (without repetition)
+    indices = np.linspace(0, num_pulses - 1, cpi_len, dtype=int)
```
bhawkins: This strikes me as a little strange as it seems very unlikely that anything would remain coherent over the entire observation (except the caltone). So the noise estimate is basically just the quietest pulse of the six pulses we pluck out of the file. I guess that's okay.

share/nisar/examples/noise_spectra_analysis.py
```diff
@@ -0,0 +1,501 @@
...
+    ends = np.where(diff == -1)
+
+    # For each frequency bin, compute max streak
+    for b in range(rfi_mask_blk.shape[1]):
+        start_idxs = starts[0][starts[1] == b]
+        end_idxs = ends[0][ends[1] == b]
+
+        if len(start_idxs) != len(end_idxs):
+            # If a streak starts but never ends (shouldn't happen with padding)
+             continue
```
bhawkins:
```suggestion
        assert len(start_idxs) == len(end_idxs), "expected an end to every beginning"
```
If it really won't happen, then I think an assertion is a good way to document it.

BTW the vectorization in this function is pretty impressive.

share/nisar/examples/noise_spectra_analysis.py
```diff
@@ -0,0 +1,501 @@
...
+    with h5py.File(input_file, "r") as f_in, h5py.File(output_file, "a") as f_out:
+        if input_group_path in f_in:
+            # Check if the group already exists in output and delete it if necessary
+            if input_group_path in f_out:
+                del f_out[input_group_path]
+
+            # Copy the group
+            f_in.copy(input_group_path, f_out[output_group_path])
+        else:
+            warnings.warn(f"'{input_group_path}' does NOT exist in the input file.")
```
bhawkins: Do we need this clause? It potentially makes the output format unpredictable.

share/nisar/examples/noise_spectra_analysis.py
```diff
@@ -0,0 +1,501 @@
...
+                num_freq_bins = len(freqs)
+
+                # Create output group for output HDF5
+                base_path = f"/science/LSAR/QA/data/freq{freq}/{pol}"
+                out_grp = f_out.require_group(base_path)
+
+                # Center Frequency
+                if "freqCenter" in out_grp:
+                    del out_grp["freqCenter"]
+                dset_fc = out_grp.create_dataset('freqCenter', data=fc)
```
bhawkins:
```suggestion
                dset_fc = out_grp.require_dataset('centerFrequency', (), np.float64)
                dset_fc[()] = fc
```
Let's spell this the same way as in our other products ([example](https://github-fn.jpl.nasa.gov/NISAR-ADT/NISAR_PIX/blob/83d5e276f7d72093ce571977cec3cba8012508c4/XML/L0/nisar_L0B_RSD.xml#L264)).

share/nisar/examples/noise_spectra_analysis.py
```diff
@@ -0,0 +1,501 @@
...
+                dset_fc.attrs["description"] = (
+                    f"Raw data center frequency"
+                )
+                dset_fc.attrs["units"] = "Hz"
```
bhawkins:
```suggestion
                dset_fc.attrs["units"] = np.bytes_("Hz")
```
In all other products we're careful encode strings as fixed-length ASCII. The way all these are done here we end up with variable-length UTF-8 strings.

share/nisar/examples/noise_spectra_analysis.py
```diff
@@ -0,0 +1,501 @@
...
+                dset_fc.attrs["dtype"] = str(dset_fc.dtype)
+
+                # Center Frequency
+                if "freqSampling" in out_grp:
+                    del out_grp["freqSamping"]
+                dset_fs = out_grp.create_dataset('freqSampling', data=fs)
```
bhawkins:
```suggestion
                dset_fs = out_grp.require_dataset('sampleRate', (), np.float64)
                dset_fs[()] = fs
```
Use similar spelling to other products.

share/nisar/examples/noise_spectra_analysis.py
```diff
@@ -0,0 +1,501 @@
...
+
+                # Center Frequency
+                if "freqCenter" in out_grp:
+                    del out_grp["freqCenter"]
+                dset_fc = out_grp.create_dataset('freqCenter', data=fc)
+                dset_fc.attrs["description"] = (
+                    f"Raw data center frequency"
+                )
+                dset_fc.attrs["units"] = "Hz"
+                dset_fc.attrs["dtype"] = str(dset_fc.dtype)
```
bhawkins:
```suggestion
```
We shouldn't store dtype attributes. The data type is already encoded by the HDF5 API.

share/nisar/examples/noise_spectra_analysis.py
```diff
@@ -0,0 +1,501 @@
...
+
+        # Range Spectra Frequencies
+        if "rangeSpectraFrequencies" in top_grp:
+            del top_grp["rangeSpectraFrequencies"]
+        dset_freq = top_grp.create_dataset('rangeSpectraFrequencies', data=freqs)
+        dset_freq.attrs["description"] = (
+            f"Radio Frequency (RF) frequency coordinates for range power spectra"
+        )
+        dset_freq.attrs["units"] = "decibel re 1/hertz"
+        dset_freq.attrs["dtype"] = "dataset[list[list[float32]]]"
```
bhawkins: These look like they're at the wrong indentation level. These should be frequency-dependent.

bhawkins (commented):
Thanks for the quick turnaround on this tool. There's just some minor cleanup I'd like to see.

share/nisar/examples/noise_spectra_analysis.py
hb6688: @bhawkins , Thanks for pointing it out. I do not think coherent processing interval (CPI) is the correct term.  The purpose is to use a few pulses to estimate minimum eigenvalue of their sample covariance matrix which is an estimate of pulse power (regardless if data is noise-only or not).  The selected pulses are purposely spread across the entire data to avoid possible multiple RFI in consecutive pulses which could affect the minimum eigenvalue estimation. I don't have a good name for it.

How about:  '-p', '--proc-interval'?

share/nisar/examples/noise_spectra_analysis.py
hb6688: @bhawkins, I am not sure if I understood this particular comment from you?  I use this function to copy over identification group from the input file.  Are you referring to the following lines:

------
        if input_group_path in f_in:
            # Check if the group already exists in output and delete it if necessary
            if input_group_path in f_out:
                del f_out[input_group_path]
-------

hb6688:
This is the latest screenshot of output HDF5. If you have any questions about the script, please take a look at my result analysis notebook: /scratch/bohuang/alos/data/output/noise/spectra_result.ipynb

output_screenshot3

hb6688:
I believe this script meets our objectives.  Based on my result analyses, the outputs were correct.  Please let me know otherwise.

Thank you for your review.

rad-eng-59:
Hi @hb6688 
As discussed in the meeting, let's set the default values to the ones for your noise-only test case and make sure the threshold is high enough not to detect ant RFI hit. I think `14 dB` should work but please test it.
If possible and time allows It would be useful to log or print out some info for debugging purpose, for instance the input parameters being set as well as what band/pol/slice will be processed and elapsed time to show the progress. Currently, nothing is printed out or logged in `stdout` while testing your branch with instrument L0B w/ `RCID=1`. Btw, It took around `40 min` on `nisar-adt-dev-3`.
Please let me know when done with addressing comments and I'll retest your branch prior to approving it. 
Thanks.

hb6688:
@rad-eng-59 , Hi Hirad, does the spectrum frequency and its response of the dataset you processed look OK to you?  I think this is important.  The dataset I processed looked good to me (HDF5 output).  I changed the default margin to 14, and I am testing it.  I stored block size and num of FFT in the HDF5 attributes.  I do think some print statements are beneficial.  Currently I only display Frequency, polarization and output file path.  I will add a few print statements.  I already vectorized the RFI count operation so 40 minutes is probably as good as it gets.

rad-eng-59:
> @rad-eng-59 , Hi Hirad, does the spectrum frequency and its response of the dataset you processed look OK to you? I think this is important. The dataset I processed looked good to me (HDF5 output). I changed the default margin to 14, and I am testing it. I stored block size and num of FFT in the HDF5 attributes. I do think some print statements are beneficial. Currently I only display Frequency, polarization and output file path. I will add a few print statements. I already vectorized the RFI count operation so 40 minutes is probably as good as it gets.

I'm currently retesting your most recent updates. I've simply checked RFI counts not the spectrum. But per your spectrum plots from notebook, they look reasonable to me. The envelop is dominated by `80 MHz` BPF in qFSP w/ slight asymmetry due to RF anti-aliasing BPF.

share/nisar/examples/noise_spectra_analysis.py
```diff
@@ -0,0 +1,448 @@
...
+             "Default: 2048"
+    )
+    parser.add_argument(
+        "-b",
+        "--block-size",
+        dest="num_pulses_blk",
+        type=int,
+        required=False,
+        default="4000",
+        help="Number of pulses per batch to be averaged in Azimuth-Time. Default: 5000",
```
bhawkins:
```suggestion
        default="4096",
        help="Number of pulses per batch to be averaged in Azimuth-Time. Default: 4096",
```
Powers of two will be an integer multiple of the HDF5 block size and therefore will allow more efficient reading. Since this size is much larger than the block size, it doesn't matter much, though.

hb6688:
@rad-eng-59 , Hi Hirad, There is no more RFI hit when margin is raised to 14 dB.  I think it can be used as the default.

rfi_hit

share/nisar/examples/noise_spectra_analysis.py
```diff
@@ -0,0 +1,448 @@
...
+                out_grp = f_out.require_group(base_path)
+
+                # Center Frequency
+                if "centerFrequency" in out_grp:
+                    del out_grp["centerFrequency"]
+                dset_fc = out_grp.require_dataset('centerFrequency', (), np.float64)
+                dset_fc[()] = fc
+                dset_fc.attrs["description"] = (
+                    f"Raw data center frequency"
+                )
```
bhawkins:
```suggestion
                dset_fc.attrs["description"] = np.bytes_(
                    f"Raw data center frequency"
                )
```
It looks like you updated the units but not the description attributes. Not a huge deal.

bhawkins (approved):
Thanks for the quick updates!

share/nisar/examples/noise_spectra_analysis.py
hb6688: @bhawkins , I addressed your comments with regards to block size and description.  I tested it without any issues.

thanks,
Bo

rad-eng-59:
> @rad-eng-59 , Hi Hirad, There is no more RFI hit when margin is raised to 14 dB. I think it can be used as the default.
> rfi_hit

Did you check  both `HH` and `HV` to make sure there is no hit ?
Btw, I still see the unit in dataset name `noisePowerEstimateDB`. I thought you would remove it per @bhawkins comment?
Other than that, all seems good per my testing. 
Thank you.

share/nisar/examples/noise_spectra_analysis.py
```diff
@@ -0,0 +1,450 @@
...
+                del top_grp["thresholdMargin"]
+            dset_margin = top_grp.create_dataset('thresholdMargin', data=thresh_margin)
+            dset_margin.attrs["description"] =  np.bytes_(
+                f"RFI frequency domain detection margin"
+            )
+            dset_margin.attrs["units"] = np.bytes_("decibel")
+
+        # Copy Identification group from input L0B file
+        input_group_path = f'/science/LSAR/identification'
+        output_group_path = f'/science/LSAR'
```
rad-eng-59: As a minor suggestion, you could get these paths from `raw` object, via `raw._RootPath` and `raw._IdentificationPath` to avoid hardcoding and any unlikely but potential spec changes

share/nisar/examples/noise_spectra_analysis.py
```diff
@@ -0,0 +1,450 @@
...
+                    del out_grp["interferenceMaxStreak"]
+                dset_streak = out_grp.create_dataset('interferenceMaxStreak', data=rfi_bin_max_streak)
+                dset_streak.attrs["description"] =  np.bytes_(
+                    f"Maximum number of consecutive pulses where RFI detection threshold was exceeded "
+                    f"(in each freuqency bin)"
+                )
+                dset_streak.attrs["units"] = np.bytes_("1")
+
+                print(f'Processed: frequency{freq} {pol}')
+                print()
```
rad-eng-59, line 390: Just a minor suggestion:
```suggestion
                print(f'Processed: frequency{freq} {pol}\n')
```

share/nisar/examples/noise_spectra_analysis.py
```diff
@@ -0,0 +1,450 @@
...
+        Default: 6 
+
+    Returns
+    -------
+    noise_pwr_pulse_est_db: float
+        Estmated noise power per pulse using minimum Eigenvalue
+    """
+
+    data_estimate = select_pulses_for_estimation(raw_data, num_pulses_proc)  # [num_pulses_proc, num_range]
+    sample_cov = data_estimate @ data_estimate.conj().T / data_estimate.shape[1]  # [num_pulses_proc x num_pulses_proc]
```
rad-eng-59: Lines are too long.
Make sure to run `flake8` to check for PEP8 or simply `autopep8 -i <your script>`

rad-eng-59 (approved):
LGTM. I have provided some minor comments.
You may want to  check your script for PEP8 compatibility. I see some long lines or missing spaces and inline code comments.

hb6688:
@rad-eng-59 , I will clean it up and add runtime progress print statements tonight.  Thanks for your comments.

hb6688:
@rad-eng-59 @bhawkins @gmgunter, I added one final update by changing the completion progress to percentage.  Please take a look.

PR #92, opened 2025-08-13
Add minimum python version build to CI matrix

rtburns-jpl:


.github/workflows/build-and-run.yml
```diff
@@ -42,6 +42,16 @@ jobs:
...
+        include:
+          - os:
+              label: Linux
+              runner: ubuntu-latest
+              ctest_exclude: -E ".*($ALL_EXCLUDE)"
+            deps:
+              label: Minimum Python supported
+              env-file: environment.yml
+              extra-specs: |
+                python=3.9
```
gmgunter: Is there a way to specify default matrix arguments in GHA? It's a bit frustrating to need to copy-paste 7 lines each time we add a new min-reqs job.

gmgunter (commented):
One comment from our review together this morning. Otherwise, LGTM.

.github/workflows/build-and-run.yml
```diff
@@ -29,23 +34,26 @@ jobs:
...
         build-type: [ Release, Debug ]
-
+        include:
+          - os:
+              label: Linux
+            deps:
+              label: Minimum Python supported
+              extra-specs: |
+                python=3.9
+            build-type: RelWithDebInfo
```
gmgunter: Let's define a `build-type` env variable above and default it to this value.

That way, when we add additional min-reqs jobs, we won't have to copy-paste this line for each.

.github/workflows/build-and-run.yml
```diff
@@ -29,23 +34,26 @@ jobs:
...
             # background reader/writer test sometimes fails on macOS due to slow runtimes.
         deps:
           - label: Latest
-            env-file: environment.yml
-            extra-specs: ""
         build-type: [ Release, Debug ]
-
+        include:
+          - os:
+              label: Linux
```
gmgunter: Is there are way to default `os.label` to Linux? So that we don't need to specify it explicitly whenever we add a new min-reqs job?

gmgunter: Maybe we could just get rid of `os.label` (and `deps.label`). These currently are only used to annotate jobs with a description of their behavior.

Maybe a better approach would be to manually label jobs using a parameter called e.g. `action-label`. Any jobs that aren't manually labelled could fall back to using a label which is some combination of the runner type and build type.

gmgunter (commented):
Reviewed this offline together

.github/workflows/build-and-run.yml
rtburns-jpl: I wasn't really able to figure out a good way to have a defaultable action-label, because `env` isn't available when evaluating the job name. But I think the way it is currently might be OK actually, because expanding an empty os.label just evaluates to an empty string. So I dropped the os.label specifier here, and it just uses the deps.label instead, so the job names seem to look OK and will only require specifying either os.label or deps.label for future jobs.

gmgunter: Neat, that works pretty well -- I'm happy with that solution, thanks!

gmgunter (approved):
LGTM, thanks!

rad-eng-59:
Hi @rtburns-jpl and @gmgunter. Recently after merging more recent updates from develop branch, I can not push my updates to my isce3 fork on public github,  'https://github.com/rad-eng-59/isce3.git'. 
Here is the error on my local machine complaining about files under `.github/` wule trying to push an update to my fork:
```
To https://github.com/rad-eng-59/isce3.git
 ! [remote rejected]     noise-rcid1-pub -> noise-rcid1-pub (refusing to allow a Personal Access Token to create or update workflow `.github/workflows/build-and-run.yml` without `workflow` scope)
error: failed to push some refs to 'https://github.com/rad-eng-59/isce3.git'
```
FYI, I have been using classic personal token to push update to isce3 on public github so far w/o any problem. The token expires next year.
Any idea?
Thank you

rad-eng-59:
Please ignore my earlier comment. I have fixed it.

PR #93, opened 2025-08-14
Add optional X and Y spacing parameters to the `get_product_geometry` for the L2 product

xhuang-jpl:
This PR adds optional X and Y spacing parameters to `get_product_geometry` for the L2 product. Both parameters default to None, and when provided, the output spacing will use the specified values.


python/packages/nisar/workflows/get_product_geometry.py
```diff
@@ -213,6 +227,17 @@ def get_radar_grid(nisar_product_obj, args):
     geogrid_obj = grid_obj.geogrid
     wavelength = grid_obj.wavelength
 
+    # Update the Geogrid according to the specified spacing x and y
+    if (args.spacing_x is not None) and (args.spacing_y is not None):
```
gmgunter: What if `args.spacing_x` is not None but `args.spacing_y` *is* None? (Or vice versa)

xhuang-jpl: Good catch @gmgunter , in that way, it will compute the parameters at the same resolution with GCOV.

xhuang-jpl: @gmgunter  I have separated spacing_x and spacing_y, and any parameter set to None will default to GCOV’s geogrid values.

python/packages/nisar/workflows/get_product_geometry.py
```diff
@@ -213,6 +227,19 @@ def get_radar_grid(nisar_product_obj, args):
     geogrid_obj = grid_obj.geogrid
     wavelength = grid_obj.wavelength
 
+    # Update the Geogrid according to the specified spacing x and y
+    if args.spacing_x is not None:
+        geogrid_obj.width = int((geogrid_obj.end_x -
+                                 geogrid_obj.start_x)/
+                                args.spacing_x + 0.5)
```
hfattahi: @xhuang-jpl why do you need +0.5 ? doesn't make sense to me.

python/packages/nisar/workflows/get_product_geometry.py
```diff
@@ -213,6 +227,19 @@ def get_radar_grid(nisar_product_obj, args):
...
+        geogrid_obj.spacing_x = args.spacing_x
+
+    if args.spacing_y is not None:
+        geogrid_obj.length = int((geogrid_obj.end_y -
+                                  geogrid_obj.start_y)/
+                                 -abs(args.spacing_y) + 0.5)
```
hfattahi: same as above what is +0.5 doing here

python/packages/nisar/workflows/get_product_geometry.py
xhuang-jpl: Thanks @hfattahi, I have removed the +0.5 and use the `np.ceil `function instead to recompute the new width and length.

xhuang-jpl:
@gmgunter  and @hfattahi could you take another look? Thank you

python/packages/nisar/workflows/get_product_geometry.py
```diff
@@ -52,6 +52,20 @@ def get_parser():
                         choices=['A', 'B'],
                         help='Frequency band: "A" or "B"')
 
+    parser.add_argument('--spacing-y',
+                        dest='spacing_y',
+                        required=False,
+                        default=None,
+                        type=float,
+                        help='Y spacing, the same unit with the L2 product (default: None)')
```
gmgunter: What if the input is an L1 product?

Not sure how the user of this interface is supposed to know what `default: None` means.

```suggestion
                        help='Y spacing, the same unit with the L2 product (only'
                        ' applicable if the input is a NISAR L2 product). Defaults'
                        ' to the Y-coordinate spacing of the product.')
```

python/packages/nisar/workflows/get_product_geometry.py
```diff
@@ -52,6 +52,20 @@ def get_parser():
...
+
+    parser.add_argument('--spacing-x',
+                        dest='spacing_x',
+                        required=False,
+                        default=None,
+                        type=float,
+                        help='X spacing, the same unit with the L2 product (default: None)')
```
gmgunter:
```suggestion
                        help='X spacing, the same unit with the L2 product (only'
                        ' applicable if the input is a NISAR L2 product). Defaults'
                        ' to the X-coordinate spacing of the product.')
```

python/packages/nisar/workflows/get_product_geometry.py
```diff
@@ -213,6 +227,19 @@ def get_radar_grid(nisar_product_obj, args):
     geogrid_obj = grid_obj.geogrid
     wavelength = grid_obj.wavelength
 
+    # Update the Geogrid according to the specified spacing x and y
+    if args.spacing_x is not None:
+        geogrid_obj.width = int(np.ceil((geogrid_obj.end_x -
```
gmgunter, line 267: Won't this cause the output geo grid to extend past the endpoint of the original geo grid? What happens in that case? Shouldn't we use `np.floor` instead of `np.ceil` to avoid this?

python/packages/nisar/workflows/get_product_geometry.py
xhuang-jpl: those parameters are implemented under the L2 product, so it will not have impact on the L1 product. In addition, L1 products grid normally do not use the x and y spacing convention, right? I like your change of the help information.

python/packages/nisar/workflows/get_product_geometry.py
xhuang-jpl: That is true. Since those are the produce geometric parameters, I think its geogrid should cover the L2 product geogrid to avoid the case that the last pixes of each row in L2 product cannot find its corresponding geometric parameters.

python/packages/nisar/workflows/get_product_geometry.py
```diff
@@ -52,6 +52,24 @@ def get_parser():
...
+                        default=None,
+                        type=float,
+                        help='Y spacing, the same unit with the L2 product (only'
+                        ' applicable if the input is a NISAR L2 product). Defaults'
+                        ' to the Y-coordinate spacing of the product.')
+
+    parser.add_argument('--spacing-x',
+                        dest='spacing_x',
+                        required=False,
+                        default=None,
```
gshiroma, line 68:
```suggestion
```
Same here.

python/packages/nisar/workflows/get_product_geometry.py
```diff
@@ -52,6 +52,24 @@ def get_parser():
                         choices=['A', 'B'],
                         help='Frequency band: "A" or "B"')
 
+    parser.add_argument('--spacing-y',
+                        dest='spacing_y',
+                        required=False,
+                        default=None,
```
gshiroma, line 58:
```suggestion
```
These lines are not needed when `required` is `False` and `default` is `None`.

python/packages/nisar/workflows/get_product_geometry.py
```diff
@@ -52,6 +52,24 @@ def get_parser():
...
+                        type=float,
+                        help='Y spacing, the same unit with the L2 product (only'
+                        ' applicable if the input is a NISAR L2 product). Defaults'
+                        ' to the Y-coordinate spacing of the product.')
```
gshiroma:
```suggestion
                        help='Y spacing, in the same unit as the L2 product. Can'
                        ' be positive or negative. This argument is only'
                        ' applicable if the input is a NISAR L2 product. Defaults'
                        ' to the Y-coordinate spacing of the product.')
```

Just adding that Y spacing can be positive or negative.

gshiroma (approved):
LGTM! Thank you, @xhuang-jpl . just have a couple of minor comments below.

python/packages/nisar/workflows/get_product_geometry.py
xhuang-jpl: Thank you @gshiroma

python/packages/nisar/workflows/get_product_geometry.py
xhuang-jpl: I think they are necessary because we have explicitly used those parameters in the our code,  we should keep consistent in case the `add_argument` function might be updated in future.

xhuang-jpl:
@gmgunter and @hfattahi can you take a look this PR? Thanks

python/packages/nisar/workflows/get_product_geometry.py
gmgunter: > I think its geogrid should cover the L2 product geogrid to avoid the case that the last pixes of each row in L2 product cannot find its corresponding geometric parameters.

I'm not sure I understand what you mean. Are you saying that the new geo grid (with the user-specified spacing) needs to extend beyond the border of geo grid of the L2 product itself?

I guess I don't understand the context. What are you planning to do with the output of `get_product_geometry.py`?

xhuang-jpl: sorry for the confusion. This plot might help you understand my points.

image

gmgunter: Thanks, @xhuang-jpl. I think I understand correctly that the output product frame will extend beyond the GCOV frame if you use `np.ceil`.

What I don't understand is why that's desirable. What are you using the output product for? Why does it need to cover the full GCOV frame? I need some context here.

xhuang-jpl: The soil moisture SAS needs the Local Incidence Angle as the input. However, the LIA generation using the full resolution GCOV geogrid is very time consuming (i.e., GCOV resolutions at 10m and 20m are too high), so we decided to use a coarser LIA since the SM product is on a 200m grid.  If we use the np.floor, some pixels in the above plot might not find their the corresponding LIA.

gmgunter: I see, thanks. Will you create the LIA layer on exactly the same grid as the SM product? Or will you need to interpolate it to the SM product grid?

If you plan to interpolate, don't you need additional margin around the edges of the grid, depending on the width of the interpolation kernel? Currently you only have enough margin for bilinear or nearest neighbor interpolation.

gshiroma: That's a good point, @gmgunter . @xhuang-jpl , if you want to match the grid of your multilooked GCOV product exactly, you can also pass the coordinates of the bounding box.

gshiroma: If you your multilooking window has 5 pixels in the X direction and you have 3 extra pixels in the right edge. Depending on the multilooking algorithm you're using, it may just ignore those 3 pixels. This will make the multilooked GCOV image to be a bit smaller than your local-incidence angle image. Of course, you can update the SM workflow to handle these extra local-incidence angle pixels. Not sure what you prefer.

xhuang-jpl: > I see, thanks. Will you create the LIA layer on exactly the same grid as the SM product? Or will you need to interpolate it to the SM product grid?
> 
> If you plan to interpolate, don't you need additional margin around the edges of the grid, depending on the width of the interpolation kernel? Currently you only have enough margin for bilinear or nearest neighbor interpolation.

Thank you @gmgunter,  the LIA will be the same projection with GCOV, and then convert to the EASEGRID 2.0 projection at 200 meters.  If we could get the LIA over the missing pixels, shouldn't the interpolation perform better? what is your thoughts that we should not use `np.ceil`?

gmgunter: Sorry for the confusion -- I'm not suggesting that you should use `np.floor` instead of `np.ceil` anymore. I didn't understand exactly what you were doing when I made my initial comment. I see why you're using `np.ceil` now.

Now I'm wondering instead if you need to add some additional margin when determining the geo grid that you want to compute the LIA data on.

What type of interpolation are you planning to use? If you're certain that the SM product grid is contained entirely within the GCOV grid, and you're using bilinear or nearest neighbor interpolation, then the existing implementation seems fine to me -- just using `np.ceil` will be enough to ensure you can interpolate the LIA data onto the SM grid without missing data.

But if you're using a higher order interpolator, you may need to create a LIA raster with a few pixels of extra margin around the border, beyond the extents of the original GCOV grid. The number of extra pixels depends on which interpolator you plan to use.

Also, since the GCOV product and the SM product are in two different projections, it can be tricky to guarantee that the SM coordinate grid is entirely contained within the GCOV coordinate grid. Even if all four corners of the SM grid are inside the bounds of the GCOV grid, there may still be some pixels in the SM grid that fall slightly outside the GCOV grid if the transformation between the two coordinate systems is not affine.

So you may need to add some margin when computing the LIA layer, either by passing `--margin-x` and `--margin-y` flags to this script or by passing the coordinates of a bounding box, as @gshiroma suggested above. (The bounding box would define the extents of the geo grid on which to compute the local incidence angle data.)

xhuang-jpl: Thank you @gmgunter  for the comments. We are using an  aggregation method (i.e. mean) instead of the nearest or bilinear interpolation, but I agree with your points that adding some margins could ensure that we have sufficient points because of the projection transformation.

xhuang-jpl: @gmgunter would you mind taking another look? I have added the margin to x and y. Thank you

gmgunter (approved):
LGTM, thanks for addressing my comments!

PR #96, opened 2025-08-14
Add the water mask source and update the L2 InSAR product mask attributes

xhuang-jpl:
This PR is to add the water mask source and update the L2 InSAR product mask attributes to response to the [request](https://github-fn.jpl.nasa.gov/NISAR-ADT/NISAR_PIX/pull/327/files).


share/nisar/defaults/insar.yaml
```diff
@@ -31,6 +31,9 @@ runconfig:
             dem_file_description:
 
             water_mask_file:
+            # PLACEHOLDER - Water mask file description, optional
```
gmgunter: What does "PLACEHOLDER" mean in this context?

gmgunter (approved):
LGTM, thank you!

hfattahi (approved):
LGTM.

hfattahi:
@vbrancat please take a look at this PR.

vbrancat (approved):
LGTM!

share/nisar/defaults/insar.yaml
xhuang-jpl: a ha, I simply copied the description from the dem_file_description.

nemo794: FWIW, in GCOV we removed the word "PLACEHOLDER" (because its meaning is slightly ambiguous), and instead used "NOT YET IMPLEMENTED".

https://github.com/isce-framework/isce3/blob/9c5887161eb8d24317a7ed2d107a43fe9c4a2497/share/nisar/defaults/gcov.yaml#L65

However, via this PR, I think this parameter -- and the corresponding water mask description parameter -- are now implemented... so let's remove the "PLACEHOLDER" language?

xhuang-jpl: I agree, will remove it for both DEM and Water Mask

python/packages/nisar/products/insar/InSAR_base_writer.py
```diff
@@ -140,6 +140,10 @@ def __init__(self,
         if self.dem_source is None:
             self.dem_source = "None"
 
+        self.water_mask_source = ancillary_group["water_mask_file_description"]
+        if self.water_mask_source is None:
+            self.water_mask_source = "None"
```
nemo794: Should this be "None" or "(NOT SPECIFIED)"?

gmgunter (approved):
LGTM, thanks for updating the comments!

python/packages/nisar/products/insar/InSAR_base_writer.py
xhuang-jpl: @vbrancat what do you think? I think either one will be fine.

hfattahi: I like NOT SPECIFIED better if possible

nemo794 (approved):
Thanks, @xhuang-jpl !

PR #97, opened 2025-08-15
Expose `rtc_sigma0_raster` in the `GeocodeCov` interface

gshiroma:
This PR exposes the `rtc_sigma0_raster option` to the `GeocodeCov` interface, allowing users to save memory, at the cost of a lower performance, by writing the raster directly to disk instead of storing it in the memory.


gshiroma:
@gmgunter and @hfattahi , I copied this PR from the internal repo. Would you mind approving it here too? Thanks!

cxx/isce3/geocode/GeocodeCov.cpp
```diff
@@ -480,7 +484,7 @@ void Geocode<T>::geocodeInterp(
             else
                 rtc_memory_mode = isce3::core::MemoryModeBlocksY::MultipleBlocksY;
 
-            if (out_geo_rtc_gamma0_to_sigma0 != nullptr) {
+            if (out_geo_rtc_gamma0_to_sigma0 != nullptr && output_rtc_sigma == nullptr) {
                 std::string vsimem_ref = (
                     "/vsimem/" + getTempString("geocode_cov_areaproj_rtc_sigma0"));
```
gmgunter: Looks like you might've missed this change when copying from the internal repo.

```suggestion
                    "/vsimem/" + getTempString("geocode_cov_interp_rtc_sigma0"));
```

gmgunter (approved):
LGTM, thanks!

Note: this is a copy of https://github-fn.jpl.nasa.gov/isce-3/isce/pull/2213, which was reviewed and approved internally.

cxx/isce3/geocode/GeocodeCov.cpp
gshiroma: Good catch. I forgot to run `pull` the changes before pushing them to the public repo. Thanks!

hfattahi (approved):
LGTM. Reviewed on internal repo.

PR #99, opened 2025-08-16
Add a function to compute a radar grid that spans an input geocoded grid (MCR #90132)

gmgunter:
Adds a new function `isce3.geometry.get_bounding_radar_grid()` that returns a `RadarGridParameters` object that spans an input `GeoGridParameters` object. The output radar grid is the smallest radar grid that contains all points within the input geocoded grid over a user-specified height range. Additional margin may be added to extend the radar grid azimuth & range extents.

This functionality is useful for implementing the static layers workflow. Many of the layers in the static layers product are computed from an input radar grid, but the input to the static layers workflow is a geo grid -- not a radar grid. Therefore we will need the ability to compute a radar grid that spans the input geo grid.


gmgunter:
@gshiroma @nemo794 this is a copy of a PR that was approved on the internal GitHub server: https://github-fn.jpl.nasa.gov/isce-3/isce/pull/2097

gshiroma (approved):
LGTM.  This PR was reviewed internally. Thank you, @gmgunter . Just make sure that the unit tests pass.

nemo794 (approved):
Ditto @gshiroma . Thanks, @gmgunter !

PR #100, opened 2025-08-19
Populate GCOV & GSLC Static Layers Data Access

gshiroma:
This PR populates GCOV & GSLC static layers data access template provided in the runconfig field `static_layers_data_access` and saved it as the product's H5 dataset `staticLayersDataAccess`. The template is updated by replacing the placeholder "{granule_id}" with the product's granule ID. An exception will be raised if the placeholder "{granule_id}" is present in staticLayersDataAccess` but the `partial_granule_id` is empty.


python/packages/nisar/products/writers/BaseL2WriterSingleInput.py
```diff
@@ -808,11 +808,47 @@ def populate_identification_l2_specific(self):
...
+                             ' the runconfig field'
+                             ' "static_layers_data_access", but the'
+                             ' field "partial_granule_id" was not provided')
+                error_channel = journal.error('get_static_layers_data_access')
+                error_channel.log(error_msg)
+                raise NotImplementedError(error_msg)
+
+            static_layers_data_access = \
+                static_layers_data_access.replace('{granule_id}',
+                                                  self.granule_id)
```
nemo794: granule IDs should not include the file extension, e.g. ".h5". Tecently, you mentioned issues with PGE/PCM including the ".h5" extension in the partial granule ID.

They should fix that, but to be safe for this function, please strip any file extensions to ensure that the granule ID does **not** include the file extension. We do not want the file extension to be populated into the link; otherwise, it will result in invalid links.

python/packages/nisar/products/writers/BaseL2WriterSingleInput.py
```diff
@@ -808,11 +808,47 @@ def populate_identification_l2_specific(self):
             'identification/platformName',
             default='(NOT SPECIFIED)')
 
+    def get_static_layers_data_access(self):
```
nemo794: As discussed this morning during the ISCE3 meeting, please make this a standalone function that takes `template_string` and `granule_id` as input arguments?

If this new function is unable to populate the provided template string, please have it return default value of e.g. "(NOT SPECIFIED)".

That way, InSAR can reuse the same function with the same behavior, and we can avoid copy-paste between two PRs.

Thanks!

nemo794 (commented):
As discussed during ISCE3 meeting this morning...

python/packages/nisar/products/writers/BaseL2WriterSingleInput.py
```diff
@@ -42,6 +42,52 @@
...
+    """
+    Read static layers data access template from the input runconfig field
+    `static_layers_data_access` and replace placeholder(s) with
+    the actual values.
+
+    Parameters
+    ----------
+    static_layers_data_access_runconfig: scalar
+        Value from runconfig field
+        `ceos_analysis_ready_data.static_layers_data_access`
```
nemo794: Let's keep this function generic -- there is nothing about this function which requires the template string to come from the runconfig, nor is there anywhere in this function which requires it to come from a particular place in the runconfig.

While cleaning that up, let's also make the docstring easier for future users to understand what is expected.

Also, please update the type annotation to `str or None`. (I don't think `scalar` is a Python type..?)

```suggestion
    template_string: str or None
        Template string for the static layers data access. This string must include the substring '{granule_id}'.
        Example: "https://my.template.string/{granule_id}.h5"
        If None, or if the required substring is missing, this function will return "(NOT SPECIFIED)".
```

python/packages/nisar/products/writers/BaseL2WriterSingleInput.py
```diff
@@ -42,6 +42,52 @@
...
+    granule_id: str
+        Product's granule ID
```
nemo794: Let's be more specific about what is expected from the user, and also let the user know what the behavior of the function will be.

```suggestion
    granule_id: str or None
        The granule ID, which will be used to replace the '{granule_id}' substring in `template_string`.
        Example: "NISAR_L2_PR_GCOV_088_039_D_110_2005_SHSH_A_20251114T221751_20251114T221827_T00407_N_F_J_001"
        If None, or if `granule_id` is "(NOT SPECIFIED)", this function will return '(NOT SPECIFIED)'.
```

python/packages/nisar/products/writers/BaseL2WriterSingleInput.py
```diff
@@ -42,6 +42,52 @@
...
+
+    Returns
+    -------
+    static_layers_data_access: str
+        An URL representing the static layers data access with all
+        placeholders replaced. Returns "(NOT SPECIFIED)" if
+        `static_layers_data_access_runconfig` is None or an empty
+        string.
```
nemo794: Let's remove the unnecessary reference to the runconfig. Let's also clearly represent what the function does so that there are no surprises for the user.

```suggestion
    static_layers_data_access: str
        Copy of `template_string` with the '{granule_id}' placeholder replaced.
        If function was unable to generate this string, an error
        is logged and a default of "(NOT SPECIFIED)" is returned.
```

python/packages/nisar/products/writers/BaseL2WriterSingleInput.py
```diff
@@ -42,6 +42,52 @@
 LUT_1D_RG_DATASETS_LUT_EXPANSION_ALONG_AZ_MARGIN_IN_PIXELS = 5
 
 
+def get_static_layers_data_access(static_layers_data_access_runconfig,
+                                  granule_id):
```
nemo794: Please add type annotations

python/packages/nisar/products/writers/BaseL2WriterSingleInput.py
```diff
@@ -42,6 +42,52 @@
...
+    """
+    Read static layers data access template from the input runconfig field
+    `static_layers_data_access` and replace placeholder(s) with
+    the actual values.
```
nemo794: Is this description still accurate...? (It's not. 😉) Please update per the new standalone function.

python/packages/nisar/products/writers/BaseL2WriterSingleInput.py
```diff
@@ -42,6 +42,52 @@
...
+    -------
+    static_layers_data_access: str
+        An URL representing the static layers data access with all
+        placeholders replaced. Returns "(NOT SPECIFIED)" if
+        `static_layers_data_access_runconfig` is None or an empty
+        string.
+    """
+
+    if not static_layers_data_access_runconfig:
+        return '(NOT SPECIFIED)'
```
nemo794: Let's simplify the logic of this function, and establish very clear cases and delineate the behavior for each case. Something like:

```suggestion
    default = "(NOT SPECIFIED)"
    if (template_string is None):
        log.error(...)
        return default
    elif (granule_id is None) or (granule_id == "(NOT SPECIFIED)"):
        log.error(...)
        return default
    elif "{granule_id}" not in template_string:
        log.error(..)
        return default
    else:
        return template_string.replace('{granule_id}', granule_id)
```

nemo794 (commented):
Thank you for breaking this out into a separate function! Let's make this function modular, reusable, and generic, improve the docstring, and simplify the logic. Thanks!

python/packages/nisar/products/writers/BaseL2WriterSingleInput.py
gshiroma: I don’t think it’s a good idea to require the user to include the `{granule_id}` substring. For example, when generating a custom product for a NISAR frame, the user may prefer to specify the granule ID directly in the runconfig without populating the `partial_granule_id` field. Similarly, if they generate their own static layers, they should be able to point to that product instead.

python/packages/nisar/products/writers/BaseL2WriterSingleInput.py
gshiroma: Sounds good. Just to note, in the case above the behavior should be to raise an error rather than return '(NOT SPECIFIED)'.

python/packages/nisar/products/writers/BaseL2WriterSingleInput.py
gshiroma: I can remove the references to the runconfig, but I think it’s useful to keep the error message explicit about the runconfig `static_layers_data_access parameter`, so it’s clear where the error originates.

python/packages/nisar/products/writers/BaseL2WriterSingleInput.py
gshiroma: This logic doesn’t look correct. Logging the error may cause pyre to raise an exception, which conflicts with the `return default`.

gshiroma:
> Thank you for breaking this out into a separate function! Let's make this function modular, reusable, and generic, improve the docstring, and simplify the logic. Thanks!

Thank you, @nemo794 , I addressed your comments. Please, let me know if you have any other comments or concerns.

python/packages/nisar/products/writers/BaseL2WriterSingleInput.py
```diff
@@ -42,6 +42,58 @@
...
+    """
+    Read the static layers data access template and replace the placeholder
+     "{granule_id}" with the product's granule ID.
+
+    Parameters
+    ----------
+    static_layers_data_access_template: str or None
+        Template string for static layers data access. The string may contain
+        the substring "{granule_id}", which will be replaced with the product’s
+        granule ID.
```
nemo794: As a standalone function, this function only knows about `static_layers_data_access_template` and `granule_id`.  This function has no way of knowing what is the "**product's** granule ID". Let's remove reference to the product.

```suggestion
        the substring "{granule_id}", which will be replaced with `granule_id`.
```

python/packages/nisar/products/writers/BaseL2WriterSingleInput.py
```diff
@@ -808,12 +860,17 @@ def populate_identification_l2_specific(self):
...
 
-        self.copy_from_runconfig(
-            '{PRODUCT}/metadata/ceosAnalysisReadyData/staticLayersDataAccess',
-            'ceos_analysis_ready_data/static_layers_data_access',
-            default='(NOT SPECIFIED)')
+        static_layers_data_access = get_static_layers_data_access(
+            static_layers_data_access_runconfig, self.granule_id)
+
+        self.set_value(
+            'identification/staticLayersDataAccess',
```
nemo794, line 819: Thank you for moving this to the `identification` Group so quickly!

python/packages/nisar/products/writers/BaseL2WriterSingleInput.py
nemo794: Oof, I didn't know that about pyre. Fair enough...

Could we make these warnings then?

nemo794: More importantly, there are edge cases missing from the updated code.

Here's one example: What happens if the template string "https://my.static.link/granule_id.h5" is passed in, and also the user provides an actual granule ID? If I'm the user, I would expect the function to make use of the granule ID. Instead, it silently ignores it.

Later on, someone will look at the HDF5 file and wonder why the Dataset is populated with "https://my.static.link/granule_id.h5". There is no indication of the SAS' behavior in either the docstring nor the log output, so this will be very hard to debug.

python/packages/nisar/products/writers/BaseL2WriterSingleInput.py
nemo794: Let's define the behavior more clearly.

```suggestion
        Template string for static layers data access. If the string contains
        the substring "{granule_id}", that substring will be replaced with the product’s
        granule ID.
```

nemo794 (commented):
Some additional comments.

python/packages/nisar/products/writers/BaseL2WriterSingleInput.py
gshiroma: Thank you, @nemo794 , I don't think we need to force the user to include the substring `{granule_id}`. Notice that they can provide the final URL or even a DOI. 

If they make a mistake and forget the curly braces, I don't think it should be the responsibility of the SAS to try to figure this out. Notice that the "replace" behavior  is the same strategy that we used for populating the partial granule ID (example [here](https://github.com/isce-framework/isce3/blob/711ccd8798c61b229b97022d726ddf4305fa72e3/python/packages/nisar/products/writers/BaseWriterSingleInput.py#L132), [here](https://github.com/isce-framework/isce3/blob/711ccd8798c61b229b97022d726ddf4305fa72e3/python/packages/nisar/products/writers/SLC.py#L346), and [here](https://github.com/isce-framework/isce3/blob/711ccd8798c61b229b97022d726ddf4305fa72e3/python/packages/nisar/products/insar/granule_id.py#L239)), and also OPERA RTC-S1 [static_layers_data_access](https://github.com/opera-adt/RTC/blob/c5cc7403af16bb67fd14f43bf2aec15c6a422297/src/rtc/h5_prep.py#L485) and [product_id](https://github.com/opera-adt/RTC/blob/c5cc7403af16bb67fd14f43bf2aec15c6a422297/src/rtc/rtc_s1_single_job.py#L59).

python/packages/nisar/products/writers/BaseL2WriterSingleInput.py
gshiroma: Thank you, @nemo794 . To me they mean the same thing, but I like your wording better. Before I change it, are you okay with  "product's granule ID" in the end of your sentence? From your first comment, it seemed that you wanted me to remove the word "product's".

gshiroma: How about this one:
```
        Template string for static layers data access. If the string contains
        the substring "{granule_id}", that substring will be replaced with
        the contents of `granule_id`.
```

python/packages/nisar/products/writers/BaseL2WriterSingleInput.py
nemo794: Let me revise the example: what happens if I pass in the template string "https://my.static.link/static.h5", and I also pass in an actual granule ID? As the user, I would expect the function to make use of the granule ID I'm intentionally providing as an input parameter. Instead, the function as-is ignores the granule ID silently.

python/packages/nisar/products/writers/BaseL2WriterSingleInput.py
nemo794: That works, thanks.

python/packages/nisar/products/writers/BaseL2WriterSingleInput.py
```diff
@@ -42,6 +42,58 @@
...
+        Template string for static layers data access. The string may contain
+        the substring "{granule_id}", which will be replaced with the product’s
+        granule ID.
+        If set to `None` or an empty string, the function returns
+        "(NOT SPECIFIED)"
+    granule_id: str or None
+        The granule ID, which will be used to replace the substring
+        "{granule_id}" in `static_layers_data_access_template`.
+        If the granule ID is required but not provided (i.e., `None`,
+         empty string, or "(NOT SPECIFIED)"), the function raises an error.
```
nemo794: What do you mean by "required"? Can we be more specific about the behavior, perhaps something like this instead?

```suggestion
        If the template string contains "{granule_id}", but `granule_id` is None,
         empty string, or "(NOT SPECIFIED)", the function raises an error.
```

python/packages/nisar/products/writers/BaseL2WriterSingleInput.py
gshiroma: Your example is fine. There's no need to substitute anything there. In the output product you'd have: `staticLayersDataAccess = b"https://my.static.link/static.h5"`

python/packages/nisar/products/writers/BaseL2WriterSingleInput.py
```diff
@@ -42,6 +42,58 @@
...
+                         " or is invalid")
+            error_channel = journal.error('get_static_layers_data_access')
+            error_channel.log(error_msg)
+            raise ValueError(error_msg)
+
+        static_layers_data_access = \
+            static_layers_data_access_template.replace('{granule_id}',
+                                                       granule_id)
+
+    return static_layers_data_access
```
nemo794: A small suggestion, to:
1) Remove a variable
2) Improve code grok-ability: providing an explicit `else:` case makes it clear that if the substring is not present, no modifications were made to the template string.

```suggestion
    if '{granule_id}' in static_layers_data_access_template:
        if not granule_id or granule_id == '(NOT SPECIFIED)':
            error_msg = ('The placeholder "{granule_id}" is included in'
                         ' the static layers data access template,'
                         ' but the `granule_id` was not provided'
                         " or is invalid")
            error_channel = journal.error('get_static_layers_data_access')
            error_channel.log(error_msg)
            raise ValueError(error_msg)

        return static_layers_data_access_template.replace('{granule_id}',
                                                       granule_id)
    else:
        return static_layers_data_access_template
```

python/packages/nisar/products/writers/BaseL2WriterSingleInput.py
```diff
@@ -42,6 +42,58 @@
...
+
+    Parameters
+    ----------
+    static_layers_data_access_template: str or None
+        Template string for static layers data access. The string may contain
+        the substring "{granule_id}", which will be replaced with the product’s
+        granule ID.
+        If set to `None` or an empty string, the function returns
+        "(NOT SPECIFIED)"
+    granule_id: str or None
```
nemo794: Let's also include a comment that if the template string is valid but does NOT contain the substring "{granule_id}", then `granule_id` will be ignored.

python/packages/nisar/products/writers/BaseL2WriterSingleInput.py
nemo794: extra space

```suggestion
        If the granule ID is required but not provided (i.e., `None`,
        empty string, or "(NOT SPECIFIED)"), the function raises an error.
```

nemo794 (commented):
Some additional thoughts

python/packages/nisar/products/writers/BaseL2WriterSingleInput.py
gshiroma: Thank you, @nemo794 , I removed "product's" from "product's granule ID".

python/packages/nisar/products/writers/BaseL2WriterSingleInput.py
gshiroma: Thank you, @nemo794 . Please note that the separation from `static_layers_data_access_template` and `static_layers_data_access` was intentional. The function receives the `static_layers_data_access_template` and returns `static_layers_data_access`. 

I know a lot of people prefer to use less variables and lines in the code, but I tend to prefer better code legibility.  

For example, I'm not a fan of this line you suggested:
```
        return static_layers_data_access_template.replace('{granule_id}',
                                                       granule_id)
```

I think there's too much going on in a single line and I usually tend to break it multiple lines.

I also, prefer not to put a `return` in the `else` statement if that's not needed. I think it's an unnecessary, makes the code slightly more difficult to read (you need to ensure that the `if` above "returns" in all cases), and it's ugly (probably because I don't tend to do it).

python/packages/nisar/products/writers/BaseL2WriterSingleInput.py
gshiroma: I updated that in the commit https://github.com/isce-framework/isce3/pull/100/commits/74efc5bdf2c21a8bc8a3aec4582bcc3e32d212bd to to clarify that if `{granule_id}` is not in the template, `granule_id` is ignored.

share/nisar/defaults/gcov.yaml
```diff
@@ -134,6 +134,11 @@ runconfig:
             # the corresponding metadata field will be set to "(NOT SPECIFIED)".
             product_doi:
 
+            # Data access for the static layers product associated with the
+            # output product (URL or DOI). If not provided, the corresponding
+            # metadata field will be set to "(NOT SPECIFIED)".
```
nemo794, line 139: Let's complete this comment by making the remaining nominal behavior explicit:

```suggestion
            # metadata field will be set to "(NOT SPECIFIED)".
            # If this string contains the substring "{granule_id}", that substring
            # will be replaced with the granule ID.
```

nemo794 (commented):
Thank you for the updates to the docstring and explaining the behavior! One more minor comment...

python/packages/nisar/products/writers/BaseL2WriterSingleInput.py
```diff
@@ -42,6 +42,62 @@
 LUT_1D_RG_DATASETS_LUT_EXPANSION_ALONG_AZ_MARGIN_IN_PIXELS = 5
 
 
+def get_static_layers_data_access(
```
xhuang-jpl: Thank you @gshiroma , would it be possible put this function somewhere else as a common function so that the InSAR can use?

xhuang-jpl:
Thank you @gshiroma, if possible, please put the `get_static_layers_data_access`  funciton somewhere under the `isce/python/packages/nisar/products` folder as a common function?  Others look good to me.

python/packages/nisar/products/writers/BaseL2WriterSingleInput.py
gmgunter: If you're not sure where else to put it, I could imagine moving it to [python/packages/nisar/products/utils.py](https://github.com/isce-framework/isce3/blob/8101d85db2e1a2f115c79c3043d3ec0e132cc13b/python/packages/nisar/products/utils.py)

share/nisar/defaults/gcov.yaml
gshiroma: Thank you, @nemo794 , these comments were updated in the commit [aa218dc](https://github.com/isce-framework/isce3/pull/100/commits/aa218dc90fa9f58f8feaa58cd757b3f18593841d).

python/packages/nisar/products/writers/BaseL2WriterSingleInput.py
gshiroma: Thank you, @xhuang-jpl and @gmgunter , I moved the function `get_static_layers_data_access()` into [python/packages/nisar/products/utils.py](https://github.com/isce-framework/isce3/blob/8101d85db2e1a2f115c79c3043d3ec0e132cc13b/python/packages/nisar/products/utils.py) in the commit [2194f35](https://github.com/isce-framework/isce3/pull/100/commits/2194f35fa37e1f65b20335427f3668044acaad46)

python/packages/nisar/products/utils.py
```diff
@@ -16,4 +17,60 @@ def to_bytes(s: str | ArrayLike) -> np.ndarray:
...
+        If set to `None` or an empty string, the function returns
+        "(NOT SPECIFIED)"
+        If the string is valid, i.e., it is not `None` or an empty string,
+        and it does not contain the substring "{granule_id}", then
+        `granule_id` is ignored.
+    granule_id: str or None
+        The granule ID, which will be used to replace the substring
+        "{granule_id}" in `static_layers_data_access_template`.
+        If the template string contains "{granule_id}", but `granule_id` is
+        `None`, empty string, or "(NOT SPECIFIED)"), the function raises an
```
nemo794:
```suggestion
        `None`, empty string, or "(NOT SPECIFIED)", the function raises an
```

nemo794 (approved):
ack -- one minor typo, but otherwise LGTM. Thanks!

xhuang-jpl (approved):
LGTM, thank you @gshiroma

gshiroma:
Thank you, @nemo794 and @xhuang-jpl !

PR #101, opened 2025-08-19
Change the `instrumentName` from `LSAR` to `L-SAR` for InSAR product

xhuang-jpl:
This PR is to change the instrumentName from LSAR to L-SAR for the InSAR product.

### DIFF
--- a/python/packages/nisar/products/insar/InSAR_base_writer.py
+++ b/python/packages/nisar/products/insar/InSAR_base_writer.py
@@ -1136,7 +1136,7 @@ class InSARBaseWriter(h5py.File):
             ),
             DatasetParams(
                 "instrumentName",
-                f"{radar_band_name}SAR",
+                f"{radar_band_name}-SAR",
                 (
                     "Name of the instrument used to collect the remote"
                     " sensing data provided in this product"


gmgunter (approved):
Thanks for the speedy update! LGTM!

nemo794 (approved):
Love it! Thank you, @xhuang-jpl !

python/packages/nisar/products/insar/InSAR_base_writer.py
```diff
@@ -1136,7 +1136,7 @@ def add_identification_to_hdf5(self):
             ),
             DatasetParams(
                 "instrumentName",
-                f"{radar_band_name}SAR",
+                f"{radar_band_name}-SAR",
```
nemo794, line 1139: In a future PR, perhaps consider updating this to copy from the input product, same as RSLC, GSLC, and GCOV.

Reason: This current implementation will produce the correct output for NISAR datasets, but what happens for e.g. ALOS PALSAR? In that case, RSLC/GSLC/GCOV will populate this with "PALSAR", but InSAR would populate it with "L-SAR".

xhuang-jpl: I remember the decision has been made 2 years ago that InSAR should compute it itself to use either `L` or `S` based on its frequency, but I cannot remember the reason.

PR #102, opened 2025-08-19
Fix GSLC statitstics filled with zeros

seongsujeong:
This PR corresponds to the PR #2136 in the internal GitHub. This PR was tested with the [NISAR sample data](https://science.nasa.gov/mission/nisar/sample-data/) by comparing the stats in the attributes with what was computed by numpy. The difference was less than 0.002%.


python/packages/nisar/workflows/gslc.py
```diff
@@ -198,8 +203,9 @@ def run(cfg):
                                           subswaths=sub_swaths)
 
                 # write geocoded blocks to respective HDF5 datasets
-                for gslc_dataset, gslc_data_blk in zip(gslc_datasets,
-                                                       gslc_data_blks):
+                for stats_complex, gslc_dataset, gslc_data_blk in zip(stats_complex_list,
+                                                                      gslc_datasets,
+                                                                      gslc_data_blks):
```
gshiroma, line 208: It looks like this for loop iterates over the number of data blocks for each polarization, while the length of `stats_complex_list` corresponds to the number of polarizations (not the number of GSLC data blocks).

I think this change was introduced when you moved the statistics computation after the mantissa truncation, which is fine, but you probably need to adjust which element you’re pulling from `stats_complex_list`.

gshiroma: @seongsujeong clarified offline that this for loop iterates over each polarizations, not data blocks as I thought. Each element of the list `gslc_data_blks` correspond to the same slice of a different polarization. Thank you for the clarification, @seongsujeong . I'm resolving this comment and approving this PR.

gshiroma (approved):
LGTM. We reviewed this PR internally. I had one additional comment, which @seongsujeong addressed it. Thanks!

hfattahi (approved):
LGTM.

PR #103, opened 2025-08-20
Remove unused `Topo::topo()` overload (MCR NOT NEEDED)

gmgunter:
Removes an overload of `Topo::topo()` that takes a `DEMInterpolator` object instead of taking the input DEM as a `Raster` object.

The implementation of this overload is almost identical to the other overload, except that it doesn't load the DEM one-block-at-a-time. Instead, the user must load the entire region of interest into the `DEMInterpolator` before calling `Topo::topo()`.

The removed overload is unused in ISCE3 (it's not used anywhere in the C++ code and isn't exposed in the pybind11 bindings), except in a unit test for the metadata cubes. I've modified the unit test to instead create a fixed-height DEM raster to pass to `Topo::topo()`.

Removing this overload will make Topo easier to maintain in the future since there won't be a ton of copy-pasted code necessitating duplicated updates.


gmgunter:
@hfattahi @bhawkins this is a copy of a PR that you approved on the internal GitHub server: https://github-fn.jpl.nasa.gov/isce-3/isce/pull/2211

gshiroma (approved):
LGTM! Thank you, @gmgunter .

bhawkins (approved):

PR #104, opened 2025-08-20
Revise logic for bounding box computation when crossing antimeridian

seongsujeong:
This is an equivalent PR to PR #2162 in the internal GitHub, which was approved by @gshiroma and @gmgunter.


gshiroma (approved):
LGTM. We reviewed this PR internally. Thank you, @seongsujeong

gmgunter:
Hmm, is there any way these changes could've affected resample_slc?

It's very strange that the macOS release build failed due to the resample_slc unit test failing, but the Linux builds and macOS debug build are all passing :confused:

```
AssertionError: Phase standard deviation between ground-truth and resample output 10.062402725219727 is greater than maximum acceptable value 10.0
```

Some other workflow runs that launched around the same time this morning on other PR branches all passed ([1](https://github.com/isce-framework/isce3/actions/runs/17105333363), [2](https://github.com/isce-framework/isce3/actions/runs/17107326220)) so it seems like the issue is specific to this branch. But I'm not sure what would've caused the test to fail in this peculiar way.

seongsujeong:
Manually re-running the failed test allowed it to complete successfully. Not sure what happened in the initial run. Maybe the CI server was hit by cosmic ray? :$

gmgunter:
:sparkles::sparkles::sparkles:

gmgunter (approved):
LGTM, thanks!

PR #108, opened 2025-08-21
Add CEOS `staticLayersDataAccess` to L2 InSAR products

xhuang-jpl:
This PR is to add the staticLayersDataAccess to the L2 InSAR products to response to this PIX [PR# 330](https://github-fn.jpl.nasa.gov/NISAR-ADT/NISAR_PIX/pull/330/files),  and some functions such as  `get_static_layers_data_access` are copied from this PR #100 .


python/packages/nisar/products/insar/utils.py
```diff
@@ -573,3 +573,45 @@ def generate_insar_subswath_mask(ref_rslc_obj,
         (len(azi_idx_arr),
          len(rg_idx_arr))).astype(np.uint8)
 
+def get_static_layers_data_access(static_layers_data_access_runconfig,
```
nemo794: Once #100 is merged, please simply call that function. No need for duplicate code. :)

share/nisar/schemas/insar.yaml
```diff
@@ -47,6 +47,8 @@ runconfig:
             # TEC correction is applied
             tec_file: str(required=False)
 
+        ceos_analysis_ready_data: include('ceos_analysis_ready_data_options', required=False)
```
vbrancat: @nemo794 can you please comment why we are adding this in the SDS portion of the schema

vbrancat: Also, can we just simplify things and have one line called "`static_layers_data_access`". Not sure why we need to add a specific group for CEOS data

vbrancat (approved):
LGTM. @nemo794 can you please have a look?

share/nisar/schemas/insar.yaml
nemo794: My guess is that @xhuang-jpl simply mimicked what was done in GSLC/GCOV. Not sure the original reason. 🤷 Could probably move it?

xhuang-jpl: I have simplified it and now and  just added one line to the schema.

xhuang-jpl: after talking with @gshiroma , the `static_layers_data_access` should be in the same group with the `product_doi`

xhuang-jpl: @nemo794 could you review this PR again? Thank you

python/packages/nisar/products/insar/InSAR_base_writer.py
```diff
@@ -1167,7 +1172,7 @@ def add_identification_to_hdf5(self):
             ),
             DatasetParams(
                 "granuleId",
-                granule_id,
+                self.granule_id,
```
nemo794, line 1175: It looks like `self.granule_id` is being saved as the string "None", whereas this function is looking for the Python `None`. Could you please adjust to ensure the correct version of None is getting passed (if appropriate)? Thanks!

share/nisar/schemas/insar.yaml
```diff
@@ -76,6 +76,9 @@ runconfig:
             product_type: enum('GUNW', 'RIFG', 'RUNW', 'RIFG_RUNW_GUNW', 'ROFF_RIFG_RUNW_GOFF_GUNW', 'ROFF', 'GOFF', 'ROFF_GOFF')
             product_version: include('product_version_options', required=False)
             product_doi: include('product_doi_options', required=False)
+            # Data access for the static layers product associated with the output
+            # product (URL or DOI)
```
nemo794: We updated the GSLC/GCOV runconfig docstring in #100. Could you please update here to be similar?

```suggestion
            # Data access for the static layers product associated with the
            # output product (URL or DOI). If not provided, the corresponding
            # metadata field will be set to "(NOT SPECIFIED)".
            # If this string contains the substring "{granule_id}", that substring
            # will be replaced with the granule ID.
```

https://github.com/isce-framework/isce3/blob/fecf943053d18eddcac9ded8e68a5f2f767b9673/share/nisar/defaults/gcov.yaml#L137C1-L141C52

python/packages/nisar/products/insar/InSAR_base_writer.py
xhuang-jpl: Thank you @nemo794 , I used `(NOT SPECIFIED)` to keep consistent with the GCOV/GSLC instead of None

share/nisar/schemas/insar.yaml
xhuang-jpl: Fixed. Thank you

python/packages/nisar/products/insar/InSAR_base_writer.py
nemo794: Even better! Thank you, @xhuang-jpl !

nemo794 (approved):
LGTM! Great addition. Thank you, @xhuang-jpl !

PR #109, opened 2025-08-25
Increase the DEM margin for RTC in the Y direction

gshiroma:
A straight line in projected coordinates does not correspond to a straight line in geographic coordinates. As a result, deriving minimum and maximum latitude values directly from the geogrid corner X and Y extents may be inaccurate. To address this, we add a margin when loading the DEM subset.

For RTC processing, this issue is more pronounced in the Y direction. Since RTC does not use block processing along the X axis, the resulting blocks are very wide. In the middle of such blocks, the true minimum and maximum latitude values can differ significantly from those estimated using only the bounding-box corners.

The plot below shows the latitudes and longitudes of a straight horizontal line in UTM Zone 16N. Notice that the latitude changes even though the Y coordinate remain constant in UTM coordinates:

Screenshot 2025-08-25 at 4 45 57 PM

The example above is based on the geometry for a NISAR GCOV product and shows a discrepancy of about 0.04 deg. between the latitude estimated at the geogrid corners and at its center. At 70 deg. latitude, this corresponds to roughly 4.5 km, or about 150 DEM pixels assuming 30 m spacing.

To address this, we increase the margin in the Y direction from `100` to `200` pixels.


gshiroma:
Update runtime differences did not change significantly (2 runs for each value):
- 5 MHz single pol: -1.78% (faster)
- 20+5 MHz dual-pol dataset 1: +3.24% (slower)
- 20+5 MHz dual-pol dataset 2: -2.46% (faster)
- 40+5 MHz:  +1.03 % (slower)
- 80 MHz: -2.46 % (faster)

@hfattahi

hfattahi (approved):
LGTM

cxx/isce3/geometry/RTC.cpp
```diff
@@ -1167,8 +1167,23 @@ void _RunBlock(const int jmax, const int block_size,
...
+    inaccurate. To address this, we add a margin when loading the DEM subset.
+
+    For RTC processing, this issue is more pronounced in the Y direction. Since
+    RTC does not use block processing along the X axis, the resulting blocks
+    are very wide. In the middle of such blocks, the true minimum and maximum
+    latitude values can differ significantly from those estimated using only
+    the bounding-box corners.
+    */
+    int dem_margin_x_in_pixels = 100;
+    int dem_margin_y_in_pixels = 200;
```
seongsujeong, line 1183: Is the pixel from DEM or from GCOV? How much is this margin robust with different DEM / GCOV posting?

seongsujeong (commented):
LGTM overall, but I've left a question about this fix.
Also please let me know if you need a data to stress the algorithm. I can point you to a strip passing next to the south pole hole.

cxx/isce3/geometry/RTC.cpp
gshiroma: The margin is given in DEM pixels. It's robust to the parameters we have in production with a DEM sampled at `1` arcsec and products produced with DEM in geographic coordinates with latitudes under 70 latitude (see the PR description for details).

Notice that GCOV products processed with latitude beyond `65` degrees use the polar stereo DEM (see `stage_dem.py` code [here](https://github.com/isce-framework/isce3/blob/d4ad732710539b9425c1171c266db3cd02f67577/python/packages/nisar/workflows/stage_dem.py#L561)) and therefore are not affected by this issue. 

I thought about updating the code to be robust to any DEM and any latitude, but that would require significant changes to the function `loadDemFromProj()`. Basically, the function would traverse the block boundary and update the lat/lon min and max. Not sure if we want to do that? I'm fine with either options. @seongsujeong @hfattahi .

seongsujeong: Based on the time constraints, I think the simple fix like this will be desirable. I would suggest to submit an issue to keep track of this, and improve this down the road.

seongsujeong (approved):
LGTM

cxx/isce3/geometry/RTC.cpp
gshiroma: Sounds good. Thank you, @seongsujeong .

PR #110, opened 2025-08-26
nullify azimuth mean

bhawkins:
This adds an option to the RSLC workflow to remove a running mean in the azimuth direction. The intent is to remove the calibration tone, assuming it is stationary vs slow time for the duration of the block. For NISAR this is somewhat preferable to a range-frequency notch filter because
* The caltone frequency can be mode-dependent, and its aliased image after on-board DSP may also be mode-dependent even if the RF tone is not. It's nice to avoid the bookkeeping.
* The signal is not necessarily stationary in range. It may change abruptly across beam transitions or transmit gaps. This can cause filter artifacts.
* The azimuth DC component is not relevant to the image because the radar squint means it's not in the processing band.

A similar approach is to apply a cosine-weighted notch function to the azimuth. One nice feature of that approach is you can control the width of the notch, which is necessary if the caltone is not constant vs slow time. However, the proposed approach has a similar parameter: the length of the moving average. Thus is it can handle some variation as well, plus it has the benefit of continuity across image frame boundaries (e.g., for seamless mosaics of independently processed images).

@hb6688 You might find this helpful when testing RFI mitigation algorithms.

### DIFF
--- a/python/packages/nisar/workflows/focus.py
+++ b/python/packages/nisar/workflows/focus.py
@@ -1816,6 +1816,8 @@ def focus(runconfig, runconfig_path=""):
                 z[np.isnan(z)] = 0.0
                 if cfg.processing.zero_fill_gaps:
                     fill_gaps(z, swaths[:, pulse:pulse+nblock, :], 0.0)
+                if cfg.processing.nullify_azimuth_mean:
+                    z -= z.mean(axis=0)
                 raw_mm[block_out] = z
 
             raw_clean, rfi_likelihood = process_rfi(cfg, raw_mm, temp)
--- a/share/nisar/defaults/focus.yaml
+++ b/share/nisar/defaults/focus.yaml
@@ -415,6 +415,12 @@ runconfig:
             # cal pulses.
             zero_fill_gaps: True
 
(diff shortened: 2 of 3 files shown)


rad-eng-59 (approved):
LGTM. Thank you for the addition.
FYI, the Caltone is pretty table with no jumps as long as RX-window related parameters are fixed. In most NISAR modes, around every `N`-second the caltone starting phase jumps (not magnitude) and that lead to something like slow-time phase noise which can smear caltone in frequency domain a bit on top of existing jitter/phase noise (pretty low). That's why I like frequency domain approach to control the notch with with some smoothing windowing such as inverse of Hann window (no box filtering). One can also apply the slow-time filtering over all range lines (assumign not too large < 35 seconds) or larger portion to avoid any potential phase discontinuity.
I wonder small-block averaging can potentially introduce slow-time phase artifacts in AZ IPR or if the block size of the mean is different from one pass to another in interferometry.
However, I have a feeling neither of approaches, time and frequency domain can fully remove the caltone at the boundary issues every `N`-seconds but perhaps, the mitigation of around `~30 dB` is good enough.

hb6688 (approved):
@bhawkins LGTM

PR #112, opened 2025-08-26
Parallel stats

bhawkins:
In the RSLC workflow, the image statistics are computed by the CPU while the GPU focuses the next block. If the GPU is fast enough, the stats calculation can become a bottleneck. This PR enables computing stats in parallel.

The implementation is hidden behind macros to support compilers that don't support OpenMP. I got more than I bargained for after removing trailing whitespace, but I figure may as well leave that commit in.

This PR was approved internally [here](https://github-fn.jpl.nasa.gov/isce-3/isce/pull/2233) by @gmgunter and @gshiroma

### DIFF
--- a/cxx/isce3/math/Stats.cpp
+++ b/cxx/isce3/math/Stats.cpp
@@ -174,7 +174,7 @@ void StatsRealImag<T>::update(const std::complex<T>* values,
             #pragma omp single
             {
                 partial_stats.resize(threads);
-            } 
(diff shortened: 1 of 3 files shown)


gmgunter (approved):
LGTM, thanks!

gshiroma (approved):
LGTM. We reviewed this PR internally. Thank you, @bhawkins .

PR #113, opened 2025-08-26
Update QA v13.0.0 -> v14.0.0

nemo794:
This PR Updates QA version from v13.0.0 to v14.0.0 in the Dockerfile. It also updates the runconfig schemas for all eight L1/L2 science products.

Note: This is a preliminary QA release in anticipation of R5.0.0. There will be an additional QA release (likely a patch release) in the next week for R5.

### DIFF
--- a/share/nisar/defaults/gcov.yaml
+++ b/share/nisar/defaults/gcov.yaml
@@ -594,8 +594,9 @@ runconfig:
 
                 # True to raise an exception if one or more metadata LUTs
                 # contain all non-finite (e.g. Nan, +/- Inf) values, or if one or more
-                # z-dimension height layers in a 3D LUT has all non-finite values.
-                # False to quiet the exception (although it will still be logged).
+                # z-dimension height layers in a 3D LUT ("metadata cube") has
+                # all non-finite values. False to quiet the exception (although
+                # it will still be logged).
                 # Default: True
                 metadata_luts_fail_if_all_nan: true
 
--- a/share/nisar/defaults/gcov.yaml
+++ b/share/nisar/defaults/gcov.yaml
@@ -579,7 +579,8 @@ runconfig:
 
             software_config:
 
-                # True to use memory map(s) to cache select Dataset(s).
+                # True to cache selected dataset(s) into intermediate
+                # memory-mapped flat file(s), which speeds up repeat access.
                 # False to always read data directly from the input file.
                 # Generally, enabling caching should reduce runtime.
                 # Default: True
--- a/share/nisar/defaults/gcov.yaml
+++ b/share/nisar/defaults/gcov.yaml
@@ -577,14 +577,27 @@ runconfig:
                 # Default: True
                 qa_reports: true
 
+            software_config:
+
+                # True to use memory map(s) to cache select Dataset(s).
+                # False to always read data directly from the input file.
+                # Generally, enabling caching should reduce runtime.
+                # Default: True
+                use_cache: true
+
+                # True to delete the nested QA scratch directory in
+                # `scratch_path` and its contents when QA SAS is finished.
+                # Default: True
+                delete_qa_scratch_files: true
+
             validation:
 
-                # True to raise an exception if one or more metadata cubes contains
-                # all non-finite (e.g. Nan, +/- Inf) values, or if one or more
-                # z-dimension height layers in a 3D cube has all non-finite values.
+                # True to raise an exception if one or more metadata LUTs
+                # contain all non-finite (e.g. Nan, +/- Inf) values, or if one or more
+                # z-dimension height layers in a 3D LUT has all non-finite values.
                 # False to quiet the exception (although it will still be logged).
                 # Default: True
-                metadata_cubes_fail_if_all_nan: true
+                metadata_luts_fail_if_all_nan: true
 
             qa_reports:
 
--- a/share/nisar/defaults/focus.yaml
+++ b/share/nisar/defaults/focus.yaml
@@ -499,7 +499,8 @@ runconfig:
 
             software_config:
 
-                # True to use memory map(s) to cache select Dataset(s).
+                # True to cache selected dataset(s) into intermediate
+                # memory-mapped flat file(s), which speeds up repeat access.
                 # False to always read data directly from the input file.
                 # Generally, enabling caching should reduce runtime.
                 # Default: True
--- a/share/nisar/defaults/focus.yaml
+++ b/share/nisar/defaults/focus.yaml
@@ -497,14 +497,27 @@ runconfig:
                 # Default: True
                 noise_equivalent_backscatter: true
 
+            software_config:
+
+                # True to use memory map(s) to cache select Dataset(s).
+                # False to always read data directly from the input file.
+                # Generally, enabling caching should reduce runtime.
+                # Default: True
+                use_cache: true
+
+                # True to delete the nested QA scratch directory in
+                # `scratch_path` and its contents when QA SAS is finished.
+                # Default: True
+                delete_qa_scratch_files: true
+
             validation:
 
-                # True to raise an exception if one or more metadata cubes contains
-                # all non-finite (e.g. Nan, +/- Inf) values, or if one or more
-                # z-dimension height layers in a 3D cube has all non-finite values.
+                # True to raise an exception if one or more metadata LUTs
+                # contain all non-finite (e.g. Nan, +/- Inf) values, or if one or more
+                # z-dimension height layers in a 3D LUT has all non-finite values.
                 # False to quiet the exception (although it will still be logged).
                 # Default: True
-                metadata_cubes_fail_if_all_nan: true
+                metadata_luts_fail_if_all_nan: true
 
             qa_reports:
 
--- a/share/nisar/defaults/focus.yaml
+++ b/share/nisar/defaults/focus.yaml
@@ -514,8 +514,9 @@ runconfig:
 
                 # True to raise an exception if one or more metadata LUTs
                 # contain all non-finite (e.g. Nan, +/- Inf) values, or if one or more
-                # z-dimension height layers in a 3D LUT has all non-finite values.
-                # False to quiet the exception (although it will still be logged).
+                # z-dimension height layers in a 3D LUT ("metadata cube") has
+                # all non-finite values. False to quiet the exception (although
+                # it will still be logged).
                 # Default: True
                 metadata_luts_fail_if_all_nan: true
 
--- a/share/nisar/defaults/gslc.yaml
+++ b/share/nisar/defaults/gslc.yaml
@@ -362,14 +362,27 @@ runconfig:
                 # Default: True
                 point_target_analyzer: true
 
+            software_config:
+
+                # True to use memory map(s) to cache select Dataset(s).
+                # False to always read data directly from the input file.
+                # Generally, enabling caching should reduce runtime.
+                # Default: True
+                use_cache: true
+
+                # True to delete the nested QA scratch directory in
+                # `scratch_path` and its contents when QA SAS is finished.
+                # Default: True
+                delete_qa_scratch_files: true
+
             validation:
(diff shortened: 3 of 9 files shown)


gmgunter (approved):
LGTM, thanks!

share/nisar/defaults/gcov.yaml
```diff
@@ -570,14 +570,27 @@ runconfig:
                 # Default: True
                 qa_reports: true
 
+            software_config:
+
+                # True to use memory map(s) to cache select Dataset(s).
```
gshiroma:
```suggestion
                # True to use memory map(s) to cache select dataset(s).
```
Is there a specific reason "Dataset" is capitalized here? If not, we should lowercase it.

share/nisar/defaults/gcov.yaml
```diff
@@ -570,14 +570,27 @@ runconfig:
...
+                # False to always read data directly from the input file.
+                # Generally, enabling caching should reduce runtime.
+                # Default: True
+                use_cache: true
```
gshiroma, line 587: I don’t think this prevents the machine from using its own caching system. So, could we rename this field to something like `use_memory_map` for clarity? Also, as far as I know, memory mapping only works when the data is contiguous/linear ("binary"). GCOV imagery is compressed, so is it memory mappable? Does this option require any specific assumptions about the input file in order to be memory mapped?

share/nisar/defaults/gcov.yaml
```diff
@@ -570,14 +570,27 @@ runconfig:
...
-                # True to raise an exception if one or more metadata cubes contains
-                # all non-finite (e.g. Nan, +/- Inf) values, or if one or more
-                # z-dimension height layers in a 3D cube has all non-finite values.
+                # True to raise an exception if one or more metadata LUTs
+                # contain all non-finite (e.g. Nan, +/- Inf) values, or if one or more
+                # z-dimension height layers in a 3D LUT has all non-finite values.
                 # False to quiet the exception (although it will still be logged).
                 # Default: True
-                metadata_cubes_fail_if_all_nan: true
+                metadata_luts_fail_if_all_nan: true
```
gshiroma: Does this flag apply to both LUTs and metadata cubes? I understand metadata cubes are a type of 3D LUT, but that’s not the terminology we’ve used in our documentation, code, or this YAML file. So, it may be worth clarifying that this also applies to metadata cubes (or radar grid cubes, in the case of L2 products).

share/nisar/defaults/gcov.yaml
```diff
@@ -570,14 +570,27 @@ runconfig:
                 # Default: True
                 qa_reports: true
 
+            software_config:
+
+                # True to use memory map(s) to cache select Dataset(s).
+                # False to always read data directly from the input file.
```
gshiroma, line 584: What does "directly" mean here? I thought memory mapping was applied directly from the files. Or does this process 'flatten' the data first to make it memory-mappable?

gshiroma (commented):
Thank you, @nemo794 , I have a couple of minor comments/questions below.

share/nisar/defaults/gcov.yaml
nemo794: For HDF5 Datasets, we've been using a capital "D".

share/nisar/defaults/gcov.yaml
nemo794: QA adopted the convention of "LUT" as the generic for all metadata look-up tables, which includes metadata cubes. PR comment: https://github.com/isce-framework/nisarqa/pull/9#discussion_r1932748405

share/nisar/defaults/gcov.yaml
nemo794: As a reminder, the full description for this parameter is:

```
                # True to use memory map(s) to cache select Dataset(s).
                # False to always read data directly from the input file.
                # Generally, enabling caching should reduce runtime.
```

This parameter is under the `software_config` group in the runconfig so it is clear that this parameter applies only to the QA software, and the first line of the description clearly states what QA is using the cache for.

As a bonus, by keeping it as a generic `use_cache`, this allows QA the freedom to cache other e.g. intermediary files in the future without a change to the input parameters. We have no plans at this point to do so, but why not stay flexible?

> Also, as far as I know, memory mapping only works when the data is contiguous/linear ("binary"). GCOV imagery is compressed, so is it memory mappable? Does this option require any specific assumptions about the input file in order to be memory mapped?

TBH, a significant portion of the overall QA speedup from using the memory mapping is that we only pay the penalty of de-compressing each Dataset one time. Please feel free to take a look at the implementation: https://github.com/isce-framework/nisarqa/blob/cd38dce99b1a76f7e4688bfa6936665a8294c1e4/src/nisarqa/input_product_readers/_utils.py#L333

share/nisar/defaults/gcov.yaml
nemo794: The QA code doesn't directly memory-map the Datasets within the HDF5 file -- that's not possible. What we're doing is reading the Datasets (block-by-block) and writing them to an intermediate flat file that's memory-mapped. This is useful because it speeds up repeated access to the data by subsequent modules in the QA code (especially irregular access that doesn't respect chunk boundaries).

Please feel free to take a look at the implementation: https://github.com/isce-framework/nisarqa/blob/cd38dce99b1a76f7e4688bfa6936665a8294c1e4/src/nisarqa/input_product_readers/_utils.py#L333

share/nisar/defaults/gcov.yaml
gshiroma: QA may have adopted this convention, but it’s not the terminology we use in this runconfig.(see it [here](https://github.com/isce-framework/isce3/blob/675bde1e32611f856606d11f95d8654b70fe1ac1/share/nisar/defaults/gcov.yaml#L445)). We know metadata cubes are the only "3D LUTs" in the product, but the user doesn't know that.  All I'm asking is to make it clear that this flag applies to metadata cubes (radar grid cubes for L2 products).  Something very simple like: 
```
... z-dimension height layers in a 3D LUT (radar grid cubes) has all non-finite values.
```

share/nisar/defaults/gcov.yaml
gshiroma: Yes, it's clear that this only applies to QA software. My point is that you are not enabling/disabling the [system's native memory caching](https://en.wikipedia.org/wiki/Cache_hierarchy). So, it would be better to rename this field for clarity, e.g., `use_memory_map` or `use_memory_map_for_caching`.

share/nisar/defaults/gcov.yaml
gshiroma: We define that in the product specs document, but here it's outside that context and appears to be a grammatical error.

share/nisar/defaults/gcov.yaml
gshiroma: Thank you, @nemo794 , for the clarification. It would be great to add that explanation to the comments.

share/nisar/defaults/gcov.yaml
nemo794: I'm glad we agree that its clear that this only applies to the QA software -- not the kernel cache.

There's plenty of software that exposes options for configuring its internal file cache without explicitly specifying that it doesn't affect the kernel cache. [GDAL](https://gdal.org/en/stable/api/python/general.html#osgeo.gdal.SetCacheMax), for example.

gshiroma: Yes, it's clear that this only applies to QA software. However, it's not clear that it doesn't apply to the system cache (such as `osgeo.gdal.SetCacheMax()` you pointed out). We should clarify that. Thank you, @nemo794 .

gmgunter: FWIW, I think it's pretty intuitive and clear that this is configuring the *software-managed* file cache -- not the kernel cache. And the description in the comment above is very clear that this is controlling the use of memory-mapping.

Note that changing this parameter name would require an upstream change to the QA software and necessitate and new QA release to be integrated into ISCE3. I don't really think this is worth bikeshedding about.

gshiroma: Thank you, @gmgunter . I think the field name `use_cache` is misleading and one can think of the system's memory cache.  It would've been nice if we have reviewed these runconfig changes in the QA repo before they were merged into the QA code.

gmgunter: >It would've been nice if we have reviewed these runconfig changes in the QA repo before they were merged into the QA code.

That's fair. Sorry about that. AFAIA this is the first time it's been an issue. In future, we should try and get feedback before implementing runconfig changes in QA.

>However, it's not clear that it doesn't apply to the system cache (such as `osgeo.gdal.SetCacheMax()` you pointed out).

I'm not sure if I'm misunderstanding your comment, but it seems like we might disagree about the behavior of `osgeo.gdal.SetCacheMax()`. That function is just controlling the behavior of GDAL's own internal block cache -- not the kernel cache. I don't believe there's any way to control the kernel cache in this way (i.e. to disable it or cap its size for an individual library or process)[^1][^2][^3].

[^1]: https://unix.stackexchange.com/a/590014
[^2]: https://serverfault.com/a/1107488
[^3]: https://unix.stackexchange.com/a/255000

gshiroma: Thank you, @gmgunter ,sounds good. I appreciate you putting us in the loop about these changes to the runconfig.

In terms of `osgeo.gdal.SetCacheMax()`, I thought that function was controlling the memory's higher level cache, something like the "virtual memory". Or maybe I should have made the distinction between the machine cache and the operating system cache (I'm not an expert on that). Anyways, I thought that since the PR has not been merged yet, we had more flexibility to choose a better name than `use_cache`.

gmgunter: >Anyways, I thought that since the PR has not been merged yet, we had more flexibility to choose a better name than `use_cache`.

Yeah, it's kind of clunky that we maintain the QA codebase separately but need to copy the QA runconfig contents into the SAS runconfig files maintained in ISCE3, which necessitates an ISCE3 PR after the changes have already stabilized in the QA code. I think if we had anticipated all the pain points this would cause, we probably would've pushed back harder when PGE requested that we merge the two runconfigs.

It's tricky to change runconfig parameter names now, because that would break the interface with the QA v14.0.0 code. However, personally I'm in favor of adopting some of your suggestions for improving the comments in the YAML file since that won't cause an interface break and the changes could simply be upstreamed in the next QA release.

In the future, hopefully we can avoid this issue by looping you into any changes in the QA codebase that affect the runconfig.

share/nisar/defaults/gcov.yaml
gmgunter: What do you think about using the phrase "metadata cubes" here instead of "radar grid cubes" so that we can copy the same verbiage into all of the runconfig files? IIUC, "radar grid cubes" is specific to the L2 products whereas "metadata cubes" is generic. (I'm not sure what to call the cubes in the L1 products -- "geo grid cubes" I guess?)

share/nisar/defaults/gcov.yaml
gshiroma: Thank you, @gmgunter , for your consideration . I'm resolving this comment.

share/nisar/defaults/gcov.yaml
gshiroma: Sounds good to me. Thank you, @gmgunter .

nemo794:
Thank you, @gshiroma . Comments have been addressed. Could you please take another look? Thanks!

gshiroma (approved):
LGTM. Thank you, @nemo794 !

nemo794:
Thank you, @gshiroma and @gmgunter ! 🚀

nemo794:
FYI -- there was a merge conflict in the Dockerfile with soil moisture, so after the reviews were complete I rebased this branch to resolve the conflict.

PR #114, opened 2025-08-27
Use time-dependent transmit antenna pattern (NOT FOR R5)

bhawkins:
The EAP code was using a scalar time stamp to compute the transmit antenna pattern rather than the vector of time stamps in the current analysis block. That would cause it to miss changes to the beamforming weights over slow time.

### DIFF
--- a/python/packages/nisar/antenna/pattern.py
+++ b/python/packages/nisar/antenna/pattern.py
@@ -400,27 +400,27 @@ class AntennaPattern:
             if tx_pol == "L":
                 tx_bmf_pat[tx_pol] = (
                     self.tx_bmf['H'].form_pattern(
-                        t, slant_range, nearest=nearest,
+                        tgroup, slant_range, nearest=nearest,
                         channel_adj_factors=self.channel_adj_fact_tx['H']) +
                     1j * self.tx_bmf['V'].form_pattern(
-                        t, slant_range, nearest=nearest,
+                        tgroup, slant_range, nearest=nearest,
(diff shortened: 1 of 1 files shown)


Tyler-g-hudson (approved):
Had a look over the modified code and `TxBMF.form_pattern()` - The logic as you've explained it makes sense. LGTM

rad-eng-59 (approved):
Thanks for the fix.

python/packages/nisar/antenna/pattern.py
```diff
@@ -400,27 +400,27 @@ def form_pattern(self, tseq, slant_range: Linspace,
             if tx_pol == "L":
                 tx_bmf_pat[tx_pol] = (
                     self.tx_bmf['H'].form_pattern(
-                        t, slant_range, nearest=nearest,
+                        tgroup, slant_range, nearest=nearest,
```
rad-eng-59: I believe the time should be `tseq` not `tgroup` for all the lines here and below. `tgroup` is a portion of `tseq`. 
```suggestion
                        tseq, slant_range, nearest=nearest,
```

rad-eng-59 (changes requested):

python/packages/nisar/antenna/pattern.py
bhawkins: Agreed. Thanks for catching that!

rad-eng-59 (approved):
LGTM. Thank you for the fix.

bhawkins:
Merging despite MacOS failure #131

PR #118, opened 2025-08-27
Limit EAP dynamic range (NOT FOR R5)

bhawkins:
This PR includes the option from @rad-eng-59 to limit the dynamic range of the two-way antenna pattern to avoid excessive amplification of noise and the possibility of inf/nan values in the imagery and/or noise estimates. It extends that patch with the ability to control the dynamic range with a parameter in the RSLC configuration file.

### DIFF
--- a/share/nisar/defaults/focus.yaml
+++ b/share/nisar/defaults/focus.yaml
@@ -399,6 +399,13 @@ runconfig:
                 range: 12.0
                 azimuth: 12.0
 
+            # Maximum allowed dynamic range of the elevation antenna pattern
+            # correction.  This is useful when the swath contains regions with
+            # very small antenna gain and one wants to avoid unduly amplifying
+            # the noise in these regions.  Disabled when set to null (the
+            # default).
+            eap_dynamic_range_limit_db: null
+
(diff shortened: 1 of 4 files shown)


share/nisar/defaults/focus.yaml
```diff
@@ -399,6 +399,13 @@ runconfig:
                 range: 12.0
                 azimuth: 12.0
 
+            # Maximum allowed dynamic range of the elevation antenna pattern
```
rad-eng-59:
```suggestion
            # Maximum allowed dynamic range of the two-way elevation antenna pattern
```

share/nisar/schemas/focus.yaml
```diff
@@ -238,6 +238,12 @@ runconfig:
         range: num(min=0.0, required=False)
         azimuth: num(min=0.0, required=False)
 
+      # Maximum allowed dynamic range of the elevation antenna pattern
```
rad-eng-59:
```suggestion
      # Maximum allowed dynamic range of the two-way elevation antenna pattern
```

rad-eng-59 (approved):
LGTM. Thanks for the addition.

PR #119, opened 2025-08-28
Refactor Topo multithreading to reduce synchronization overhead (MCR #90132)

gmgunter:
This PR refactors the multi-threading strategy used by `Topo::topo()` to reduce latencies due to thread synchronization.

The previous implementation was designed to iterate over azimuth lines in serial and only parallelized over range samples within each azimuth line. I think this approach might've been chosen because there are a few values that only need to be computed once per azimuth line (e.g. the platform speed) and then can be reused by each thread.

This parallelization strategy introduces lots of latency due to thread synchronization overhead, though. Profiling the Topo runtime using `py-spy record --native` (a sampling profiler) shows that threads spend as much as 30% of the overall runtime in a function called `do_wait`.

This update refactors the `Topo::topo()` method to parallelize over lines instead.

I've also modified the implementation such that each call to `rdr2geo` uses the previous result of `rdr2geo` as its initial guess, which should typically reduce the number of iterations required for `rdr2geo` to converge. Previously, the implementation used the center point of the DEM interpolator as the initial guess for each call to `rdr2geo`.

I removed the "Topo progress" messages that printed the percent of completed azimuth lines within the current block, since they interfered with the new parallelization strategy (and were probably unnecessarily frequent anyway, IMO). There is still a status message indicating when each Topo block completes processing.

These updates significantly improve the overall runtime of the Topo module. On nisar-adt-rtc-1 (an r7i.4xlarge instance), the Topo runtime for generating Static Layers sample products at 20m x 20m posting decreased by ~38% (N=3). On nisar-adt-dev-3 (a g4dn.48xlarge instance, the runtime decreased by ~43% (N=3).

### DIFF
--- a/cxx/isce3/geometry/Topo.cpp
+++ b/cxx/isce3/geometry/Topo.cpp
@@ -176,62 +176,73 @@ topo(Raster & demRaster, TopoLayers & layers)
         // Reset output block sizes in layers
         layers.setBlockSize(blockLength, _radarGrid.width());
 
-        // Allocate vector for storing satellite position for each line
-        std::vector<Vec3> satPosition(blockLength);
+        // Pre-compute some values that only need to be computed once per
+        // range line.
+        auto sensingTime = std::vector<double>(blockLength);
+        auto satPosition = std::vector<Vec3>(blockLength);
+        auto satVelocity = std::vector<Vec3>(blockLength);
+        auto satSpeed = std::vector<double>(blockLength);
+        auto TCNbases = std::vector<Basis>(blockLength);
 
-        // For each line in block
-        double tline;
+        #pragma omp parallel for
         for (size_t blockLine = 0; blockLine < blockLength; ++blockLine) {
-
-            if (blockLine % std::max((int) (blockLength / 100), 1) == 0)
-                printf("\rTopo progress (block %d/%d): %d%%",
-                       (int) block + 1, (int) nBlocks,
-                       (int) (blockLine * 1e2 / blockLength)),
-                       fflush(stdout);
-
             // Global line index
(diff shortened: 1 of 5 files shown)


gmgunter:
@bhawkins @gshiroma this is a copy of a PR that you approved on the internal GitHub server: https://github-fn.jpl.nasa.gov/isce-3/isce/pull/2225

gshiroma (approved):
Looks great! 30 to 43% runtime improvement is huge! Thank you, @gunter , for working on this and for addressing all my comments.  PS. We reviewed this PR internally.

cxx/isce3/geometry/Topo.cpp
```diff
@@ -179,59 +180,76 @@ topo(Raster & demRaster, TopoLayers & layers)
...
-                    _radarGrid.lookSide(), _threshold, _numiter, _extraiter);
-                totalconv += geostat;
-
-                // Save data in output arrays
-                _setOutputTopoLayers(llh, layers, blockLine, pixel, pos, vel,
-                        TCNbasis, demInterp);
+        // Get the midpoint of the DEM block in LLH coordinates.
+        // This should always be safe since the call to `computeDEMBounds()`
+        // above loads a block of the DEM raster.
+        assert(demInterp.hasRaster());
```
gshiroma:
```suggestion
        assert(demInterp.haveRaster());
```
Interesting. Looks like this bug hasn't been caught by our internal CI.

gmgunter: @rtburns-jpl those Debug build CI jobs are paying off! :wink:

bhawkins (approved):

PR #120, opened 2025-08-28
Update `dumpconfig.py` to generate 40 MHz GCOV products with RTC DEM upsampling set to `1`

gshiroma:
This PR updates `dumpconfig.py` to generate 40 MHz GCOV products with RTC DEM upsampling set to `1`. This PR mirrors the production rule specified by GCOV parameter set files.


hfattahi (approved):
LGTM. This PR does not impact Ops and does not need MCR. Can be included in R5.

gmgunter:
FYI I've been using `dem_upsampling=2` in the Static Layers runconfig and haven't provided any guidance to the On-Demand team to change this value when processing finer grid postings.

This reminds me I should probably exercise the workflow with finer grid postings to make sure it doesn't run out-of-memory and finishes processing in a reasonable amount of time. So far, the smallest pixel spacing that I've tested with is 10m x 5m, but I think @hfattahi is interested in running with finer postings now.

gshiroma:
> LGTM. This PR does not impact Ops and does not need MCR. Can be included in R5.

You are faster than me, @hfattahi . I was messaging you on slack and you had already approved this PR. Thanks!

python/packages/nisar/workflows/dumpconfig.py
```diff
@@ -610,7 +610,22 @@ def dumpconfig_gslc_gcov(
...
+        first_frequency_bandwidth = int(np.round(
+            rslc.getSwathMetadata().processed_range_bandwidth / 1e6))
+
+        # By default GCOV products are generated with RTC DEM upsampling
+        # set to `2`. However, 40 MHz products require much more memory
+        # and therefore are processed with the RTC DEM upsampling set
+        # to `1`
+        if first_frequency_bandwidth == 40:
+            rtc_group = groups["processing"]["rtc"]
+            rtc_group["dem_upsampling"] = 1
```
Tyler-g-hudson: This is an automatic behavior - do we want to include a manual override in the CLI parameters to permit custom behaviors? All the other parameters in the parameter set files can be overridden

gshiroma:
> FYI I've been using `dem_upsampling=2` in the Static Layers runconfig and haven't provided any guidance to the On-Demand team to change this value when processing finer grid postings.
> 
> This reminds me I should probably exercise the workflow with finer grid postings to make sure it doesn't run out-of-memory and finishes processing in a reasonable amount of time. So far, the smallest pixel spacing that I've tested with is 10m x 5m, but I think @hfattahi is interested in running with finer postings now.

That's a good point @gmgunter . Since RTC is computed in slant-range, it may not change much with the geogrid spacing. Ideally, we'd use the RTC DEM upsampling = 2.

python/packages/nisar/workflows/dumpconfig.py
gshiroma: Hi @Tyler-g-hudson , sure we can. I’d like to avoid making too many changes to this script and delaying the review/merge. If you think this is worth adding, how about we handle this in a separate PR?

gshiroma: @Tyler-g-hudson , we talked about this offline. I just updated the feature in commit [af1cac0](https://github.com/isce-framework/isce3/pull/120/commits/af1cac06be4ba5b3c146f32ef191ad5927fa679f) with your suggestions. Please, let me know if that looks okay.

python/packages/nisar/workflows/dumpconfig.py
```diff
@@ -283,14 +283,32 @@ def init_argparse():
...
+        ),
+        None,
+    )
+    geocoding_group.add_argument(
+        "--rtc-dem-upsampling",
+        dest="rtc_dem_upsampling",
+        type=int,
+        help=(
+            "RTC DEM upsampling."
+        ),
```
Tyler-g-hudson: This looks pretty good! Just some small final documentation improvements:
```suggestion
        required=False,
        help=(
             "Sets the RTC DEM upsampling value. Setting this value overrides the "
             "automatic RTC setting behavior of the --nisar-defaults parameter."
        ),
```

And in the `--nisar-defaults` parameter help string:

```python
    geometry_group.add_argument(
        "--nisar-defaults",
        action="store_true",
        required=False,
        help=(
            "If given, all geospatial parameters (spacing and snap values) not "
            "provided by the user will be filled with default values in meters that "
            "are suited to the workflow and bandwidth of the frequencies in the RSLC "
            "product. Only valid for products with 5, 10, 20, or 77 MHz bandwidths for "
            "all frequencies. This is suggested for use in conjunction with the "
            "--generate-epsg parameter. For GCOV products, this argument may also "
            "modify the RTC value to the NISAR default value for their band."
        ),
    )
```

Tyler-g-hudson (commented):
Looks good! Some final documentation stuff before approval

python/packages/nisar/workflows/dumpconfig.py
gshiroma: Thank you, @Tyler-g-hudson ! These descriptions were updated in commits https://github.com/isce-framework/isce3/pull/120/commits/c6f5c21708703cb17bef01b10a99de49cb7af781 and https://github.com/isce-framework/isce3/pull/120/commits/102481bf6e8c54f1200f178f78cf02ec57671774.

Tyler-g-hudson (approved):
LGTM!

gshiroma:
> LGTM!

Thank you, @Tyler-g-hudson !

PR #122, opened 2025-08-28
 Add subswath mask for ionosphere

oberonia78:
This PR adds the capability to use the subswath mask for ionosphere estimation.
To mitigate low-coherence areas caused by partially focused regions, the subswath mask is now applied during ionosphere estimation.

Specifically, this PR:

- Introduces functions to read and interpret the subswath mask.
- Updates `ionosphere.py` to incorporate the subswath mask in the ionosphere filter, reducing unwrapping errors in noisy areas.


python/packages/nisar/workflows/ionosphere.py
```diff
@@ -1506,6 +1534,23 @@ def run(cfg: dict, runw_hdf5: str):
...
+                        src_main_h5[subswath_mask_freq_a_path].read_direct(
+                            subswath_mask_main_image,
+                            np.s_[row_start:row_start + block_rows_data, :])
+                        # Subswath mask may not be available when frequency B
+                        # was not requested in the runconfig for interferogram
+                        # In this case, we decimate subswath_mask in frequencyA
+                        if subswath_mask_freq_b_path in src_side_h5:
+                            src_side_h5[subswath_mask_freq_b_path].read_direct(
+                                subswath_mask_side_image,
+                                np.s_[row_start:row_start + block_rows_data, :])
```
hfattahi, line 1503: How does this logic work for 77 MHz data without freq B?

python/packages/nisar/workflows/ionosphere.py
```diff
@@ -1303,6 +1307,12 @@ def run(cfg: dict, runw_hdf5: str):
             side_conn_image = None
             diff_ms_conn_image = None
 
+            # ionosphere method using sub-bands uses subswath mask in freq A
+            # ionosphere method using side-band uses subswath masks in freq A and B
+            subswath_mask_image = None
+            subswath_mask_main_image = None
```
hfattahi, line 1269: what is the difference between subswath_mask_image and subswath_mask_main_image?

python/packages/isce3/atmosphere/split_band_estimation.py
```diff
@@ -314,6 +315,69 @@ def get_mask_array(
...
+            if low_band_array is not None:
+                low_band_array = decimate_freq_a_array(
+                    slant_main,
+                    slant_side,
+                    low_band_array)
+            if high_band_array is not None:
+                high_band_array = decimate_freq_a_array(
+                    slant_main,
+                    slant_side,
+                    high_band_array)
```
hfattahi, line 366: These to if statements for low and high bands are within the if statement for side_array above. I don't understand how that is possible that the side array and low and high band exists all at the same time!

hfattahi (commented):
looks good overall. Few comments just to make sure things are clear to me!

python/packages/isce3/atmosphere/main_band_estimation.py
```diff
@@ -366,6 +367,55 @@ def get_valid_area(
...
+        mask_array : numpy.ndarray
+            2D mask array extracted from coherence or
+            connected components
+            1: valid pixels,
+            0: invalid pixels.
+        """
+        # decimate subswath mask
+        # when side array is also used.
+        if side_array is not None:
+            main_array = decimate_freq_a_array(
```
seongsujeong, line 404: I have not taken a look at this function `decimate_freq_an_array` but its name seems to tell that it extracts sub swath mask from frequency A.
My question is: Can we always assume that we have frequency A?

seongsujeong (commented):
Looks good overall. I have a minor question as below:

python/packages/isce3/atmosphere/main_band_estimation.py
oberonia78: If we don't have frequency A interferogram, it sound like we cannot produce the ionosphere and interferograms. So, `ionosphere_runconfig.py` or `insar_runconfig.py` will raise the error before running this function.

seongsujeong: Thanks for the explanation. I take your explanation as like the module preceding this code will raise error before the process reaches this point.

seongsujeong (approved):
LGTM

python/packages/nisar/workflows/ionosphere.py
oberonia78: As you see in line 1425 (previously1405), this subswath_mask is for `main_diff_ms_band` method. when we don't have frequency B, we extract it from main band only (line 1345).

python/packages/nisar/workflows/ionosphere.py
oberonia78: To avoid confusion, I made two separate variables. subswath_mask_image for 77MHz, and subswaht_mask_main_image is for 40 and 20 MHz. I thought two separate variables are better in reading code.

python/packages/isce3/atmosphere/split_band_estimation.py
oberonia78: This was designed some time ago and has been preserved intentionally. The same high-level interface is implemented in both `split_band_estimation.py` (77 MHz) and `main_band_estimation.py` (20 / 40 MHz).

When the ionosphere object is created, `ionosphere.py` selects the appropriate estimation method based on the user-specified one, and then calls `get_subswath_mask_array` through a unified interface (see line 1704 in ionosphere.py).

To support all estimation modes consistently, `get_subswath_mask_array` in both implementations is required to accept the same input arguments. For this reason, `main_array` and `side_array` are defined with default values of None. In the 77 MHz split-band case, these arguments are simply unused and do not need to be provided, whereas they are required for the 20 / 40 MHz main- and side-band methods.

xhuang-jpl (approved):
Thank you @oberonia78 . LGTM.

PR #125, opened 2025-08-29
Allow downloading DEM with polar stereo coordinates

hfattahi:
This PR modifies stage_dem to allow for downloading DEMs from our polar stereo DEMs with a bounding box in polar stereo coordinates. 
I have added two epsg input parameters : 1) --bbox-epsg which represents the epsg of the input bounding box 2) --dem_epsg which corresponds to the epsg of the output DEM. With these two options and with the changes in this PR, user may expect the following behaviors which to a very good extent maintains the previous behavior of stage_dem and adds new functionality:

1. extract dem with a box in lat/lon (epsg 4326): User needs to provide the bbox (epsg is 4326 by default and dem_epsg is equal to polygon espg by default)
2. extract dem with a box in polar coordinates by providing the box in polar coordinates and corresponding --epsg. No need to provide --dem_epsg as it defaults to --epsg
3. If user asks for a DEM in polar coordinate and our DEM does not cover that latitude, stage_dem, falls back to lat/lon DEM and provides a DEM in lat/lon
4. If user wants the DEM in polar stereo but does not know the box in polar stereo, they may provide the bbox in lat/lon with bbox-epsg 4326 and specify the DEM epsg using --dem_epsg . (In this case stage_dem converts the lat/lon to dem_epsg before extracting the DEM)

### DIFF
--- a/python/packages/nisar/workflows/stage_dem.py
+++ b/python/packages/nisar/workflows/stage_dem.py
@@ -41,9 +41,26 @@ def cmdLineParse():
                         help='Filepath to user DEM.')
     parser.add_argument('-m', '--margin', type=int, action='store',
                         default=5, help='Margin for DEM bounding box (km)')
-    parser.add_argument('-b', '--bbox', type=float, action='store',
-                        dest='bbox', default=None, nargs='+',
-                        help='Spatial bounding box in latitude/longitude (WSEN, decimal degrees)')
+    parser.add_argument('-b', '--bbox', type=float, nargs=4,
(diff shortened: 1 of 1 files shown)


seongsujeong (approved):
LGTM. I've approved the corresponding PR in the internal repository.

gshiroma (approved):
Looks great! This PR brings significant improvements over the current version. We reviewed this PR internally. Thank you, @hfattahi , for your patience in addressing all my comments.

gmgunter (approved):
LGTM, thank you!

PR #126, opened 2025-08-29
Set stage_dem --bbox nargs to 4

Tyler-g-hudson:
Currently, stage_dem will quietly accept if a number other than 4 values (i.e. a malformed WSEN lat/lon) are given to the `--bbox` parameter. This PR locks this parameter to only accept exactly 4 values when given.


gmgunter:
Nice catch! But @hfattahi beat you to the punch in #125 :wink:

Hopefully that PR will get merged today for R5 with the same fix

Tyler-g-hudson:
Closing since this fix is in #125 already

PR #132, opened 2025-09-04
Include beta0 area normalization in focusing

bhawkins:
This PR includes the beta0 area normalization term in the image scaling during focusing so that it doesn't have to be bookkept as a separate calibration parameter elsewhere. Needs testing, so marking as draft.

### DIFF
--- a/python/packages/nisar/workflows/focus.py
+++ b/python/packages/nisar/workflows/focus.py
@@ -30,7 +30,8 @@ from isce3.core import DateTime, TimeDelta, LUT2d, Attitude, Orbit
 from isce3.focus import make_los_luts, fill_gaps, make_cal_luts, Notch
 from isce3.geometry import los2doppler
 from isce3.io.gdal import Raster, GDT_CFloat32
-from isce3.product import RadarGridParameters
+from isce3.product import (RadarGridParameters,
+    get_radar_grid_nominal_ground_spacing)
 from nisar.workflows.yaml_argparse import YamlArgparse
 import nisar.workflows.helpers as helpers
 from ruamel.yaml import YAML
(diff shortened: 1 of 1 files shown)


python/packages/nisar/workflows/focus.py
```diff
@@ -1622,6 +1629,9 @@ def focus(runconfig, runconfig_path=""):
         dop[frequency] = scale_doppler(dop_ref, band.center / fc_ref)
         blocks_bounds[frequency] = plan_processing_blocks(cfg, ogrid[frequency],
                                         dop[frequency], dem, orbit)
+        # So does output pixel area (beta0 convention).
+        daz, _ = get_radar_grid_nominal_ground_spacing(ogrid[frequency], orbit)
```
gshiroma, line 1633: It looks like you are using a single `daz` value for each frequency. However, the variation in ground velocity across the radar grid may not be negligible. As you know, the ground velocity depends on the topography, but also on the off-nadir angle, which may be significant over a wide swath.

gshiroma (commented):
Quick comment below for your consideration.

python/packages/nisar/workflows/focus.py
bhawkins: That's possible, but for now I'd prefer to keep it simple. I debated including an azimuth term at all since the s/c velocity is roughly constant and it could be absorbed in our abscal factor (which can have a slope term). However, I ended up including a simple one in case we produce images at different PRFs. My main goal for now is to make the processor gain (measured from beta0) the same across modes/subbands, which really only requires accounting for the range pixel size. I'll get some results with this config and evaluate.

gshiroma: Sounds good to me, @bhawkins , I'll create an issue before approving this PR to remind us to update it later.

gshiroma: @bhawkins , I created the issue #171 so we don't forget about this approximation.

gshiroma (approved):
This PR uses an approximation for the ground velocity that should be improved in the future. This is documented in issue #171. For now, the changes look good.

hfattahi (approved):
LGTM!

Tyler-g-hudson (approved):
LGTM!

Issue #130, opened 2025-09-03
InSAR test failures when calculating diff_phase in ionosphere.py/compute_differential_phase


jshimada47:
Several of the InSAR benchmarking test cases failed with variants of the following error when run with the r05.00.0-v0.25.0 Docker image last night:
```
 - journal: successfully ran crossmul in 84.871 seconds
 - Traceback (most recent call last):
 -   File "<frozen runpy>", line 198, in _run_module_as_main
 -   File "<frozen runpy>", line 88, in _run_code
 -   File "/opt/conda/lib/python3.12/site-packages/nisar/workflows/insar.py", line 130, in <module>
 -     run(insar_runcfg.cfg, out_paths, persist.run_steps)
 -   File "/opt/conda/lib/python3.12/site-packages/nisar/workflows/insar.py", line 85, in run
 -     ionosphere.run(cfg, out_paths['RUNW'])
 -   File "/opt/conda/lib/python3.12/site-packages/nisar/workflows/ionosphere.py", line 1060, in run
 -     insar_ionosphere_pair(iono_insar_cfg, runw_hdf5)
 -   File "/opt/conda/lib/python3.12/site-packages/nisar/workflows/ionosphere.py", line 883, in insar_ionosphere_pair
 -     compute_differential_phase(phase_first,
 -   File "/opt/conda/lib/python3.12/site-packages/nisar/workflows/ionosphere.py", line 538, in compute_differential_phase
 -     diff_phase = first_data_block * np.conj(second_data_block)
 -                  ~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~
 - ValueError: operands could not be broadcast together with shapes (1000,961) (1000,927)

```
Before running the test cases, I'd updated their runconfigs to conform to the version 00408 Parameter Set Files.  The runconfigs had not been updated recently so some of the changes I made may date back to release 4.0.7 or earlier.

The above error was from the 20+5 mountain_glaciers interferograms (RIFG, RUNW, GUNW) test case.  The full set of test cases that failed with this error are

20+5 mountain glaciers, interferograms (RIFG, RUNW, GUNW)
20+5 mountain glaciers, all 5 products
40+5 ice sheets, interferograms
40+5 ice sheets, all 5 products
40+5 mountain glaciers, interferograms
40+5 mountain glaciers, all 5 products

The array shapes varied from case to case, but the errors were the same.

[...]

PR #133, opened 2025-09-04
Fix issue in ionosphere method (main_diff_ms_band)

oberonia78:
**Problem**

1main_diff_ms_band1 builds a differential interferogram by combining a frequency-A interferogram with a frequency-B interferogram. The current code reads:

- freq A from RIFG
- freq B from RUNW

In operations, A and B are produced with different multilook factors, yielding different pixel spacings. Because freq-B spacing is not an integer multiple of freq-A spacing, the decimation step on the freq-A interferogram fails.

**Solution**

Read freq A from RUNW instead of RIFG. The RUNW products for freq A and freq B are generated with the same multilook and pixel spacing, eliminating the non-integer decimation issue. Since RUNW stores unwrapped phase, we re-wrap it to a complex phasor before forming the differential interferogram:

### DIFF
--- a/python/packages/isce3/signal/interpolate_by_range.py
+++ b/python/packages/isce3/signal/interpolate_by_range.py
@@ -30,13 +30,21 @@ def decimate_freq_a_array(
 
     resampling_scale_factor = int(np.round(spacing_side / spacing_main))
 
-    x_cand = np.arange(1, width + 1)
-
-    # find the maximum of the multiple of resampling_scale_factor
-    decimate_width_end = np.max(x_cand[x_cand % resampling_scale_factor == 0])
+    # ensure the decimated width matches len(slant_side)
+    n_side = len(slant_side)
+    decimate_width_end = first_index + n_side * resampling_scale_factor
+    if decimate_width_end > width:
+        # shift start left so it fits; clamp to 0
+        first_index = max(0, width - n_side * resampling_scale_factor)
+        decimate_width_end = first_index + n_side * resampling_scale_factor
     decimated_array = target_runw[
         :, first_index:decimate_width_end:resampling_scale_factor]
 
+    if decimated_array.shape[1] < n_side:
+        pad_cols = n_side - decimated_array.shape[1]
+        pad = np.zeros((decimated_array.shape[0], pad_cols),
+                       dtype=decimated_array.dtype)
+        decimated_array = np.concatenate([decimated_array, pad], axis=1)
     return decimated_array
 
 def interpolate_freq_b_array(
--- a/python/packages/isce3/signal/interpolate_by_range.py
+++ b/python/packages/isce3/signal/interpolate_by_range.py
@@ -30,21 +30,13 @@ def decimate_freq_a_array(
 
     resampling_scale_factor = int(np.round(spacing_side / spacing_main))
 
-    # ensure the decimated width matches len(slant_side)
-    n_side = len(slant_side)
-    decimate_width_end = first_index + n_side * resampling_scale_factor
-    if decimate_width_end > width:
-        # shift start left so it fits; clamp to 0
-        first_index = max(0, width - n_side * resampling_scale_factor)
-        decimate_width_end = first_index + n_side * resampling_scale_factor
+    x_cand = np.arange(1, width + 1)
+
+    # find the maximum of the multiple of resampling_scale_factor
+    decimate_width_end = np.max(x_cand[x_cand % resampling_scale_factor == 0])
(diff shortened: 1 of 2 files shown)


xhuang-jpl (approved):
Thanks @oberonia78 I have tested this PR using the 40Mhz and 80Mhz data, which works fine without issues. Please address the CI failures.

hfattahi:
To be clear, the shape of frequency A RSLC is indeed a multiple integer of frequency B RSLC. The issue here is not with the dimensions of RSLC but with the dimensions of wrapped interferograms which can get to an inconsistency when RIFG products are made with a different posting (e.g. 30 m for cryosphere regions), than the RUNW posting (e.g., 80 m). Based on the configurations for GUNW production, this could be potentially an issue over Cryo regions. @oberonia78 would you please confirm if you have tested this PR with Cryo configurations (producing all 5 products) and with different modes?

python/packages/nisar/workflows/ionosphere.py
```diff
@@ -461,6 +461,16 @@ def compute_differential_phase(phase_first,
...
     """
+    def _to_complex_if_needed(arr):
+        """
+        Ensure an array is complex. If it's already complex, return as-is.
+        Otherwise interpret values as phase (radians) and convert with exp(j*phi).
+        """
+        if np.iscomplexobj(arr):
+            return arr
+        # Cast to float to avoid integer exponentiation issues
+        return np.exp(1j * arr.astype(np.float32, copy=False))
```
hfattahi:
```suggestion
        return np.exp(1j * arr)
```

hfattahi (commented):
One minor comment. @vbrancat is also running some tests. Once I have confirmation that all tests are running, we can approve this.

vbrancat (approved):
LGTM, @jungkyoJung, all my tests passed

python/packages/isce3/signal/interpolate_by_range.py
```diff
@@ -30,13 +30,21 @@ def decimate_freq_a_array(
...
+        first_index = max(0, width - n_side * resampling_scale_factor)
+        decimate_width_end = first_index + n_side * resampling_scale_factor
     decimated_array = target_runw[
         :, first_index:decimate_width_end:resampling_scale_factor]
 
+    if decimated_array.shape[1] < n_side:
+        pad_cols = n_side - decimated_array.shape[1]
+        pad = np.zeros((decimated_array.shape[0], pad_cols),
+                       dtype=decimated_array.dtype)
+        decimated_array = np.concatenate([decimated_array, pad], axis=1)
```
hfattahi: I don't understand in what situation this would be needed. Perhaps safe to remove?

python/packages/isce3/signal/interpolate_by_range.py
```diff
@@ -30,13 +30,21 @@ def decimate_freq_a_array(
...
-
-    # find the maximum of the multiple of resampling_scale_factor
-    decimate_width_end = np.max(x_cand[x_cand % resampling_scale_factor == 0])
+    # ensure the decimated width matches len(slant_side)
+    n_side = len(slant_side)
+    decimate_width_end = first_index + n_side * resampling_scale_factor
+    if decimate_width_end > width:
+        # shift start left so it fits; clamp to 0
+        first_index = max(0, width - n_side * resampling_scale_factor)
+        decimate_width_end = first_index + n_side * resampling_scale_factor
```
hfattahi: I am concerned about this logic. It looks like this check was put in for a case where frequency A and B may not have same starting range. 1) NISAR RSLC by spec will have same starting range for A and B. 2) even if some test data would have A and B with different starting range, the logic here would lead to decimated image not being on the same grid. So I suggest to revert this change please.

oberonia78:
This PR has been tested with 
20+5 mountain glaciers, interferograms (RIFG, RUNW, GUNW)
40+5 ice sheets, interferograms
40+5 mountain glaciers, interferograms
40+5 mountain glaciers, all 5 products
in issue (https://github.com/isce-framework/isce3/issues/130) The above tests passed.

hfattahi (approved):
LGTM. Thanks. Heads up to @vbrancat and @xhuang-jpl that there has been some activities since they last reviewed. If still fine with them, then we can merge this PR.

python/packages/isce3/signal/interpolate_by_range.py
oberonia78: This logic was deleted for now. We may need to revisit this issue once NISAR RSLC's main and side bands are fixed.

PR #134, opened 2025-09-06
Make a Robust Check for RCID in Noise Est

rad-eng-59:
Quick PR to make a more robust RCID check in noise estimator workflow for noise-only product where the last value of RCID array can be invalid similar to the first one surprisingly!
This is actually not a bug in the code but rather needed due to less reliable telemetry carrying invalid RCID values for a certain observation such as noise-only case.


python/packages/nisar/noise/noise_estimation_from_raw.py
```diff
@@ -218,9 +219,12 @@ def extract_noise_only_lines(raw, freq_band, txrx_pol, max_lines=18944):
...
             else:
-                # XXX Use last value to check for RCID given the delay
-                # in updating low-rate telemetry (DRT) for NISAR!
-                if rcids[-1] in rcid_special:
+                # XXX Use mid, last, most counted value to check for RCID
+                # given delay in updating low-rate telemetry (DRT) for NISAR!
+                rcid_mode = stats.mode(rcids).mode
+                if (rcids[-1] in rcid_special or
+                    rcids[len(rcids) // 2] in rcid_special or
+                        rcid_mode in rcid_special):
```
bhawkins: After talking this through with Alice, we're just scratching the surface of the complexity here. The basic problem is that the low-rate telemetry is not synchronized with the echo data. The L0B file contains all the telemetry that arrived in the L0A data stream during the observation, so any given entry may actually correspond to the previous, current, or next observation. (I've actually seen telemetry from 2 minutes in the future show up!)

When the L0B processor needs to use the telemetry to make a decision, it applies a logical filter with eight different conditions on timing and other values to select the "correct" ones. Even then, apparently the RCID is not always chosen correctly. The L0B processor therefore does not use the telemetry to determine the RCID and instead gets it from the processing system which determines it from the observation plan. 

The most reliable way to get the RCID is thus by parsing the file name or the `granuleId` dataset in the identification group. I'm not a huge fan of relying on file names, but the `granuleId` might not be too bad. Another option would be to punt and make this a command-line option.

If you really want to attempt to use the telemetry, I'd suggest the following changes:
```suggestion
                valid_rcids = rcids[rcids > 0]
                rcid_mode = stats.mode(valid_rcids).mode
                if (valid_rcids[len(valid_rcids) // 2] in rcid_special or
                        rcid_mode in rcid_special):
```
I think the check with the final index invites false positives when an observation occurs before an rx-only observation. The rcid > 0 check appears in the L0B code to filter out missing/invalid values. The midpoint and mode checks seem kind of redundant to me and make the failure scenarios more complicated, so I'd recommend picking one or the other to keep it simple, but I'm not adamant about that.

bhawkins (commented):
This is a good idea but not a full solution. I have some suggestions and think at least a one-line change would be prudent.

python/packages/nisar/noise/noise_estimation_from_raw.py
rad-eng-59: I don't think RCID can be negative given they are unsigned integer starting from `1` (min RCID). However,  I'll apply the recommendation as it is. Thanks.

rad-eng-59: Corrections:
- First of all the values are always `uint8` for RCIDs so there is no way we get negative value. 
- This logic of yours is not working either. I guess it's safe to simply check for noise-only RCID inside DRT RCIDs given the code first checks if there is any noise-only range lines (sniffer pulses) and if not then it checks for RCIDs. The only trivial scenario that it can go wrong (mistakenly treated as noise-only instead of science) is when the observation is too short ( `<< 2 seconds`) where there is no noise-only range lines (no LNA/BYPS Cal Lines). In that case, I don't think it matters anyway. 

Here is an example of DRT of `RCID=1` that causes the failure in the workflow:
```

(Pdb) rcids
array([191, 191,   1,   1,   1,   1,   1,   1,   1,   1,   1,   1,   1,
         1,   1,   1,   1,   1,   1,   1,   1,   1,   1,   1,   1,   1,
         1,   1,   1,   1,   1,   1,   1,   1,   1,   1,   1,   1,   1,
         1,   1,   1,   1,   1,   1,   1,   1,   1,   1,   1,   1,   1,
         1,   1,   1,   1,   1,   1,   1,   1,   1,   1,   1,   1,   1,
         1,   1,   1,   1,   1,   1,   1,   1,   1,   1,   1,   1,   1,
         1,   1,   1,   1,   1,   1,   1,   1,   1,   1,   1,   1,   1,
         1,   1,   1,   1,   1,   1,   1,   1,   1,   1,   1,   1,   1,
         1,   1,   1,   1,   1,   1,   1,   1,   1,   1,   1,   1,   1,
       190, 190, 255, 255, 134, 134, 134, 134, 134, 134, 134, 134, 134,
       134, 134, 134, 134, 134, 134, 134, 134, 134, 134, 134, 134, 134,
       134, 134, 134, 134, 134, 134, 134, 134, 134, 134, 134, 134, 134,
       134, 134, 134, 134, 134, 134, 134, 134, 134, 134, 134, 134, 134,
       134, 134, 134, 134, 134, 134, 134, 134, 134, 134, 134, 134, 134,
       134, 134, 134, 134, 134, 134, 134, 134, 134, 134, 134, 134, 134,
       134, 134, 134, 134, 134, 134, 134, 134, 134, 134, 134, 134, 134,
       134, 134, 134, 134, 134, 134, 134, 134, 134, 134, 134, 134, 134,
       134, 134, 134, 134, 134, 134, 134, 134, 134, 134, 134, 134, 134,
       134, 134, 134, 134, 134, 134, 134, 254, 254, 255, 255, 137, 137,
       137, 137, 137, 137, 137], dtype=uint8)
```

Tyler-g-hudson: Brian has said here that the best estimate for RCID is to simply check the granuleId dataset, that makes sense to me; what is the reason for doing this instead of that?

rad-eng-59: Because that field doesn't exist or invalid in old product and simulation products. Besides, last time I check we do not rely on filename anywhere to extract info. 
However, I can extract RCID from granule ID if exists  but in case the code is used to exercise any simulation or old lab data for testing purposes it may lead to unexpected outcome.

rad-eng-59: FYI, our `Raw` product reader currently do not parse `granuleID` which is another proof that it is relatively new compared to when the original code is written.

Tyler-g-hudson: Well, it's a bit troubling when our product readers are unable to read our products, at least not completely...
I think it's best if we use the best available info if it's indeed available, then we can fall back to more approximate 
means like this if we have to

rad-eng-59:
CCed @gshiroma for that bug fixes regarding noise product in RSLC workflow. @gshiroma please confirm if this one-line fix in `focus.py`avoid that failure you have reported earlier today.  Thank you so much

rad-eng-59:
FYI, this failure on Mac OS has nothing to do with this PR.

Tyler-g-hudson (approved):
Approved pending review by @gshiroma

gshiroma (approved):
LGTM. I can confirm that this PR fixes the `sr_noise` not declared issue I encountered in my previous tests. Thank you, @rad-eng-59 !

PR #135, opened 2025-09-08
AZ Blocked DC Removal in Doppler Est Workflow 

rad-eng-59:
To mitigate large internal signal such as loud Caltone which can bias Doppler centroid estimation, the option for DC removal in along track is introduced in Doppler estimator workflow. 
The DC removal in AZ is done in blocking and the duration of the block is set by user in seconds regardless of PRF.
Optionally, it can be done by windowed/smoothed notch filtering in AZ spectrum by using inverse raised -cosine with proper width/size as shown in [here](https://github.jpl.nasa.gov/NISAR-L0B-STUFF/l0b_with_mitigated_tone).

### DIFF
--- a/python/packages/nisar/pointing/doppler_lut_from_raw.py
+++ b/python/packages/nisar/pointing/doppler_lut_from_raw.py
@@ -114,9 +114,11 @@ def doppler_lut_from_raw(raw, *, freq_band='A', txrx_pol=None,
         of beams or RX channels. By default, all beams are included
         in the polyfitting of Doppler over entire swath of DM2 products.
     duration_dc_remove_az: float or None, default=0.5
-        Max AZ-block duration in (seconds) to remove DC in order to mitigate
-        internal calibration signal such as Caltone that can bias Doppler
-        estimation. If None, no DC removal in AZ.
+        Approximate AZ-block duration in (seconds) to remove DC in order to
+        mitigate internal calibration signal such as Caltone that can bias
+        Doppler estimation. Must be a positive value not greater than
+        `az_block_dur`! Its value may be modified to make it an integer
+        fraction of `az_block_dur`! If None, no DC removal in AZ.
     out_path : str, default='.'
         Output directory for dumping PNG files, if `plot` is True.
     plot : bool, default=False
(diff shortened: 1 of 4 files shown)


python/packages/nisar/pointing/doppler_lut_from_raw.py
```diff
@@ -518,6 +531,12 @@ def time_dop_est(echo, prf, lag=1, axis=None):
                     )
         else:  # single channel
             echo = raw_dset[slice_line]
+        # Remove DC in AZ to mitigate internal cal signals such as
+        # Caltone and its intermods that can largely bias Doppler
+        # centroid est.
+        if duration_dc_remove_az is not None:
+            for slice_rm_dc in _slice_gen(num_lines, nrgl_dc_rm):
+                echo[slice_rm_dc] -= np.nanmean(echo[slice_rm_dc], axis=0)
```
bhawkins, line 562: It makes sense to have separate parameters for `az_block_dur` and `duration_dc_remove_az`, but this implementation isn't ideal since there are many scenarios where `duration_dc_remove_az` is not respected. As `n=mod(az_block_dur, duration_dc_remove_az)` gets smaller the notch for the last block gets wider, up to the limit `n=1` where you're just zeroing out a pulse. This could be avoided by implementing some sort of producer/consumer pattern, e.g., by making a class derived from `isce3.io.BackgroundReader` that subtracts the blockwise mean.

The docstring does say it's really the max duration, so the behavior is documented. Also 0 Hz is almost antipodal to the expected NISAR Doppler centroid, so it doesn't actually matter much if the DC filter is wider than specified. Since this function is in the `nisar` namespace, I don't consider this a blocking issue.

bhawkins (approved):
This is an essential feature for commissioning. Thanks!

python/packages/nisar/pointing/doppler_lut_from_raw.py
rad-eng-59: Kindly reminder to review this PR prior to our next pointing exercise.  
FYI, put some checks and force the DC-removal block size to be an integer fraction of Doppler block size to avoid partial blocks. Updated default values to make it more practical and suitable for NISAR. Tested over `2-min` `80 MHz` SP half-swath with relatively larger topography using the default values as they are in this update. Not sure if I can post them here but I can definitely provide the results off-line if any JPL reviewer is interested.
Thank you.

hfattahi (approved):
LGTM! Thanks @rad-eng-59

rad-eng-59:
FYI, this workflow `test.cxx.isce3.geometry.metadata_cubes.metadata_cubes` causes failure on Mac OS in all recent PRs using the most updated develop branch. Any idea @gshiroma ? Thank you

hfattahi:
> FYI, this workflow `test.cxx.isce3.geometry.metadata_cubes.metadata_cubes` causes failure on Mac OS in all recent PRs using the most updated develop branch. Any idea @gshiroma ? Thank you

@rtburns-jpl @gmgunter I assume we continue ignoring this Mac failure?

gmgunter:
Ideally we should try to figure out what's causing it and resolve it. But yeah, for PRs like this one that are clearly unrelated to the metadata cubes, I think it makes sense to ignore the failing job as long as the other CI jobs are passing.

rad-eng-59:
@bhawkins  would you mind taking another look at this PR given new updates?
@hfattahi and @bhawkins I believe I should add Caltone removal to EL Null workflow as well to avoid potential artifacts in range direction while forming null in particular for  `RCID=34,35`. 
I'm not sure if there is enough time to have that addition being reviewed and approved by EOD Wednesday through ISCE3 review process. Let me know if you want me to open a similar PR for EL Null workflow to be reviewed by Wednesday. 
Otherwise, I'm going to implement it and use it off line. 
Thank you.

hfattahi:
@rad-eng-59 for RSLC processing, @bhawkins ended up with a more complex algorithm for Dithered data to remove caltone. I wonder if you need to do the same or do you think that is not needed for Doppler estimation?

bhawkins:
I'd hesitate to use this kind of estimator on dithered data anyhow.

rad-eng-59:
@hfattahi this Doppler estimator workflow doesn't support dithered data as explained in the users guide. So, we are good. Can I go ahead and merge it? It sees like there is no conflict either with the base branch.

PR #136, opened 2025-09-08
ST-EVD Replacement of Number of Range Blocks with Number of Range Samples

hb6688:
Removed:
num_rng_blks.

Added:
num_samples_rng_blk. Default = 256. num_samples_rng_blk > 2 x cpi_len. In addition, 5 x cpi_len is recommended

I am also advocating default number of emitters to be reduced from 16 to 12 as CPI Length = 32 pulses.  This is to avoid unnecessary RFI mitigation artifact

Please note that num_rng_blks is still kept as a parameter for FDNF.


python/packages/isce3/signal/rfi_process_evd.py
```diff
@@ -48,9 +48,9 @@ def run_slow_time_evd(
         time. Hence the standard (STD) deviation of multiple dominant EVs across slow time 
         defined by this parameter are compared. The one with the maximum STD is used for RFI
         Eigenvalue first difference computation.
-    num_rng_blks: int, default=1
-        Number of range bin blocks to be processed for EVD, default=1
-        When num_rng_blks=1, all range samples are used to compute EVD
+    num_samples_rng_blk: int
+        Number of range samples per range block for computation of sample covariance matrix.
+        default=256, in general num_samples_rng_blk > 2 x cpi_len is needed
```
gmgunter: For the sake of consistency, let's put the default value on the same line as the parameter type, since that's what we've been doing elsewhere in this docstring.

```suggestion
    num_samples_rng_blk: int, default=256
        Number of range samples per range block for computation of sample covariance matrix.
        In general, num_samples_rng_blk > 2 x cpi_len is needed
```

python/packages/nisar/workflows/focus.py
```diff
@@ -1101,7 +1101,7 @@ def process_rfi(cfg: Struct, raw_data: np.ndarray,
             num_max_trim=opt_evd.num_max_trim,
             num_min_trim=opt_evd.num_min_trim,
             max_num_rfi_ev=opt_evd.max_num_rfi_ev,
-            num_rng_blks=opt.num_range_blocks,
+            num_samples_rng_blk=opt.num_samples_rng_blk,
```
gmgunter, line 1104: Hmm, I don't think these changes are quite ready for review yet. Please test your changes to `focus.py` with both the ST-EVD algorithm and the FDNF algorithm. I don't see how either could be run without errors in the current state.

```suggestion
            num_samples_rng_blk=opt_evd.num_samples_rng_blk,
```

python/packages/nisar/workflows/focus.py
```diff
@@ -1111,7 +1111,7 @@ def process_rfi(cfg: Struct, raw_data: np.ndarray,
         rfi_likelihood = isce3.signal.rfi_freq_null.run_freq_notch(
             raw_data,
             opt_fnf.num_pulses_az,
-            num_rng_blks=opt.num_range_blocks,
+            num_samples_rng_blk=opt.num_samples_rng_blk,
```
gmgunter: The `isce3.signal.rfi_freq_null.run_freq_notch()` doesn't currently take a `num_samples_rng_blk` parameter, so this won't work.

https://github.com/isce-framework/isce3/blob/6093f479bca52ba4c903f403e9f853dfee2fb754/python/packages/isce3/signal/rfi_freq_null.py#L61-L75

Also, `opt` doesn't have a `num_samples_rng_blk` attribute.

Did you mean to do this instead?

```suggestion
            num_rng_blks=opt_fnf.num_range_blocks,
```

share/nisar/defaults/focus.yaml
```diff
@@ -214,17 +214,20 @@ runconfig:
...
+                    # is applied in range diretion. It is recommended that this 
+                    # parameter is at least 2 x cpi_len to avoid sample covariance 
+                    # estimation loss and 5 x cpi_len to have a desirable performance.
+                    num_samples_rng_blk: 256
                     # Maximum allowable number of emitters can be detected and
                     # suppressed per CPI. ST-EVD would ignore emitters exceeding
                     # this limit. This number should be less than cpi_length.
                     # Defaults to 16.
-                    max_emitters: 16
+                    max_emitters: 12
```
gmgunter: >I am also advocating default number of emitters to be reduced from 16 to 12 as CPI Length = 32 pulses. This is to avoid unnecessary RFI mitigation artifact

Can you give some more context for why we're changing this? Maybe you have some slides showing results with different values of `max_emitters` that you could share via email (not publicly)?

Also, this doesn't really seem related to the other changes in this PR. In general, I'd advocate for splitting up unrelated changes into separate PRs when possible, in order to make individual PRs easier to merge and to make the overall commit history more easily navigable.

share/nisar/defaults/focus.yaml
```diff
@@ -214,17 +214,20 @@ runconfig:
...
                     cpi_length: 32
+                    # Number of range samples per range block when data blocking
+                    # is applied in range diretion. It is recommended that this 
+                    # parameter is at least 2 x cpi_len to avoid sample covariance 
+                    # estimation loss and 5 x cpi_len to have a desirable performance.
+                    num_samples_rng_blk: 256
                     # Maximum allowable number of emitters can be detected and
                     # suppressed per CPI. ST-EVD would ignore emitters exceeding
                     # this limit. This number should be less than cpi_length.
                     # Defaults to 16.
```
gmgunter, line 234: Please update this comment to reflect the change to the default value.

```suggestion
                    # Defaults to 12.
```

share/nisar/defaults/focus.yaml
```diff
@@ -259,6 +262,8 @@ runconfig:
                         y: [5.0, 2.0]
 
                 freq_notch_filter:
+                    # Number of subdivisions in range.  Defaults to 1.
+                    num_range_blocks: 1
```
gmgunter: So the ST-EVD algorithm now takes `num_samples_rng_blk` as a parameter, whereas the FDNF algorithm still takes `num_range_blocks`. Is there any reason for this discrepancy?

I think your main reason for passing the block width (instead of the number of blocks) as a parameter was that this made it easier to use the same set of parameters for processing swaths of very different sizes (such as ALOS data vs NISAR freq A data vs NISAR freq B data). Doesn't that same justification also apply to the FDNF algorithm?

python/packages/isce3/signal/rfi_process_evd.py
gmgunter: >in general num_samples_rng_blk > 2 x cpi_len is needed

Is it *needed* or *recommended*?

In the YAML parameter description, you [say](https://github.com/isce-framework/isce3/blob/6093f479bca52ba4c903f403e9f853dfee2fb754/share/nisar/defaults/focus.yaml#L222-L224)

>It is recommended that this parameter is at least 2 x cpi_len to avoid sample covariance estimation loss and 5 x cpi_len to have a desirable performance.

What do you think about copying that description here in this docstring? It seems more clear and detailed than what you've written here.

gmgunter (commented):
Thanks, @hb6688. Here are a few initial comments.

python/packages/isce3/signal/rfi_process_evd.py
hb6688: @gmgunter , I found out about the failure in focus.py due to configuration updates in the YAML file and am working on resolving those errors.  In addition, I agree with your suggestion that FDNF should also incorporate num_samples_rng_blk instead of keep using num_rng_blks.  I wanted to make the changes for ST-EVD first but I think it is a good idea to do it for both right now.

share/nisar/defaults/focus.yaml
hb6688: @gmgunter , I think it was probably me who recommended this value to be at 16 at the time, which is half of the default CPI length, essentially allowing half of the overall number of pulses to be nulled in the worst case.  After processing a lot of ALOS dataset as well as NISAR commissioning data frames, I believe this number is too high in a lot of cases.  It is perhaps only needed in the extreme case.  The negative consequence of removing half of the pulses in the worst case even if there was RFI is most likely worse than that of not removing them.  The reason is that it is unlikely that the RFI Eigenvalues from 10-16 are dominant ones.  Their impact is far less than those from Eigenvalue 1-5 for example.  I am afraid that if this value is left at 16, some users may find quite a bit of RFI mitigation artifact.

python/packages/isce3/signal/rfi_process_evd.py
hb6688: @gmgunter , In general, an acceptable sample covariance estimation results in SINR degradation (after RFI processing) of less than ~3dB compared with that of using a true covariance matrix.  This requires at least cpi x 2 + 1 number of samples.  With the number of samples increasing to cpi x 5, we would have an SINR loss of ~1 dB.  The users can set whatever values they desire, but in my code, I raise an error if num_samples_rng_blk < 2 x cpi + 1.  With that in mind, I state that 2 x cpi number of samples are needed, but I think I need to make a correction that 2 x cpi + 1 number of samples are needed in order to run ST-EVD.

hb6688: I updated the wording for number of range samples per range block as follows. Note that FDNF does not estimate sample covariance matrix, hence having a data block in range with less than cpi x 2 + 1 number of samples is totally fine.

Number of range samples per range block when data blocking
is applied in range direction. It is recommended that this
parameter is at least 5 x cpi_len to avoid degradation  in SINR.
In addition, in order to avoid a run-time error for ST-EVD,
this parameter needs to be at least 2 x cpi + 1
num_samples_rng_blk: 256

hb6688: Please note that there is no 2 x cpi + 1 restriction for FDNF since there is no sample covariance estimation.

gmgunter: >In general, an acceptable sample covariance estimation results in SINR degradation (after RFI processing) of less than ~3dB compared with that of using a true covariance matrix. This requires at least cpi x 2 + 1 number of samples. With the number of samples increasing to cpi x 5, we would have an SINR loss of ~1 dB.

I'm kinda confused about exactly what this means.

By "RFI processing", do you mean RFI mitigation? Is it better to have a ~3dB "SINR degradation" or ~1db? If the RFI power is at least 3dB greater than the signal power, and the RFI processing removes only interference (it doesn't remove any signal), then presumably a 3dB SINR loss is better. But your comment seems to imply that 1dB is preferred.

How did you derive these specific SINR loss numbers?

>Number of range samples per range block when data blocking is applied in range direction.

Is there any circumstance when data blocking is *not* applied in the range direction? If not, I would probably omit the part about "when data blocking is applied in range direction".

>in order to avoid a run-time error for ST-EVD, this parameter needs to be at least 2 x cpi + 1

Could you please point out where that error will be raised in the code? I wasn't able to find it.

hb6688: @gmgunter
1.  The error checking for minimum number of samples is in compute_evd_cpi.py
2.  With regards to your question about minimum number of samples. I will try to clarify it.  In general, you would like to use as many samples as possible to estimate sample covariance matrix R^ where as the true covariance matrix R can be modeled analytically.  In the context of interference suppression/nulling, sample matrix inversion (SMI) is often used.  As you can tell from the terminology, matrix inversion is applied and this requires a good estimation in R^.  With this in mind, a rule of thumb from Reed-Mallete-Brennan (RMB) rule is that at least 2 x N number of samples need to be used, where N represents number of observations or channels.  Why 2 x N, because using 2N number of samples gives us a SINR after suppression of within ~3dB of that as if we were to use the true covariance R (SINR_R^ / SINR_R).  If you increase the number of samples in estimating R^, then the degradation is less. 5N number of samples gives a SINR loss versus that of using true covariance R of about ~1dB.

We are not doing SMI, neither are we doing MVDR beamformer. (I should point out EVD methodology is a variant of MVDR beamforming). However, we still would like to have a good estimate of the true covariance matrix so we can derive the mutually orthogonal eigenvectors accurately.  Therefore we still need to follow the RMB rule for minimum number of samples needed.  It is a trade-off between an accurate sample covariance matrix and unnecessary nulling of range samples.

hb6688:
I am not sure why the following unit test failed, any ideas?

 The following tests FAILED:
	 47 - test.cxx.isce3.geometry.metadata_cubes.metadata_cubes (Subprocess aborted)
Errors while running CTest

hb6688:
Changes in this PR:
ST-EVD and FDNF both uses num_samples_rng_blk intead of num_rng_blks.

bhawkins (approved):
Looks okay to me.

hb6688:
@bhawkins @hfattahi , I added a configuration parameter 'use_entire_pulse' to the configuration file because removing num_rng_blks option and replacing it with num_range_samples_blk takes away our ability to use all range samples from a range line since number of range samples is not an a priori.

share/nisar/schemas/focus.yaml
```diff
@@ -361,9 +360,16 @@ rfi_options:
   # Defaults to 'ST-EVD'.
   mitigation_algorithm: enum('ST-EVD', 'FDNF', required=False)
   # Number of subdivisions in range.  Defaults to 1.
-  num_range_blocks: int(min=1, required=False)
   slow_time_evd: include('slow_time_evd_options', required=False)
   freq_notch_filter: include('freq_notch_filter_options', required=False)
+  # Number of range samples per range block when data blocking
+  # is applied in range direction. It is recommended that this
+  # parameter is at least 5 x cpi_len to avoid SINR degradation.
```
hfattahi: Do you mean SNR?
```suggestion
  # parameter is at least 5 x cpi_len to avoid SNR degradation.
```

hfattahi: Or I assume you mean Signal To Interference pulse noise ratio? If yes, I would say spell it out.

hfattahi (approved):
LGTM.

share/nisar/schemas/focus.yaml
hb6688: @hfattahi I meant Signal to Interference Noise Ratio, but I think SNR degradation is a more accurate description now I think about it.

python/packages/isce3/signal/rfi_freq_null.py
```diff
@@ -85,10 +86,15 @@ def run_freq_notch(
...
         It is desirable to divide the raw data into processing blocks
         consisted of dimensions [num_pulses_az x num_samples_rng_blk]
-    num_rng_blks: int, default=1
-        Number of blocks in the range dimension, default=1
-        When num_rng_blks=1, all the range samples of the pulses within a 
-        processing block are used to estimate detection mask.
+    num_samples_rng_blk: int, default=256
+        Number of range samples per range block when data blockin is applied 
+        in range direction. It is recommended that this parameter is at least 
+        5 x cpi_len to avoid SINR degradation. In addition, in order to avoid
```
Tyler-g-hudson:
```suggestion
        5 x cpi_len to avoid SNR degradation. In addition, in order to avoid 
```
Following from your discussion with Heresh below

python/packages/isce3/signal/rfi_process_evd.py
```diff
@@ -48,9 +49,14 @@ def run_slow_time_evd(
         time. Hence the standard (STD) deviation of multiple dominant EVs across slow time 
         defined by this parameter are compared. The one with the maximum STD is used for RFI
         Eigenvalue first difference computation.
-    num_rng_blks: int, default=1
-        Number of range bin blocks to be processed for EVD, default=1
-        When num_rng_blks=1, all range samples are used to compute EVD
+    num_samples_rng_blk: int, default=256
+        Number of range samples per range block when data blockin is applied in range direction. 
+        It is recommended that this parameter is at least 5 x cpi_len to avoid SINR degradation.
```
Tyler-g-hudson:
```suggestion
        It is recommended that this parameter is at least 5 x cpi_len to avoid SNR degradation.
```

share/nisar/defaults/focus.yaml
```diff
@@ -595,6 +602,17 @@ runconfig:
...
+                    # will be generated in addition to the standard browse image PNG+KML.
+                    # The filename of each additional file will include a suffix noting
+                    # that image's frequency and polarization.
+                    # False to have the primary browse image be the only PNG+KML generated.
+                    # Note: If True, and if the input granule contains only one image,
+                    # then the additional individual PNG will be a duplicate image to the
+                    # primary browse image PNG, but with more descriptive filename.
+                    # Default: False
+                    output_individual_pngs: false
+
```
Tyler-g-hudson:
```suggestion
```
Was this added by mistake? It appears to be unrelated to the PR. Addition of new QA runconfig parameters should be handled by QA version update PR's, per @nemo794

tools/imagesets/runconfigs/rslc_REE2.yaml
```diff
@@ -123,8 +123,15 @@ runconfig:
                 # Algorithm selection.  Defaults to 'ST-EVD' for slow-time
                 # eigenvalue decomposition.
                 mitigation_algorithm: ST-EVD
-                # Number of subdivisions in range.  Defaults to 1.
-                num_range_blocks: 1
+                # Number of range samples per range block when data blocking
+                # is applied in range direction. It is recommended that this
+                # parameter is at least 5 x cpi_len to avoid SINR degradation.
```
Tyler-g-hudson:
```suggestion
                # parameter is at least 5 x cpi_len to avoid SNR degradation.
```

tools/imagesets/runconfigs/rslc_REE3.yaml
```diff
@@ -63,8 +63,15 @@ runconfig:
                 # Algorithm selection.  Defaults to 'ST-EVD' for slow-time
                 # eigenvalue decomposition.
                 mitigation_algorithm: ST-EVD
-                # Number of subdivisions in range.  Defaults to 1.
-                num_range_blocks: 1
+                # Number of range samples per range block when data blocking
+                # is applied in range direction. It is recommended that this
+                # parameter is at least 5 x cpi_len to avoid SINR degradation.
```
Tyler-g-hudson:
```suggestion
                # parameter is at least 5 x cpi_len to avoid SNR degradation.
```

python/packages/isce3/signal/rfi_freq_null.py
```diff
@@ -85,10 +86,15 @@ def run_freq_notch(
...
-        processing block are used to estimate detection mask.
+    num_samples_rng_blk: int, default=256
+        Number of range samples per range block when data blockin is applied 
+        in range direction. It is recommended that this parameter is at least 
+        5 x cpi_len to avoid SINR degradation. In addition, in order to avoid 
+        a run-time error for ST-EVD this parameter needs to be at least 
+        2 x cpi + 1.
+    use_entire_pulse: bool
+        Use all samples in the slow-time pulses for detection if this is True
+        default = False
```
Tyler-g-hudson, line 96: It'd be helpful on a quick read to know if this overwrites what was passed for `num_samples_rng_blk` or if there will be an error if both values are passed. The answer appears to be the former - that should be clarified here.
```suggestion
    use_entire_pulse: bool
        Ignore any value passed for `num_samples_rng_blk` and instead use all
        samples in the slow-time pulses for detection if this is True.
        default = False
```

Tyler-g-hudson (changes requested):
Largely looks good, just some documentation changes.
The only thing that needs to be fixed prior to merge is the addition of `output_individual_pngs` in the QA runconfig parameters; this needs to be removed unless @nemo794 has approved its presence.

share/nisar/defaults/focus.yaml
nemo794: Weird... this parameter is already be merged in ISCE3 via https://github.com/isce-framework/isce3/pull/151 . I'm not sure why it is showing up in the diff here. 

@hb6688 , perhaps rebasing on the main `develop` branch will remove this diff? (And the corresponding diff in the schema yaml?) Please address other comments first, and then perform the rebase. Once the rebase is complete, let's confirm that this diff has been removed before merging this PR, otherwise we might have unwanted merge artifacts.

Good catch, @Tyler-g-hudson !

hb6688: @nemo794 @Tyler-g-hudson , I grabbed Brian's PR136 and made changes on top of it to be able to run RFI on dithered PRF scenes.  I did not pay attention back then.  In short, I had to quickly make this work for PRF dithered cases, and that was probably where it got wrong.

Tyler-g-hudson (approved):
That was my only reservation about merging this, the rest LGTM!

Tyler-g-hudson:
Merging this since Bo is out

PR #141, opened 2025-09-16
Set RFI algo in RSLC output HDF5 based on input configuration items

hb6688:
During the RFI algorithm development, we used to have only ST-EVD.  Hence it is hardcoded to the output RSLC HDF5 file.  This PR simply updates algorithm stored in output RSLC HDF5 file based on settings in the input RSLC YAML configuration file.


bhawkins (approved):
LGTM

gmgunter (approved):
LGTM, thanks!

PR #143, opened 2025-09-17
RSLC azimuth window

bhawkins:
There are options in the focus.py configuration file for an azimuth apodization window, but the feature was never actually implemented because the azimuth antenna pattern was considered sufficient. However, the feature is useful in off-nominal scenarios like processing reduced resolution or when dealing with high-contrast scenes.

~~DRAFT STATUS: This PR is currently an ugly hack that only implements a raised cosine window. A Kaiser window would take more work to implement, especially considering that polymorphism is sort of annoying in CUDA. It's also important not to add a bunch of extra calculations to the innermost backprojection loop, though I guess it could be implemented as a post-processing step instead. I'm thinking using a `ChebyKernel` to represent the window might be a reasonably generic and efficient solution, but that'll have to wait a bit.~~

Edit: I changed the implementation to use a polynomial approximation to the window. It also skips the window calculation when possible, so there should be negligible impact on runtime in nominal processing (verified in my limited testing but TBC on a bigger scene).

### DIFF
--- a/cxx/isce3/cuda/focus/Backproject.cu
+++ b/cxx/isce3/cuda/focus/Backproject.cu
@@ -42,6 +42,10 @@ using DeviceDEMInterpolator = isce3::cuda::geometry::gpuDEMInterpolator;
 using DeviceOrbitView = isce3::cuda::core::OrbitView;
 using DeviceRadarGeometry = isce3::cuda::container::RadarGeometry;
 
+using HostWindow = isce3::core::ChebyKernel<float>;
+using DeviceWindow= isce3::cuda::core::ChebyKernel<float>;
+using DeviceWindowView = isce3::cuda::core::ChebyKernelView<float>;
+
 template<typename T>
 using DeviceLUT2d = isce3::cuda::core::gpuLUT2d<T>;
 
@@ -406,7 +410,8 @@ __global__ void sumCoherentBatch(
         const double* __restrict__ tau_atm_in,
         const int* __restrict__ kstart_in, const int* __restrict__ kstop_in,
         const size_t n, const double fc, const Kernel kernel,
-        const int batch_start, const int batch_stop, const float pedestal = 1.0f)
+        const int batch_start, const int batch_stop,
+        const std::optional<DeviceWindowView> window = std::nullopt)
 {
     // thread index (1d grid of 1d blocks)
     const auto tid = static_cast<size_t>(blockIdx.x) * blockDim.x + threadIdx.x;
@@ -434,9 +439,8 @@ __global__ void sumCoherentBatch(
     const double dtau = sampling_window.spacing();
     const auto samples = static_cast<size_t>(sampling_window.size());
 
-    const float win0 = 0.5f * (1.0f + pedestal);
-    const float win1 = 1.0f - win0;
-    const float win_freq = 2.0f * M_PI / (kstop - kstart - 1);
+    // scale pulse index to unit range
+    const float kscale = 1.0f / (kstop - kstart - 1);
 
     // loop over lines in batch
     thrust::complex<double> batch_sum = {0., 0.};
@@ -460,8 +464,13 @@ __global__ void sumCoherentBatch(
         ::sincospi(2. * fc * tau, &sin_phi, &cos_phi);
         z *= thrust::complex<double>(cos_phi, sin_phi);
 
-        const float window = win0 - win1 * std::cos(win_freq * (k - kstart));
-        batch_sum += window * z;
+        // assume branch prediction works better than an unnecessary multiply
+        if (window.has_value()) {
+            const auto x = (k - kstart) * kscale - 0.5f;
+            batch_sum += window.value()(x) * z;
+        } else {
+            batch_sum += z;
+        }
     }
 
     // add batch sum to total
@@ -479,7 +488,8 @@ ErrorCode backproject(std::complex<float>* out,
                       const Kernel& kernel, DryTroposphereModel dry_tropo_model,
                       const Rdr2GeoBracketParams& rdr2geo_params,
                       const Geo2RdrBracketParams& geo2rdr_params, int batch,
-                      float* height, float pedestal)
+                      const std::optional<DeviceWindowView> window,
+                      float* height)
 {
     // XXX input reference epoch must match output reference epoch
     if (out_geometry.referenceEpoch() != in_geometry.referenceEpoch()) {
@@ -686,7 +696,7 @@ ErrorCode backproject(std::complex<float>* out,
                 img.data().get(), rc.data().get(), pos.data().get(),
                 vel.data().get(), sampling_window, x.data().get(),
                 tau_atm.data().get(), kstart.data().get(), kstop.data().get(),
-                out_grid_size, fc, kernel, k, k + curr_batch, pedestal);
+                out_grid_size, fc, kernel, k, k + curr_batch, window);
 
         checkCudaErrors(cudaPeekAtLastError());
         checkCudaErrors(cudaStreamSynchronize(cudaStreamDefault));
@@ -709,12 +719,21 @@ ErrorCode backproject(std::complex<float>* out,
                       DryTroposphereModel dry_tropo_model,
                       const Rdr2GeoBracketParams& rdr2geo_params,
                       const Geo2RdrBracketParams& geo2rdr_params, int batch,
-                      float* height, float pedestal)
+                      const std::optional<HostWindow> window,
+                      float* height)
 {
     // copy inputs to device
     const DeviceRadarGeometry d_out_geometry(out_geometry);
     const DeviceRadarGeometry d_in_geometry(in_geometry);
     DeviceDEMInterpolator d_dem(dem);
+    std::optional<DeviceWindow> d_window;
+    std::optional<DeviceWindowView> dv_window = std::nullopt;
+    if (window.has_value()) {
+        // make a device copy held in function scope
+        d_window.emplace(DeviceWindow(window.value()));
+        // get a non-owning view
+        dv_window.emplace(d_window.value());
+    }
     ErrorCode ec;
 
     if (typeid(kernel) == typeid(HostBartlettKernel<float>)) {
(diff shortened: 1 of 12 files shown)


cxx/isce3/cuda/focus/Backproject.cu
```diff
@@ -456,7 +464,13 @@ __global__ void sumCoherentBatch(
...
         z *= thrust::complex<double>(cos_phi, sin_phi);
 
-        batch_sum += z;
+        // assume branch prediction works better than an unnecessary multiply
+        if (window.has_value()) {
+            const auto x = (k - kstart) * kscale - 0.5f;
+            batch_sum += window.value()(x) * z;
+        } else {
+            batch_sum += z;
+        }
```
gmgunter, line 473: This implementation seems reasonable to me but FYI GPUs don't have branch predictors.

https://chatgpt.com/share/68d426b9-befc-8011-9e95-9c34d82a22b1

gmgunter: If we want to over-engineer this to avoid a conditional inside the loop, I could imagine a possible approach that templates `sumCoherentBatch` on the window type, kinda like this:

```c++
struct NullWindow {};

template<class DeviceWindowView>
__global__ void sumCoherentBatch(..., DeviceWindowView window, ...)
{
    ...
    for (int k = batch_start; k < batch_stop; ++k) {
        ...
        if constexpr (std::is_same_v<DeviceWindowView, NullWindow>) {
            batch_sum += z;
        } else {
            const auto x = (k - kstart) * kscale - 0.5f;
            batch_sum += window.value()(x) * z;
        }
    }
    ...
}
```

Then, on the host side, you'd have to call it in a conditional like this:

```c++
ErrorCode backproject(..., const std::optional<HostWindow> h_window, ...)
{
    ...
    if (h_window.has_value()) {
        const auto d_window = DeviceWindow(h_window);
        const auto dv_window = DeviceWindowView(d_window);
        sumCoherentBatch<<<...>>>(..., dv_window, ...);
    } else {
        sumCoherentBatch<<<...>>>(..., NullWindow{}, ...);
    }
    ...
}
```

Not sure if that'd cause a noticeable improvement in runtime to be worth the additional code complexity.

cxx/isce3/cuda/focus/Backproject.cu
```diff
@@ -456,7 +464,13 @@ __global__ void sumCoherentBatch(
         ::sincospi(2. * fc * tau, &sin_phi, &cos_phi);
         z *= thrust::complex<double>(cos_phi, sin_phi);
 
-        batch_sum += z;
+        // assume branch prediction works better than an unnecessary multiply
+        if (window.has_value()) {
+            const auto x = (k - kstart) * kscale - 0.5f;
+            batch_sum += window.value()(x) * z;
```
gmgunter, line 470: It's not obvious to me that the support of the window function is necessarily $[-\frac{1}{2}, \frac{1}{2}]$.

Based on reading the implementation of [`ChebyKernel`](https://github.com/isce-framework/isce3/blob/4efc51149a674035ca0f127d367c22277072ee6e/cxx/isce3/core/Kernels.icc#L156-L211), it seems like the support domain is $[-\frac{w}{2}, \frac{w}{2}]$, where $w$ is the width of the kernel. Is that correct?

In that case, should we define `kscale` like this instead?

```diff
-    const float kscale = 1.0f / (kstop - kstart - 1);
+    const float w = window.has_value() ? window.value().width() : 1.0f;
+    const float kscale = w / (kstop - kstart - 1);
```

And then define `x` as:

```diff
-            const auto x = (k - kstart) * kscale - 0.5f;
+            const auto x = (k - kstart) * kscale - window.value().halfwidth();
```

(Note that we'd have to add the `halfwidth()` member function to [`Kernel`](https://github.com/isce-framework/isce3/blob/4efc51149a674035ca0f127d367c22277072ee6e/cxx/isce3/core/Kernels.h#L19).

cxx/isce3/cuda/focus/Backproject.cu
```diff
@@ -704,48 +719,57 @@ ErrorCode backproject(std::complex<float>* out,
                       DryTroposphereModel dry_tropo_model,
                       const Rdr2GeoBracketParams& rdr2geo_params,
                       const Geo2RdrBracketParams& geo2rdr_params, int batch,
+                      const std::optional<HostWindow> window,
```
gmgunter, line 722: General guidance is to pass inputs by const reference unless they're cheap to copy (typically 2-3 words in size).

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f16-for-in-parameters-pass-cheaply-copied-types-by-value-and-others-by-reference-to-const

`HostWindow` stores a `std::vector` [internally](https://github.com/isce-framework/isce3/blob/4efc51149a674035ca0f127d367c22277072ee6e/cxx/isce3/core/Kernels.h#L170) so it's potentially expensive to copy. Let's pass it by const reference here.

```suggestion
                      const std::optional<HostWindow>& window,
```

cxx/isce3/cuda/focus/Backproject.h
```diff
@@ -53,7 +55,9 @@ backproject(std::complex<float>* out,
             DryTroposphereModel dry_tropo_model = DryTroposphereModel::TSX,
             const Rdr2GeoBracketParams& rdr2geo_params = {},
             const Geo2RdrBracketParams& geo2rdr_params = {},
-            int batch = 1024, float* height = nullptr);
+            int batch = 1024,
+            const std::optional<isce3::core::ChebyKernel<float>> window = std::nullopt,
```
gmgunter, line 59: Hmm, I'm not seeing the same change in the corresponding .cu file. This overload seems to take a `const std::optional<isce3::cuda::core::ChebyKernelView<float>>` in that file, which doesn't match the type of the argument here in the header file. Did I miss something?

https://github.com/isce-framework/isce3/blob/4efc51149a674035ca0f127d367c22277072ee6e/cxx/isce3/cuda/focus/Backproject.cu#L491

cxx/isce3/cuda/focus/Backproject.cu
```diff
@@ -42,6 +42,10 @@ using DeviceDEMInterpolator = isce3::cuda::geometry::gpuDEMInterpolator;
 using DeviceOrbitView = isce3::cuda::core::OrbitView;
 using DeviceRadarGeometry = isce3::cuda::container::RadarGeometry;
 
+using HostWindow = isce3::core::ChebyKernel<float>;
+using DeviceWindow= isce3::cuda::core::ChebyKernel<float>;
```
gmgunter, line 46:
```suggestion
using DeviceWindow = isce3::cuda::core::ChebyKernel<float>;
```

cxx/isce3/cuda/focus/Backproject.h
```diff
@@ -85,6 +90,8 @@ backproject(std::complex<float>* out,
...
```
gmgunter, line 94: As mentioned in a previous comment, let's pass this by const ref.

```suggestion
            const std::optional<isce3::core::ChebyKernel<float>>& window = std::nullopt,
```

cxx/isce3/focus/Backproject.cpp
```diff
@@ -69,6 +82,7 @@ backproject(std::complex<float>* out, const RadarGeometry& out_geometry,
         const Kernel<float>& kernel, DryTroposphereModel dry_tropo_model,
         const isce3::geometry::detail::Rdr2GeoBracketParams& r2g_params,
         const isce3::geometry::detail::Geo2RdrBracketParams& g2r_params,
+        const std::optional<Window> window,
```
gmgunter, line 85:
```suggestion
        const std::optional<Window>& window,
```

cxx/isce3/focus/Backproject.h
```diff
@@ -44,6 +47,7 @@ backproject(std::complex<float>* out,
         DryTroposphereModel dry_tropo_model = DryTroposphereModel::TSX,
         const isce3::geometry::detail::Rdr2GeoBracketParams& r2g_params = {},
         const isce3::geometry::detail::Geo2RdrBracketParams& g2r_params = {},
+        const std::optional<isce3::core::ChebyKernel<float>> window = std::nullopt,
```
gmgunter, line 50:
```suggestion
        const std::optional<isce3::core::ChebyKernel<float>>& window = std::nullopt,
```

cxx/isce3/focus/Backproject.cpp
```diff
@@ -12,6 +12,7 @@
 #include <isce3/geometry/geometry.h>
 #include <isce3/geometry/rdr2geo_roots.h>
 #include <isce3/geometry/geo2rdr_roots.h>
+#include <isce3/math/complexOperations.h>
```
gmgunter, line 15: Hmm, why didn't we need this in the CUDA backprojection code?

python/packages/nisar/workflows/focus.py
```diff
@@ -1551,6 +1552,15 @@ def get_focused_sub_swaths(rawlist, out_chan, grid, orbit, doppler, dem, azres,
     return swaths
 
 
+def get_azimuth_window(cfg: Struct):
+    kind, shape = check_window_input(cfg.processing.azimuth_window,
+        msg="Azimuth window  ")
```
gmgunter, line 1557: Should we update the log message for the range window to specify "Range window" now that there are two of them?

https://github.com/isce-framework/isce3/blob/4efc51149a674035ca0f127d367c22277072ee6e/python/packages/nisar/workflows/focus.py#L1233

```diff
-    win_kind, win_shape = check_window_input(cfg.processing.range_window) 
+    win_kind, win_shape = check_window_input(cfg.processing.range_window, msg="Range window: ")
```

gmgunter, line 1557:
```suggestion
        msg="Azimuth window: ")
```

python/packages/nisar/workflows/focus.py
```diff
@@ -1551,6 +1552,15 @@ def get_focused_sub_swaths(rawlist, out_chan, grid, orbit, doppler, dem, azres,
     return swaths
 
 
+def get_azimuth_window(cfg: Struct):
```
gmgunter, line 1555: Let's annotate the return type as well.

```suggestion
def get_azimuth_window(cfg: Struct) -> isce3.core.ChebyKernelF32 | None:
```

share/nisar/defaults/focus.yaml
```diff
@@ -156,8 +156,8 @@ runconfig:
 
             # Azimuth spectral window, optional.  Defaults to no weighting.
             azimuth_window:
-                kind: Kaiser
-                shape: 0.0
+                kind: Cosine
+                shape: 1.0
```
gmgunter, line 160: Aren't these both boxcars? Can you comment on why you changed the default azimuth window here?

python/packages/isce3/focus/cheby_windows.py
```diff
@@ -0,0 +1,70 @@
...
+        if not (-0.5 <= t <= 0.5):
+            return np.float32(0)
+        x = self.beta * np.sqrt(1 - 4 * t * t)
+        return np.float32(np.i0(x) * self.scale)
+
+
+@unique
+class WindowKind(str, Enum):
+    KAISER = "kaiser"
+    COSINE = "cosine"
```
gmgunter, line 40: NBD, but I like to add something like this to `str`/`Enum` classes so that they behave a little bit more like Python 3.11 [`StrEnum`s](https://docs.python.org/3/library/enum.html#enum.StrEnum).

```suggestion
    COSINE = "cosine"
    
    def __str__(self) -> str:
        return self.value
```

Without overriding the `__str__` method, you'll get something like this when converting the enum to a string:

```python
>>> class WindowKind(str, Enum):
...     KAISER = "kaiser"
...     COSINE = "cosine"
...
>>> str(WindowKind.KAISER)
'WindowKind.KAISER'
```

With this change, you'll instead get this:

```python
>>> class WindowKind(str, Enum):
...     KAISER = "kaiser"
...     COSINE = "cosine"
...
...     def __str__(self):
...         return self.value
...
>>> str(WindowKind.KAISER)
'kaiser'
```

gmgunter (commented):
Thanks, this is a nice improvement to the backprojection code! Few comments below, but overall looks good.

gmgunter:
Just remembered to add the "GPU" label to this PR so that the CI system runs an additional job that builds & tests the CUDA code!

cxx/isce3/cuda/focus/Backproject.cu
bhawkins: Good point. How about making width==1 part of the interface and throwing an exception in the host code if it's not satisfied?

gmgunter: Sure, I guess that'd be OK with me if that's your preference.

I'm not sure why it'd be preferable to only support kernel width == 1, though. It doesn't seem like it'd be significantly less work to implement. Are you just thinking that the code will be cleaner if we require width == 1?

cxx/isce3/cuda/focus/Backproject.cu
bhawkins: I agree it'd be nice to avoid a copy. I would've thought you'd want an optional reference `std::optional<HostWindow&> window` rather than a reference to `optional`, but apparently that's not allowed. I guess I need to spend some time learning about `optional` semantics (and other C++ nuances).

PR #144, opened 2025-09-17
Add Static Layers workflow (MCR #90132)

gmgunter:
Implements the NISAR Static Layers workflow.

List of changes:

* Adds a new executable module `nisar.workflows.static` that can be used to run the workflow from the command line.
* Adds a default run configuration YAML file and a Yamale schema file for the workflow to `share/nisar/defaults/` and `share/nisar/schemas/`, respectively.
* Adds a new sub-package `nisar.static` to store utility classes and functions used by the Static Layers workflow.
* Copies the Static Layers product spec XML from the NISAR_PIX repo into `python/packages/nisar/products/XML/L2/`. (NOTE: this version of the spec XML includes the changes in [NISAR-ADT/NISAR_PIX#343](https://github-fn.jpl.nasa.gov/NISAR-ADT/NISAR_PIX/pull/343).)
* Adds small test datasets to `tests/data/` for unit testing the workflow, including a small scene near the south pole and one spanning the antimeridian. Each test dataset includes a DEM raster, a water mask raster, and a radar pointing XML file (the repo already includes orbit ephemeris XML files corresponding to these test cases).
* Adds a new function `isce3.geometry.infer_radar_grid_spacing_from_geo_grid()` to estimate the maximum possible radar grid spacing that could be used to avoid undersampling a user-specified geocoded grid.
* Updates the `isce3::core::saveToH5()` overload for serializing `Attitude` data to set the `"units"` and the `quaternions` dataset to `"1"`, matching the NISAR [spec](https://github-fn.jpl.nasa.gov/NISAR-ADT/NISAR_PIX/blob/7b59f5f09d770952a277135b8130aac536786da9/XML/L1/nisar_L1_RSLC.xml#L1689) for dimensionless quantities.

(The size of the diff is a bit intimidating, but please note that the bulk of the additional lines of code are due to the spec XML file, runconfig + schema YAML files, and radar pointing XML files used for testing -- the changes to the Python code are much smaller.)


python/packages/nisar/static/water_mask.py
```diff
@@ -0,0 +1,162 @@
...
+        indicates a not-water pixel, and 255 represents a no-data (invalid) pixel.
+
+    References
+    ----------
+    .. [1] J. Jung, "Water Mask Product Specification", JPL D-107710, 2024.
+    """
+    water_distance = np.asanyarray(water_distance)
+
+    # Get a binary mask of (ocean or inland) water pixels.
+    water = water_distance == 0
```
oberonia78, line 40:
```suggestion
    water = (water_distance >= 1) and (water_distance <=200)
```

python/packages/nisar/static/water_mask.py
```diff
@@ -0,0 +1,162 @@
...
+    r"""
+    Convert the input water distance map into a binary water mask.
+
+    Parameters
+    ----------
+    water_distance : array_like
+        The input water distance map, in the format specified by the NISAR Water Mask
+        Product Specification\ [1]_. A value of 0 indicates a water pixel. A value of
+        255 represents a no-data (invalid) pixel. Values in 1-200 represent non-water
+        pixels.
```
oberonia78, line 24:
```suggestion
        The input water distance map, in the format specified by the NISAR Water Mask
        Product Specification\ [1]_. A value of 0 indicates non-water pixel. A value of
        255 represents a no-data (invalid) pixel. Values in 1-200 represent water
        pixels.
```

tests/python/packages/nisar/static/water_mask.py
```diff
@@ -0,0 +1,66 @@
...
+
+import isce3
+
+
+def test_binarize_nisar_water_mask():
+    # Create three arrays of water pixels (filled with zeros), non-water pixels (filled
+    # with values in range [1, 200]), and invalid pixels (filled with 255).
+    water = np.zeros((20, 10), dtype=np.uint8)
+    nonwater = np.arange(1, 201, dtype=np.uint8).reshape(20, 10)
+    invalid = np.full((20, 10), fill_value=255, dtype=np.uint8)
```
oberonia78, line 16:
```suggestion
    nonwater = np.zeros((20, 10), dtype=np.uint8)
    water = np.arange(1, 201, dtype=np.uint8).reshape(20, 10)
    invalid = np.full((20, 10), fill_value=255, dtype=np.uint8)
```

tests/python/packages/nisar/static/water_mask.py
```diff
@@ -0,0 +1,66 @@
...
+
+    # Convert the water distance map into a binary water mask
+    water_mask = binarize_nisar_water_mask(water_distance)
+
+    # Check the datatype of the output mask.
+    assert water_mask.dtype == np.uint8
+
+    # Check the mask values.
+    np.testing.assert_equal(water_mask[:, :10], 1)
+    np.testing.assert_equal(water_mask[:, 10:20], 0)
```
oberonia78, line 29:
```suggestion
    np.testing.assert_equal(water_mask[:, :10], 0)
    np.testing.assert_equal(water_mask[:, 10:20], 1)
```

oberonia78 (commented):
Hi @gmgunter Thanks for the comprehensive work pulling together all the static layers. This is really awesome. I did notice a potential bug in the water-mask static layer: value 0 is being treated as water, which looks inverted (0 should indicate non-water).

python/packages/nisar/workflows/static.py
```diff
@@ -0,0 +1,379 @@
...
+    ----------
+    args : sequence of str or None, optional
+        The list of arguments. If None, the argument list is taken from `sys.argv`.
+        Defaults to None.
+    """
+    # Setup the argument parser.
+    parser = argparse.ArgumentParser(description="Run the NISAR Static Layers workflow")
+    parser.add_argument(
+        "config_file", type=Path, help="The run configuration YAML file"
+    )
```
gmgunter, line 372: Currently there's no option to specify a log file from the command line. Instead, everything gets logged to stdout and stderr. But we could certainly add this option if desired.

python/packages/nisar/workflows/static.py
```diff
@@ -0,0 +1,379 @@
...
+    dem_raster_file = dynamic_ancillary_files["dem_raster_file"]
+    water_mask_raster_file = dynamic_ancillary_files["water_mask_raster_file"]
+    logger.info(f"Open DEM raster file {dem_raster_file}")
+    dem_raster = isce3.io.Raster(dem_raster_file)
+    logger.info(f"Open water mask raster file {water_mask_raster_file}")
+    water_mask_raster = isce3.io.Raster(water_mask_raster_file)
+
+    # Construct a DEM interpolator.
+    dem_interp_method = processing_params["dem"]["interp_method"]
+    dem = isce3.geometry.DEMInterpolator(dem_raster)
```
gmgunter, line 68: If the input DEM raster is a global raster, this will try to load the entire raster contents into memory. This should be fine for operations since we'll stage a DEM that covers the scene, but another approach would be to load only the subset of the DEM required for processing.

share/nisar/defaults/static.yaml
```diff
@@ -0,0 +1,418 @@
...
+          y:
+
+      ephemeris:
+        # [OPTIONAL] UTC date and time of the start of the radar observation, as a
+        # string in ISO 8601 format with up to nanosecond precision.
+        # Orbit and attitude state vectors with earlier time tags will be discarded in
+        # order to avoid geo2rdr convergence errors due to orbit ambiguity.
+        # Must be <= `end_time`.
+        # Defaults to the start time of the orbit data in `orbit_xml_file`.
+        start_time:
```
gmgunter, line 177: Update the comment here to correctly describe the default value (it should be the start of the intersection of the orbit and attitude time intervals).

gmgunter, line 177: Update the comment with a warning that datetimes are truncated to microsecond precision if not quoted.

python/packages/nisar/workflows/static.py
```diff
@@ -0,0 +1,379 @@
...
+    # `infer_radar_grid_spacing_from_geo_grid()` or `get_bounding_radar_grid()` because
+    # these functions use `geo2rdr_bracket`, which takes different parameters than the
+    # legacy `geo2rdr` routine that's used by most of the workflow. Exposing both sets
+    # of parameters would introduce a lot of additional bookkeeping for seemingly little
+    # benefit.
+    logger.info("Estimate maximum required radar grid spacing")
+    radar_grid_params = processing_params["radar_grid"]
+    look_side = radar_grid_params["look_side"]
+    wavelength = radar_grid_params["wavelength"]
+    az_spacing, rg_spacing = isce3.geometry.infer_radar_grid_spacing_from_geo_grid(
```
gmgunter, line 101: We could add an option for the user to specify these radar grid parameters directly. If not specified, then we'll infer them from the geo grid (as is currently done).

python/packages/nisar/workflows/static.py
```diff
@@ -0,0 +1,379 @@
...
+                geo2rdr_params=geo2rdr_params,
+            )
+            reprojected_dem, los_east, los_north, local_inc_angle = geometry_layers
+
+        # Compute static mask layers (geocoded layover/shadow mask and water mask).
+        # Results are stored as GeoTIFF files in the scratch directory.
+        logger.info("Compute geocoded layover/shadow mask layer")
+        geocode_params = processing_params["geocode"]
+        with log_elapsed_time(logger.info, "Computing geocoded layover/shadow mask"):
+            layover_shadow_mask = compute_geocoded_layover_shadow_mask(
```
gmgunter, line 165: Add an option to decimate the radar grid for computing the layover/shadow mask. (The full radar grid will still be used to compute each of the other layers.)

python/packages/nisar/workflows/static.py
```diff
@@ -0,0 +1,379 @@
...
+                geo2rdr_params=geo2rdr_params,
+                **geocode_params,
+                **rtc_params,
+            )
+
+        # Infer the orbit pass direction from the orbit velocity vectors.
+        orbit_pass_direction = isce3.core.get_orbit_pass_direction(orbit)
+
+        # Pop 'product_counter' from the dict. This parameter is used to form the
+        # granule ID but doesn't correspond to any dataset in the `identification' group
```
gmgunter:
```suggestion
        # granule ID but doesn't correspond to any dataset in the 'identification' group
```

python/packages/nisar/static/product.py
```diff
@@ -0,0 +1,762 @@
...
+    chunk_size = tuple(chunk_size)
+
+    kwds = {}
+    if chunk_size != (-1, -1):
+        kwds["chunks"] = chunk_size
+        if compression_enabled:
+            kwds["compression"] = compression_type
+            kwds["shuffle"] = shuffle
+            if compression_type == "gzip":
+                kwds["compression_opts"] = compression_level
```
gmgunter: This is implicitly assuming that chunking is required in order to enable compression. Instead, we could just unconditionally populate the "compression" keyword in the output dict, and let h5py decide whether to ignore it (i.e. if chunking was disabled) or not.

python/packages/nisar/workflows/static.py
```diff
@@ -0,0 +1,379 @@
...
+            water_mask_description = get_raster_dataset_metadata_item(
+                water_mask_raster_file,
+                "water_mask_description",
+                default="(NOT SPECIFIED)",
+            )
+
+            # Populate the 'grids' group.
+            logger.info("Populate raster layers and grid coordinates in output HDF5")
+            grids_group = instrument_group.create_group("STATIC/grids")
+            dataset_creation_kwds = build_hdf5_dataset_creation_kwds_dict(
```
gmgunter, line 316: Currently, h5py will raise an error if the chunk size is larger than the dataset size. This is undesirable -- in this case, we should shrink the chunk size to match the dataset size.

python/packages/isce3/geometry/radar_grid_spacing.py
```diff
@@ -0,0 +1,149 @@
...
+        The look direction of the sensor (left-looking or right-looking).
+    wavelength : float
+        The radar central wavelength, in meters.
+    pts_per_side : int, optional
+        Side length of the NxN grid of samples used to estimate the required radar grid
+        pixel spacing necessary to avoid undersampling the geocoded grid. Must be >= 2.
+        Defaults to 5.
+    geo2rdr_params : dict or None, optional
+        An optional dict of parameters configuring the behavior of the root-finding
+        routine used in geo2rdr (Newton-Raphson implementation). The following keys are
```
gmgunter:
```suggestion
        routine used in geo2rdr (bracketing implementation). The following keys are
```

python/packages/isce3/geometry/radar_grid_spacing.py
```diff
@@ -0,0 +1,149 @@
...
+
+        'threshold':
+          Absolute azimuth time convergence tolerance, in seconds. Defaults to 1e-8.
+
+        'maxiter':
+          Maximum number of Newton-Raphson iterations. Defaults to 50.
+
+        'delta_range':
+          Step size for computing the numerical partial derivative of Doppler w.r.t
+          range, in meters. Defaults to 10.
```
gmgunter: Change these to the parameters for the bracketing version of geo2rdr

python/packages/isce3/geometry/radar_grid_spacing.py
```diff
@@ -0,0 +1,149 @@
...
+    #
+    # 1. Project the four corners of the pixel centered on the point (x, y) into radar
+    #    coordinates.
+    # 2. Compute the bounding box of the resulting quadrilateral in radar coordinates.
+    #    (NOTE: the transformation from geo -> rdr is not affine in general, but we
+    #    assume that the pixel perimeter is (approximately) a quadrilateral in radar
+    #    coordinates.)
+    # 3. Divide the bounding box into quadrants and choose the radar grid spacing to be
+    #    the dimensions of each quadrant.
+    def az_rg_spacing_at_xy_point(x: float, y: float) -> tuple[float, float]:
```
gmgunter, line 109: Upload the powerpoint presentation documenting this algorithm (or a few screenshots of the relevant slides) to Google Drive and include a link here.

python/packages/nisar/static/granule_id.py
```diff
@@ -0,0 +1,172 @@
...
+    return template.substitute(
+        MISSION=mission_id,
+        I=radar_band,
+        L=product_level,
+        PROD=product_type,
+        REL=int_to_3_digit_string(relative_orbit_number),
+        P=orbit_direction_to_char_code(orbit_pass_direction),
+        FRM=int_to_3_digit_string(frame_number),
+        Xposting=int_to_3_digit_string(int(round(x_posting))),
+        Yposting=int_to_3_digit_string(int(round(y_posting))),
```
gmgunter, line 167: Should we do something smarter here if `x_posting` and `y_posting` are in degrees instead of meters? (Note, this doesn't affect production because the output geo grid will always be in projected coordinates.)

gmgunter (commented):
Reviewed this together offline w/ @gshiroma @hfattahi @nemo794 @bhawkins @oberonia78

python/packages/nisar/static/water_mask.py
gmgunter: >value 0 is being treated as water, which looks inverted (0 should indicate non-water).

Thanks for reviewing the code! Would you mind double-checking this -- I think I'm correctly implementing it such that `1` is a water pixel and `0` is a non-water pixel in the output water mask.

In this function, `water_distance` represents the distance from water, in kilometers. So pixels where `water_distance == 0` are considered water pixels (i.e. their distance from water is zero). 

The `water` variable computed here is therefore a binary array where water pixels have the value `1` and non-water pixels have value `0`.

I can try to update the comments to make this a bit less confusing.

The unit test might help clarify the behavior a bit:

https://github.com/isce-framework/isce3/blob/24d0a7b6dd904ab0e7df746399e23ed99c0743dc/tests/python/packages/nisar/static/water_mask.py#L12-L30

The first 10 columns of the `water_distance` array are water pixels. The next 10 columns are non-water pixels, with water distances of 1-100 kilometers.

The test verifies that the first 10 columns of the output `water_mask` are filled with `1`s and the next 10 columns are filled with `0`s.

python/packages/nisar/static/water_mask.py
gmgunter: Note that this description corresponds to the input water distance map -- not the output binary water mask.

In the input water distance map, I expect that water pixels should have a value of 0 and non-water pixels will have a value of 1-200, depending on their distance from ocean or inland water.

So I think the current description is correct (but please correct me if I'm still misunderstanding!).

python/packages/nisar/workflows/static.py
gmgunter: I don't think I'll have time to implement + test this right away, but if we want to change the behavior in the future, there's an existing ISCE3 function that seems to do what we want. See here:

https://github.com/isce-framework/isce3/blob/8cc6581525e608cc95fdcd883dac87e08f83e7d0/python/extensions/pybind_isce3/geometry/DEMInterpolator.cpp#L147

python/packages/nisar/static/geo_grid.py
```diff
@@ -0,0 +1,114 @@
...
+    bottom_right : dict
+        The X and Y coordinates of the lower-right corner of the geo grid bounding box,
+        in the native units of the CRS defined by the `epsg` argument. A dict containing
+        'x' and 'y' keys. If either value in the dict is `None`, it is inferred from the
+        bounding box of `dem_raster`.
+    posting : dict
+        The X and Y pixel spacing of the output grid, in the native units of the CRS
+        defined by the `epsg` argument. A dict containing 'x' and 'y' keys. If either
+        value in the dict is `None`, it is inferred from the pixel spacing of
+        `dem_raster`.
```
gmgunter: Document that these are assumed to be positive valued

python/packages/nisar/static/geo_grid.py
```diff
@@ -0,0 +1,114 @@
...
+    def get_option(param: T | None, *, default: T) -> T:
+        return param if (param is not None) else default
+
+    # X and Y coordinate of the upper-left corner pixel of the output geo grid.
+    xmin = get_option(top_left["x"], default=dem_raster.x0)
+    ymax = get_option(top_left["y"], default=dem_raster.y0)
+
+    # X and Y spacing of the output geo grid.
+    dx = get_option(posting["x"], default=dem_raster.dx)
+    dy = -posting["y"] if (posting["y"] is not None) else dem_raster.dy
```
gmgunter: Currently, you can have a scenario where `posting.x` is not None but `posting.y` is None (and therefore is inferred from the DEM). Do we want to disallow this?

Same for `top_left` and `bottom_right`.

gmgunter: We could check whether the output epsg matches the DEM epsg. If not, then all geo grid parameters must be specified.

python/packages/nisar/static/geo_grid.py
```diff
@@ -0,0 +1,114 @@
...
+
+    # X and Y coordinate of the lower-right corner pixel of the output geo grid.
+    xmax = get_option(bottom_right["x"], default=dem_x_lr)
+    ymin = get_option(bottom_right["y"], default=dem_y_lr)
+
+    # Ensure that the bounding box is valid.
+    if not (xmax >= xmin) or not (ymax >= ymin):
+        raise ValueError(
+            f"invalid bounding box: {xmax=} must be >= {xmin=} and {ymax=} must be >="
+            f" {ymin}"
```
gmgunter:
```suggestion
            f" {ymin=}"
```

gmgunter:
```suggestion
    if not (xmax > xmin) or not (ymax > ymin):
        raise ValueError(
            f"invalid bounding box: {xmax=} must be > {xmin=} and {ymax=} must be >"
            f" {ymin}"
```

python/packages/nisar/static/geo_grid.py
```diff
@@ -0,0 +1,114 @@
...
+
+    # Divide `num` by `den` and round the result to the nearest integer.
+    def round_divide(num: float, den: float) -> int:
+        return int(np.round(num / den))
+
+    # Length and width of the output geo grid.
+    # XXX: It seems more sensible to always round up rather than rounding to the nearest
+    # integer to ensure the geo grid covers the entire requested bounding box, but the
+    # other L2 SAS workflows do it this way. We should ensure this works the same way so
+    # that the grids for any particular frame are consistent.
```
gmgunter: Open an issue about this and link to that issue here

EDIT: #155

python/packages/nisar/static/water_mask.py
```diff
@@ -0,0 +1,163 @@
...
+        prefix="reprojected-water-mask_",
+        suffix=".tif",
+    )
+
+    reproject_raster(
+        water_mask_raster_file,
+        reprojected_water_mask_raster_file,
+        geo_grid=geo_grid,
+        algorithm=resample_algorithm,
+    )
```
gmgunter: This is binarizing the input water distance map and *then* re-projecting to the output geo grid.

This could be an issue if the input water mask was a global dataset, because the entire dataset will get converted to a large binary raster -- not just the subset of the dataset that corresponds to the output geo grid.

OTOH, if we're using "mode" resampling, binarizing the data before re-projecting makes sense, because the "mode" option doesn't work well with non-categorical data.

Offline, with @gshiroma, we decided to change the order of operation such that the water distance map is re-projected to the output geo grid before converting to a binary mask. In addition, let's change the default resampling method to nearest neighbor instead of "mode".

gmgunter: Note that the InSAR workflow does things differently -- it first re-projects the water distance map using "mode" resampling and then converts to a binary mask. It also doesn't respect the fill value of 255.

Let's open an issue and see if we can improve this (and harmonize the two workflows).

EDIT: #156

python/packages/nisar/static/rtc_anf_layers.py
```diff
@@ -0,0 +1,231 @@
...
+        Minimum pixel area threshold for RTC, in dB. If set to NaN, no minimum is
+        enforced, which may lead to division by very small values and result in infinite
+        or inaccurate outputs. Defaults to NaN.
+    dem_upsample_factor : float, optional
+        DEM upsampling factor used for RTC. Defaults to 2.
+    algorithm : isce3.geometry.RtcAlgorithm or str, optional
+        Algorithm used for RTC. Defaults to the area projection algorithm.
+    area_beta_mode : isce3.geometry.RtcAreaBetaMode or str or None, optional
+        Method for estimating the area of the beta naught reference surface. Defaults to
+        ???.
```
gmgunter: Fix this

python/packages/nisar/static/rtc_anf_layers.py
```diff
@@ -0,0 +1,231 @@
...
+
+        'maxiter':
+          Maximum number of Newton-Raphson iterations. Defaults to 50.
+    output_mode : isce3.geocode.GeocodeOutputMode or str, optional
+        Algorithm used for geocoding. Defaults to the area projection algorithm.
+    interp_method : isce3.core.DataInterpMethod or str, optional
+        Interpolation method used for geocoding. Only relevant when `output_mode` is set
+        to interpolation. Ignored otherwise. Defaults to biquintic interpolation.
+    memory_mode : isce3.core.GeocodeMemoryMode, optional
+        Block processing mode to use for geocoding. Defaults to ???.
```
gmgunter: Fix this

python/packages/nisar/static/rtc_anf_layers.py
```diff
@@ -0,0 +1,231 @@
...
+        RTC scaling factor to normalize backscatter coefficients from gamma0 to sigma0,
+        accounting for local terrain.
+    gamma0_to_sigma0_factor : isce3.io.Raster
+        RTC scaling factor to normalize backscatter coefficients from gamma0 to beta0,
+        accounting for local terrain.
+    """
+    if geo2rdr_params is None:
+        geo2rdr_params = {}
+
+    geocode = isce3.geocode.GeocodeCFloat32()
```
gmgunter: This will run faster if we geocode real-valued data instead of complex-valued.

Note that we don't care about the geocoded covariance output. The gamma0-to-beta0 and gamma0-to-sigma0 layers that we do care about should be unaffected.

```suggestion
    geocode = isce3.geocode.GeocodeFloat32()
```

python/packages/nisar/static/rtc_anf_layers.py
```diff
@@ -0,0 +1,231 @@
...
+        geocode.threshold_geo2rdr = geo2rdr_params["threshold"]
+    if "maxiter" in geo2rdr_params:
+        geocode.numiter_geo2rdr = geo2rdr_params["maxiter"]
+
+    # XXX: The input SLC and output geocoded covariance rasters are required arguments
+    # but aren't used in the computation of any of the output layers of interest here,
+    # so simply pass in dummy rasters.
+    dummy_slc = make_scratch_gtiff(
+        shape=(radar_grid.length, radar_grid.width),
+        dtype=np.complex64,
```
gmgunter:
```suggestion
        dtype=np.float32,
```

python/packages/nisar/static/rtc_anf_layers.py
```diff
@@ -0,0 +1,231 @@
...
+    )
+    dummy_output_rtc_sigma = make_scratch_gtiff(
+        shape=(radar_grid.length, radar_grid.width),
+        dtype=np.float32,
+        dir_=scratch_dir,
+        prefix="dummy-output-rtc-sigma_",
+    )
+
+    # TODO: Should we pass in an input layover/shadow mask? Currently, GCOV doesn't seem
+    # to do this. We should probably adopt the same strategy here in Static Layers.
```
gmgunter: Checked with @gshiroma. Let's remove this comment.

share/nisar/defaults/static.yaml
```diff
@@ -0,0 +1,430 @@
...
+      # Must conform to the Composite Release Identifier (CRID) format specified in the
+      # L-SAR Product File Naming Conventions document (JPL D-102255).
+      # Defaults to 'D99999'.
+      composite_release_id: D99999
+      # [OPTIONAL] Processing type.
+      # Used to populate the '/science/{L,S}SAR/identification/processingType' dataset
+      # in the output product.
+      # Must be one of:
+      #   - 'PR': nominal (production)
+      #   - 'OD': custom (on-demand)
```
gmgunter:
```suggestion
      #   - 'OD': custom (off-nominal on-demand processing)
```

share/nisar/defaults/static.yaml
```diff
@@ -0,0 +1,430 @@
...
+      # Defaults to 'OD'.
+      processing_type: OD
```
gmgunter, line 71: This is kinda weird for static layers because the production system *is* the on-demand system.

What we really want here is some way to distinguish between production jobs and custom processing jobs. Maybe there's a better way to do this.

python/packages/nisar/static/product.py
```diff
@@ -0,0 +1,780 @@
...
+        A disclaimer string to accompany the water mask dataset in the granule.
+    """
+
+    # Create a dataset in the 'grids' group. (This shouldn't be used for creating large
+    # raster datasets -- instead use `create_raster_layer_dataset()`, defined below.)
+    def create_grids_dataset(name: str, data: ArrayLike) -> h5py.Dataset:
+        full_name = "/".join((grids_group.name, name))
+        dataset_spec = product_spec.get_dataset_spec(full_name)
+        dataset = create_dataset(grids_group.file, dataset_spec, data)
+        copy_units_attr_from_dataset_spec(dataset, dataset_spec)
```
gmgunter, line 375: This always copies the units from the product spec XML, which assumes that the output grid is in projected coordinates with units of "meters".

If someone ran a custom job where the output grid is in geodetic coordinates, these units will be wrong.

Should we instead infer the units by looking at the epsg code of the output geo grid?

python/packages/nisar/static/product.py
```diff
@@ -0,0 +1,780 @@
...
+    y_coords_dataset = create_grids_dataset("yCoordinates", y_coords)
+
+    # Treat `xCoordinates` and `yCoordinates` as dimension scales. Other NISAR workflows
+    # don't assign names to dimension scales so let's take the same approach here.
+    x_coords_dataset.make_scale()
+    y_coords_dataset.make_scale()
+
+    proj_dataset = create_grids_dataset("projection", geo_grid.epsg)
+    proj_dataset_attrs = build_projection_dataset_attrs_dict(geo_grid.epsg)
+    proj_dataset.attrs.update(proj_dataset_attrs)
```
gmgunter, line 400: Note that this is doing much the same thing as `set_get_geo_info()` in `h5_prep.py`

https://github.com/isce-framework/isce3/blob/526f9c477ee41928c00a26022584b6846abfe917/python/packages/nisar/workflows/h5_prep.py#L665

Perhaps, we should simply call that function here to avoid redundant functionality.

gmgunter (commented):
Reviewed this together offline as a group

tests/python/packages/isce3/geometry/radar_grid_spacing.py
```diff
@@ -0,0 +1,156 @@
...
+    ----------
+    initial_position : (float, float, float)
+        The position of the platform, in meters, in the Earth-centered, Earth-fixed
+        (ECEF) coordinate system at time t=0.
+    velocity : (float, float, float)
+        The velocity of the platform, in meters per second, in the same coordinate
+        system as `initial_position`.
+    times : iterable of float
+        The timepoints at which to sample the orbit state vectors, in seconds relative
+        to `epoch`.
```
nemo794: Can the times occur prior to the epoch? If not, what about:

```suggestion
        The timepoints at which to sample the orbit state vectors, in seconds since
        `epoch`.
```

share/nisar/defaults/static.yaml
```diff
@@ -0,0 +1,436 @@
...
+      # [OPTIONAL] UTC date and time of the start of the validity date range for the
+      # resulting product, as a string in ISO 8601 format with integer seconds
+      # precision.
+      # The output STATIC product is valid for use with NISAR products with observation
+      # datetimes not earlier than this datetime but earlier than the validity start
+      # datetime of any future superseding STATIC product.
+      # Defaults to '1970-01-01T00:00:00'.
+      # WARNING: Due to limitations with the way this YAML file is parsed, datetimes
+      # with sub-microsecond precision will be truncated to integer microseconds
+      # precision unless surrounded in quotes.
```
nemo794: This comment/warning is probably not needed here, because this parameter has a precision of integer seconds?

share/nisar/defaults/static.yaml
```diff
@@ -0,0 +1,436 @@
...
+        # If `epsg` was specified and differs from the DEM raster EPSG code, these
+        # parameters are required. Otherwise, they default to the coordinates of the
+        # upper-left corner of the input DEM raster.
+        top_left:
+          x:
+          y:
+        # [OPTIONAL] X and Y coordinates of the lower-right corner pixel in the output
+        # grid, in the coordinate system and units described by `epsg`.
+        # Must satisfy `bottom_right.x` > `top_left.x` and `bottom_right.y` <
+        # `top_left.y`.
```
nemo794: Within the sentence, would you be open to having "left" and "right" correspond to the left and right sides of the inequality?

```suggestion
        # Must satisfy `top_left.x` < `bottom_right.x` and `top_left.y` >
        # `bottom_right.y`.
```

python/packages/nisar/products/XML/L2/nisar_L2_STATIC.xml
```diff
@@ -0,0 +1,618 @@
+<?xml version="1.0"?>
+<!-- product specification version is 1.4.0 -->
+<algorithm name="L2_StaticLayers"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:noNamespaceSchemaLocation="../../SDS/pix/schema.xsd">
```
gshiroma:
```suggestion
           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
```
This schema file does not exist in ISCE3. We remove this line from the other XMLs too (see example [here](https://github.com/isce-framework/isce3/blob/8cc6581525e608cc95fdcd883dac87e08f83e7d0/python/packages/nisar/products/XML/L2/nisar_L2_GCOV.xml#L4)).

share/nisar/defaults/static.yaml
```diff
@@ -0,0 +1,436 @@
...
+        # runtime of the process.
+        # Must be one of:
+        #   - 'single_block': process the full scene as a single block
+        #   - 'geogrid': partition the geographic image grid into sub-blocks; process
+        #     the full radar image grid as a single block
+        #   - 'geogrid_and_radargrid': partition the geographic and radar image grids
+        #     into sub-blocks
+        #   - 'auto': the block processing mode is chosen by the software
+        # Defaults to 'geogrid_and_radargrid'.
+        memory_mode: geogrid_and_radargrid
```
gshiroma:
```suggestion
        memory_mode: auto
```
Let's keep auto here and tell the SDS to use `geogrid_and_radargrid` to generate products with fine spacing.

share/nisar/defaults/static.yaml
```diff
@@ -0,0 +1,436 @@
...
+        # Typically one of {'gzip', 'lzf', 'szip'}. See
+        # https://docs.h5py.org/en/stable/high/dataset.html#filter-pipeline for details.
+        # Ignored if `compression_enabled` is false.
+        # Defaults to 'gzip'.
+        compression_type: gzip
+        # [OPTIONAL] Level of compression applied by the GZIP compression filter.
+        # Must be an integer from 0 to 9 (inclusive).
+        # Ignored if `compression_enabled` is false or `compression_type` is not 'gzip'.
+        # Defaults to 4.
+        compression_level: 4
```
gshiroma:
```suggestion
        compression_level: 1
```
I'd suggest changing the `compression_level` to `1` as this can reduce runtime without impacting too much the file size. We use `compression_level` equal `1` for [GCOV](https://github-fn.jpl.nasa.gov/isce-3/isce/blob/94bedfb2e7a94c8c0b84056ae3ee9f62d50d036c/share/nisar/defaults/gcov.yaml#L107) and [GSLC](https://github-fn.jpl.nasa.gov/isce-3/isce/blob/94bedfb2e7a94c8c0b84056ae3ee9f62d50d036c/share/nisar/defaults/gslc.yaml#L99).

gshiroma (commented):
Looks great, @gmgunter ! I went through most of the code a second time, paying closer attention to the details, and it still looks excellent. Thank you for your hard work on this PR. I just have a few minor comments and suggestions below, but overall this PR is very close to being ready to merge.

gmgunter:
Thanks, @gshiroma! I just addressed all of your recent comments.

share/nisar/defaults/static.yaml
```diff
@@ -0,0 +1,436 @@
...
+        # Defaults to 1024.
+        lines_per_block: 1024
+
+      geocode:
+        # [OPTIONAL] Algorithm used for geocoding.
+        # Must be one of:
+        #   - 'interp': interpolation
+        #   - 'area_projection': pixel area-based adaptive multilooking algorithm
+        # Defaults to 'area_projection'.
+        output_mode: area_projection
```
nemo794, line 301: I know area projection is used for GCOV. Is that also true for GSLC and InSAR? GSLC and InSAR runconfigs do not include this parameter...

If not, would the STATIC outputs be geocoded such that they are (slightly) off from the GSLC/InSAR?
Or if so, does this parameter only apply to the GCOV-centric layers?

Could you please update the comment accordingly?

python/packages/nisar/static/granule_id.py
```diff
@@ -0,0 +1,172 @@
...
+    Parameters
+    ----------
+    processing_center : str
+        A string denoting the data processing center of a NISAR Static Layers granule,
+        e.g. 'JPL'.
+
+    Returns
+    -------
+    str
+        'J' for 'JPL'; 'X' otherwise.
```
nemo794, line 84: I worry that someone will pass in "J" via the runconfig, and then get "X" returned.

Would it make sense to allow for if a single character (e.g. "J", "I", etc.) is passed in, then that would be capitalized and returned? And then the default would be "X".

python/packages/nisar/workflows/static.py
```diff
@@ -0,0 +1,378 @@
...
+
+
+def run_static_layers_workflow(config_file: os.PathLike | str) -> None:
+    """
+    Run the NISAR Static Layers workflow with the specified run configuration file.
+
+    Parameters
+    ----------
+    config_file : path-like
+        The file path of the runconfig YAML file.
```
nemo794: Pendatic, but can we make this docstring specific to a runconfig yaml file for STATIC layers? (i.e. not an RSLC runconfig yaml?)

I don't like this phrasing, but to give an idea of what I'm asking:

```suggestion
        The file path to a STATIC layers runconfig YAML file.
```

python/packages/nisar/workflows/static.py
```diff
@@ -0,0 +1,378 @@
...
+from nisar.static.water_mask import binarize_and_reproject_water_mask
+
+import isce3
+from isce3.geometry import make_geo_grid_bounding_polygon
+
+
+def run_static_layers_workflow(config_file: os.PathLike | str) -> None:
+    """
+    Run the NISAR Static Layers workflow with the specified run configuration file.
+
```
nemo794, line 38: Let's note what the output is?

```suggestion
    Run the NISAR Static Layers workflow with the specified run configuration file.

    Will generate a single STATIC HDF5 granule, as specified in the runconfig.
```

python/packages/nisar/workflows/static.py
```diff
@@ -0,0 +1,378 @@
...
+            nisar.products.populate_global_attrs_from_spec(hdf5_file, product_spec)
+
+            # Get the current processing datetime (truncated to integer seconds
+            # precision).
+            processing_datetime = datetime.now(timezone.utc).replace(microsecond=0)
+
+            # XXX: It's not really obvious what should go in the `zeroDopplerStartTime`
+            # and `zeroDopplerEndTime` datasets in the 'identification' group. For now,
+            # we'll use the start & stop time of the radar grid, which is roughly
+            # analogous what they represent in other NISAR L2 products.
```
nemo794, line 274: In the XML, the description reads as: "Azimuth start time (in UTC) of the product based on the time of the orbit state vectors used to generate the product[...]"

What would the impacts be if the input radar grid is changed? Would there be impacts if e.g. number of looks is added to the product (different radar grid)?

Nothing to fix now, of course. But if you think this could be something that needs future discussion/work, could you please add this to the list of follow-up / TODO items?

Thanks!

python/packages/nisar/workflows/static.py
```diff
@@ -0,0 +1,378 @@
...
+    Parameters
+    ----------
+    args : sequence of str or None, optional
+        The list of arguments. If None, the argument list is taken from `sys.argv`.
+        Defaults to None.
+    """
+    # Setup the argument parser.
+    parser = argparse.ArgumentParser(description="Run the NISAR Static Layers workflow")
+    parser.add_argument(
+        "config_file", type=Path, help="The run configuration YAML file"
```
nemo794: Could we be specific about the contents of the file?

```suggestion
        "config_file", type=Path, help="Run configuration YAML for the STATIC workflow"
```

nemo794 (commented):
To echo @gshiroma , this PR is quite beautiful to read! Thank you @gmgunter for all of the work and care that went into combing through ISCE3, PIX, specification documents, to produce this PR -- incredible!

We reviewed this PR together as a group, and most of my comments occurred at that time. A few add'l comments here, mostly clarifying docstrings and comments.

🚀

share/nisar/defaults/static.yaml
gmgunter: Good note. To my knowledge, GSLC and InSAR do not use the area projection mode.

The geocoding step in the InSAR workflow seems to hard-code the mode to `INTERP`:

https://github.com/isce-framework/isce3/blob/8cc6581525e608cc95fdcd883dac87e08f83e7d0/python/packages/nisar/workflows/geocode_insar.py#L513

The metadata LUTs in the GSLC and GCOV products are also always geocoded using `INTERP` mode:

https://github.com/isce-framework/isce3/blob/8cc6581525e608cc95fdcd883dac87e08f83e7d0/python/packages/nisar/products/writers/BaseL2WriterSingleInput.py#L2200

In the Static Layers workflow, we want to always geocode the layover/shadow mask using `INTERP` mode as well:

https://github.com/isce-framework/isce3/blob/391527b08d71a547a0b044e46423081b993404b0/python/packages/nisar/static/layover_shadow_mask.py#L236

So this `output_mode` parameter in the Static Layers runconfig only affects the computation of the RTC ANF layers. I'll make a note of that in the description.

As such, there's no danger of these layers in the Static Layers product disagreeing with corresponding layers in the GSLC/InSAR products, because there are no RTC ANF layers in those products. The default output mode for both GCOV and Static Layers is area projection, so those two products should be in agreement here.

https://github.com/isce-framework/isce3/blob/8cc6581525e608cc95fdcd883dac87e08f83e7d0/share/nisar/defaults/gcov.yaml#L286

python/packages/nisar/workflows/static.py
gmgunter: >What would the impacts be if the input radar grid is changed?

Currently, there's no input radar grid -- the radar grid is computed internally by the workflow.

The start and end time here will depend on the time tags in the input orbit and pointing XML data, though. Nominally, I think we plan to use orbit & pointing data from the first cycle of the science orbit in order to generate the initial set of Static Layers products. But we could alternatively use the 2nd or 3rd or Nth cycle, with nearly equivalent results, except that the absolute UTC start and end time of the product would be different. So this start & end time metadata seems sort of arbitrary, IMO.

>Would there be impacts if e.g. number of looks is added to the product (different radar grid)?

It depends on the implementation. One approach we've discussed is letting the user specify the radar grid *spacing* in the runconfig, but not the start/end time of the radar grid -- those would still be computed internally by the workflow. In that case, the question is moot. But if we instead decide to make the start/end time of the radar grid configurable in the runconfig, then it could have an impact on the `numberOfLooks` layer.

python/packages/nisar/static/granule_id.py
gmgunter: Fair point, but I'm not sure that's the right way to handle it.

Remember that the processing center is also used to populate the `/science/{L,S}SAR/identification/processingCenter` dataset in the output product. I think it's preferable to use human-readable strings to describe the processing center, rather than resorting to single-character mnemonics. 

If there's a need to make the granule ID logic smarter to add special codes for other processing centers, I'd suggest that we modify this function to add additional cases besides "JPL" --> "J". For example, we could add another case that maps "NRSC" --> "N", if they want to run the same code to generate their own Static Layers products.

Since the code is open-source, other processing centers should feel welcome to open an issue or PR requesting changes here.

tests/python/packages/isce3/geometry/radar_grid_spacing.py
gmgunter: Negative time points are allowed, yes, but I think the suggestion is good anyway.

gmgunter:
Thanks so much @oberonia78, @gshiroma, and @nemo794 for your careful review and helpful comments. I think I've addressed or responded to all of your comments, apart from those that we've discussed offline. Please let me know if you have any other questions/suggestions!

python/packages/nisar/static/granule_id.py
nemo794: Makes sense. Perhaps this behavior could be documented better in the runconfig... future work...! 😬

share/nisar/defaults/static.yaml
```diff
@@ -0,0 +1,439 @@
...
+      # [OPTIONAL] Radio frequency band.
+      # Used to form the granule ID and populate the
+      # '/science/{L,S}SAR/identification/radarBand' dataset in the output product.
+      # Defaults to 'L'.
+      radar_band: L
+      # [OPTIONAL] Data processing center.
+      # Used to form the granule ID and populate the
+      # '/science/{L,S}SAR/identification/processingCenter' dataset in the output
+      # product.
+      # Defaults to '(NOT SPECIFIED)'.
```
nemo794, line 97: Per https://github.com/isce-framework/isce3/pull/144#discussion_r2399942917

```suggestion
      # '/science/{L,S}SAR/identification/processingCenter' dataset in the output
      # If 'JPL' is specified, the processing center character in the granule ID
      # will be populated with a 'J'; otherwise it will be populated with an 'X'.
      # Defaults to '(NOT SPECIFIED)'.
```

nemo794 (approved):
One minor suggestion, your call. Thank you again -- LGTM!

gshiroma (approved):
Looks great, @gmgunter ! Thanks again for your hard work on this PR. I must say that when I saw a PR with almost 9000 new lines at first was a bit daunting, but the code is very clean and well documented, which made the review process much smoother. I also appreciate that you scheduled meetings to go over the changes with us, as well as all the additional discussions around the static layers product, specs, and production rules. Thank you for dedicating so much time and effort to this product! Excellent work, @gmgunter, as always!

PR #146, opened 2025-09-22
Remove caltone with 2D variation

bhawkins:
Currently focus.py can remove a fast-time stationary caltone using a parametric notch filter, or it can remove a slow-time stationary caltone by nulling the azimuth mean. This PR adds a the capability to remove a caltone that varies in both directions (though the fast-time variation has to be rather slow to still be considered a tone at all).

The estimator could be considered a single-scale wavelet transform, or a single-bin STFT. The estimation and removal use COLA windows to avoid block-boundary artifacts. This seems to work well for the cases I've examined so far.

Note the config file interface changes. One gotcha is that to disable caltone mitigation you have to set the parameter to "none" or "disabled" or something because `null` will get overridden by the non-null default value. I'll fix that if we end up actually needing this stuff. For now I just want to get the feature out there for testing.

### DIFF
--- a/python/packages/isce3/focus/caltone.py
+++ b/python/packages/isce3/focus/caltone.py
@@ -2,15 +2,24 @@ import numpy as np
 from isce3.signal import cola_windows
 
 class ToneRemover:
+    """
+    Coherently estimate and remove a non-stationary tone-like signal.
+    """
     def __init__(self, frequency, n, window_size=64, dtype=np.complex64):
         """
         Parameters
         ----------
         frequency : float
+            The nominal tone frequency, normalized to radians/sample, e.g.,
             (f_caltone - fc) / fs
+            Note there is no check for aliasing.
         n : int
+            Number of samples in the signal to analyze.
         window_size : int, optional
-        dtype : numpy.dtype
+            Number of samples in each estimation block.  Estimates will be
+            spaced by half the window_size.
+        dtype : numpy.dtype, optional
+            Data type for filter banks.
         """
         self.dtype = dtype
         self.size = n
@@ -23,12 +32,19 @@ class ToneRemover:
             self.wavelets.append((block, analysis, synthesis))
 
     def analyze(self, z):
+        """Estimate the time-varying response of a signal `z` to a
+        tone-like filter bank.  Akin to a single-bin STFT.
+        `z` should be one-dimensional with length self.size.
+        """
         if len(z) != self.size:
             raise ValueError(
                 f"Planned for length={self.size} but got {len(z)}")
         return np.array([w.dot(z[block]) for (block, w, _) in self.wavelets])
 
     def synthesize(self, coeffs):
+        """Generate a continuous time-domain signal given its filter bank
+        response coefficients.  Akin to a single-bin inverse STFT.
+        """
         if len(coeffs) != len(self.wavelets):
             raise ValueError("Need one coefficient per wavelet.")
         z = np.zeros(self.size, self.dtype)
@@ -37,6 +53,9 @@ class ToneRemover:
         return z
 
     def remove_tone(self, z):
+        """Estimate parameters of a time-varying tone and return the signal
+        `z` with the tone subtracted.
+        """
         if z.ndim != 1:
             raise NotImplementedError("Only 1D estimation is implemented")
         coeffs = self.analyze(z)
--- a/python/packages/isce3/focus/caltone.py
+++ b/python/packages/isce3/focus/caltone.py
@@ -1,0 +1,43 @@
+import numpy as np
+from isce3.signal import cola_windows
+
+class ToneRemover:
+    def __init__(self, frequency, n, window_size=64, dtype=np.complex64):
+        """
+        Parameters
+        ----------
+        frequency : float
+            (f_caltone - fc) / fs
+        n : int
+        window_size : int, optional
+        dtype : numpy.dtype
+        """
+        self.dtype = dtype
+        self.size = n
+        # NOTE Use a consistent phase reference across blocks.
+        tone = np.exp(-1j * 2 * np.pi * frequency * np.arange(n))
+        self.wavelets = []
+        for block, window in cola_windows(n, window_size):
+            analysis = (window * tone[block] / np.sum(window)).astype(dtype)
+            synthesis = (window * tone[block]).conjugate().astype(dtype)
+            self.wavelets.append((block, analysis, synthesis))
+
+    def analyze(self, z):
+        if len(z) != self.size:
+            raise ValueError(
+                f"Planned for length={self.size} but got {len(z)}")
+        return np.array([w.dot(z[block]) for (block, w, _) in self.wavelets])
+
+    def synthesize(self, coeffs):
+        if len(coeffs) != len(self.wavelets):
+            raise ValueError("Need one coefficient per wavelet.")
+        z = np.zeros(self.size, self.dtype)
+        for i, (block, _, w) in enumerate(self.wavelets):
+            z[block] += coeffs[i] * w
+        return z
+
+    def remove_tone(self, z):
+        if z.ndim != 1:
+            raise NotImplementedError("Only 1D estimation is implemented")
+        coeffs = self.analyze(z)
+        return z - self.synthesize(coeffs)
--- a/python/packages/isce3/signal/cola_windows.py
+++ b/python/packages/isce3/signal/cola_windows.py
@@ -13,7 +13,7 @@ def cola_windows(n: int, window_size: int = 64) -> Iterator[tuple[slice, np.ndar
     n : int
         Length of array to analyze.
     window_size : int, optional
-        Length of analysis blocks.
+        Length of analysis blocks. Must be even-valued. Defaults to 64.
 
     Yields
     -------
--- a/python/packages/isce3/signal/cola_windows.py
+++ b/python/packages/isce3/signal/cola_windows.py
@@ -1,0 +1,40 @@
+import numpy as np
+from typing import Iterator
+
+def cola_windows(n: int, window_size: int = 64) -> Iterator[tuple[slice, np.ndarray]]:
+    """
+    Generate windows with the constant-overlap-add (COLA) property.
+    Hann windows are generated with 50% overlap and sum to one.
+    The first segment is a half window, and final two windows may
+    be partial.
+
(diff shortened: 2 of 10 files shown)


hb6688 (approved):

python/packages/isce3/signal/cola_windows.py
```diff
@@ -0,0 +1,40 @@
...
+    Hann windows are generated with 50% overlap and sum to one.
+    The first segment is a half window, and final two windows may
+    be partial.
+
+    Parameters
+    ----------
+    n : int
+        Length of array to analyze.
+    window_size : int, optional
+        Length of analysis blocks.
```
gmgunter:
```suggestion
        Length of analysis blocks. Must be even-valued. Defaults to 64.
```

python/packages/isce3/focus/caltone.py
```diff
@@ -0,0 +1,43 @@
...
+    def __init__(self, frequency, n, window_size=64, dtype=np.complex64):
+        """
+        Parameters
+        ----------
+        frequency : float
+            (f_caltone - fc) / fs
+        n : int
+        window_size : int, optional
+        dtype : numpy.dtype
+        """
```
gmgunter, line 23: Need to add a description of the class and its arguments

python/packages/isce3/focus/caltone.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def synthesize(self, coeffs):
+        if len(coeffs) != len(self.wavelets):
+            raise ValueError("Need one coefficient per wavelet.")
+        z = np.zeros(self.size, self.dtype)
+        for i, (block, _, w) in enumerate(self.wavelets):
+            z[block] += coeffs[i] * w
+        return z
+
+    def remove_tone(self, z):
```
gmgunter, line 55: Need docstrings here

Let's make sure to clarify the expected shape of the input arrays (is this a single block of data? A single rangeline?)

python/packages/isce3/focus/caltone.py
```diff
@@ -0,0 +1,43 @@
+import numpy as np
+from isce3.signal import cola_windows
+
+class ToneRemover:
+    def __init__(self, frequency, n, window_size=64, dtype=np.complex64):
```
gmgunter, line 8: We discussed this in our review offline.

I think the things that most need to be clarified via docstrings/comments are:

* The `analyze` method is effectively performing a single-bin FFT to get the spectral component of the signal data at the caltone frequency. 
* The `synthesize` method is performing the inverse FFT to get that signal component back in the time domain
* Once we have that signal component in the time domain, we can simply remove it from the original signal data.
* We apply the FFT in blocks because (while the caltone frequency remains constant) the amplitude and phase of the caltone may vary across the range line.

share/nisar/defaults/focus.yaml
```diff
@@ -415,11 +415,18 @@ runconfig:
...
+                # If "azimuth_mean", subtract the azimuth mean from the raw
+                # data, which is useful when the caltone is nearly constant in
+                # azimuth but not stationary in range.  Calculating the mean
+                # and subtracting it is done in blocks using
+                # rangecomp.block_size
+                # If "wavelet" then subtract an estimate that varies in 2D.
+                # No correction when null.
+                # wavelet | azimuth_mean | null
+                algorithm: azimuth_mean
+                frequency: 1216.904e6
```
gmgunter:
```suggestion
                # RF frequency of the caltone, in hertz.
                frequency: 1216.904e6
```

share/nisar/defaults/focus.yaml
```diff
@@ -415,11 +415,18 @@ runconfig:
...
+                wavelet_size: 64
```
gmgunter, line 439: How about somethign like this? (Feel free to improve verbiage)

```suggestion
                # Length of constant-overlap add (COLA) window used
                # for wavelet caltone removal. Ignored if `algorithm` is not
                # 'wavelet'.
                wavelet_size: 64
```

share/nisar/defaults/focus.yaml
```diff
@@ -415,11 +415,18 @@ runconfig:
...
+            caltone:
+                # If "azimuth_mean", subtract the azimuth mean from the raw
+                # data, which is useful when the caltone is nearly constant in
+                # azimuth but not stationary in range.  Calculating the mean
+                # and subtracting it is done in blocks using
+                # rangecomp.block_size
+                # If "wavelet" then subtract an estimate that varies in 2D.
+                # No correction when null.
+                # wavelet | azimuth_mean | null
+                algorithm: azimuth_mean
```
gmgunter: Consider adding another option "auto" here, that would allow the SAS to decide which algorithm to apply

gmgunter: Add another option called "disabled" to turns off caltone removal. Otherwise, there's no way for users to disable this. Currently, `null` will just get overridden by the default algorithm in the default runconfig.

tests/python/packages/isce3/focus/caltone.py
```diff
@@ -0,0 +1,20 @@
...
+    tone = np.exp(1j * 2 * np.pi * f * k)
+    # slow sinusoidal amplitude and quadratic phase
+    modulation = (1.0 + 0.5 * np.cos(2 * np.pi / n * k)) * np.exp(1j * np.pi * (k / n)**2)
+    modulated_tone = tone * modulation
+
+    remover = ToneRemover(f, n)
+    estimated = remover.synthesize(remover.analyze(modulated_tone))
+
+    assert all(np.abs(np.abs(estimated) - np.abs(modulated_tone)) < 0.02)
+    assert all(np.abs(np.angle(estimated * modulated_tone.conjugate())) < 0.1)
```
gmgunter: This change might be nice to get slightly more informative error messages if either assertion fails

```suggestion
    np.testing.assert_array_less(np.abs(np.abs(estimated) - np.abs(modulated_tone)), 0.02)
    np.testing.assert_array_less(np.abs(np.angle(estimated * modulated_tone.conjugate())), 0.1)
```

tests/python/packages/isce3/focus/caltone.py
```diff
@@ -0,0 +1,20 @@
...
+
+    removed = remover.remove_tone(modulated_tone)
+    assert all(np.abs(removed) < 0.1)
```
gmgunter:
```suggestion
    np.testing.assert_array_less(np.abs(removed), 0.1)
```

tests/python/packages/isce3/focus/caltone.py
gmgunter: Let's add a comment indicating that the threshold is high because of ringing near the edges

gmgunter (commented):
Reviewed this in PR Fridays 2025-09-26

share/nisar/defaults/focus.yaml
```diff
@@ -415,11 +415,28 @@ runconfig:
...
+                # If "wavelet" then subtract an estimate that varies in 2D.
+                # No correction when "disabled".
+                # Option "auto" means "wavelet" if dithered else "azimuth_mean",
+                # which is a sensible default for NISAR LSAR.
+                # auto | wavelet | azimuth_mean | disabled
+                algorithm: auto
+
+                # Calibration tone frequency (RF) in Hz.
+                # NOTE Caltone frequency changed for most modes between mode
+                # table v50 and v51, which went went into operations on
```
hfattahi:
```suggestion
                # table v50 and v51, which went into operations on
```

hfattahi (commented):
The original implementation was extensively tested. Let's test the new changes (auto algorithm).

python/packages/nisar/workflows/focus.py
```diff
@@ -1551,6 +1551,51 @@ def get_focused_sub_swaths(rawlist, out_chan, grid, orbit, doppler, dem, azres,
...
+    # azimuth mean in NISAR data since DC is outside the azimuth passband.
+    # For other systems like ALOS that's not such a great assumption.
+    if algorithm == "auto":
+        algorithm = "wavelet" if is_dithered else "azimuth_mean"
+
+    wavelets = None
+    if algorithm == "azimuth_mean":
+        log.info("Will remove azimuth mean from each block.")
+    elif algorithm == "wavelet":
+        log.info("Will remove wavelet caltone estimate from each pulse.")
```
Tyler-g-hudson, line 1589:
```suggestion
        log.info("Caltone Removal: Will remove azimuth mean from each block.")
    elif algorithm == "wavelet":
        log.info("Caltone Removal: Will remove wavelet caltone estimate from each pulse.")
    elif algorithm == "disabled":
        log.info("Caltone Removal: Will not be applied.")
```
Prefixing these lines with a short descriptive title makes it easier to search them in logs.

share/nisar/defaults/focus.yaml
```diff
@@ -425,7 +425,11 @@ runconfig:
                 # No correction when null.
                 # wavelet | azimuth_mean | null
                 algorithm: azimuth_mean
-                frequency: 1216.904e6
+                # NOTE Caltone frequency changed for most modes between mode
+                # table v50 and v51, which went went into operations on
+                # 2025-10-17T00:00:00Z.  The v50 frequency is 1216.904297e6 Hz
+                # while the v51 frequency is 1214.882813e6 Hz.
+                frequency: 1214.882813e6
```
Tyler-g-hudson, line 436: I think it isn't, but just to be certain - is caltone frequency recorded anywhere in the L0B? It would be nice if there was an `auto` option for this as well, that just removed caltone at the recorded value.

Tyler-g-hudson (approved):
A couple of very minor comments, we've already reviewed this so I just looked at the most recent set of commits.

LGTM, pending some testing we'll do today. I don't imagine there's much in the new code that would cause problems.

share/nisar/defaults/focus.yaml
Tyler-g-hudson: This comment is not meant to change code in this PR - given time constraints, this can be a later PR if it turns out to be possible.

bhawkins: In principle it could be derived from `/science/LSAR/RRSD/lowRateTelemetry/DRT/MISC/CP_IFSW_CALTONE_PHASE_STEP_H`, but it'd be better to just add a field under the swaths group (or maybe an attribute of the caltone dataset) with units of Hz.

PR #149, opened 2025-09-24
Fix the same number of looks values in the GUNW metadata

xhuang-jpl:
This PR is to fix the same number of looks values for the both wrapped and unwrapped interferogram in the GUNW metadata reported by Ian Joughin.  The issue is raised because the GUNW writer class is inherent from both the RIFG and RUNW writer classes where those two variables  `self.igram_range_looks` and `self.igram_azimuth_looks`  are  overwritten when the function `add_interferogram_to_procinfo_params_group` is called by one of the RIFG and RUNW writers.


python/packages/nisar/products/insar/GUNW_writer.py
```diff
@@ -148,6 +148,20 @@ def add_parameters_to_procinfo_group(self):
...
+        wrap_igram_azimuth_looks = proc_cfg["crossmul"]["azimuth_looks"]
+        unwrap_rg_looks = proc_cfg["phase_unwrap"]["range_looks"]
+        unwrap_az_looks = proc_cfg["phase_unwrap"]["azimuth_looks"]
+
+        unwrap_igram_range_looks = wrap_igram_range_looks
+        unwrap_igram_azimuth_looks = wrap_igram_azimuth_looks
+
+        if (unwrap_az_looks != 1) or (unwrap_rg_looks != 1):
+            unwrap_igram_range_looks = unwrap_rg_looks
+            unwrap_igram_azimuth_looks = unwrap_az_looks
```
gmgunter, line 160: @xhuang-jpl, we briefly reviewed this offline and you gave some context for this change (thank you!):

* The number of looks may differ between the wrapped product (RIFG) and the unwrapped product (RUNW).
* When the number of looks in the `phase_unwrap` group of the runconfig is set to 1 (in both azimuth and range), then no additional looks are taken in the unwrapping step, so the number of looks in the unwrapped phase is the same as the number of looks in the wrapped interferogram. This is also documented in the runconfig:

https://github.com/isce-framework/isce3/blob/32642e423e5c6385a35535d6e05facbd0222d118/share/nisar/defaults/insar.yaml#L597-L600

Thanks for this clarification. I think the thing I'm still confused about is this:

If the number of looks in the `phase_unwrap` group of the runconfig is not 1, should it be the *total* number of looks, or the *additional* number of looks *after* multilooking by the number of looks specified in the `crossmul` group of the runconfig?

xhuang-jpl: Thank you @gmgunter, if the number of looks in the `phase_unwrap`   is not 1, then it will be the total number of looks. The `unwrapping` and `crossmul` have independent number of looks.

hfattahi: Yes I agree that it should be the total number of looks in that situation. The user would only care about the final number of looks and if looks are taken in one step or two steps is just implementation detail.

hfattahi (approved):
LGTM!

python/packages/nisar/products/insar/GUNW_writer.py
Tyler-g-hudson:
```suggestion
        if (unwrap_az_looks != 1) or (unwrap_rg_looks != 1):
            unwrap_igram_range_looks = unwrap_rg_looks
            unwrap_igram_azimuth_looks = unwrap_az_looks
        else:
            unwrap_igram_range_looks = wrap_igram_range_looks
            unwrap_igram_azimuth_looks = wrap_igram_azimuth_looks
```
Just a little nitpick - these variables are always going to be assigned either way, this just organizes the assignments in an if/else block for fast reading of the logic

Tyler-g-hudson (approved):
LGTM

PR #152, opened 2025-09-25
Fix compilation with CUDA 13

rtburns-jpl:
CUDA 13 removed some deprecated error codes, and now requires using cudaDeviceGetAttribute to access the compute mode. The changes here are guarded behind CUDART_VERSION checks to ensure that builds with previous CUDA versions function the same as before.

### DIFF
--- a/cxx/isce3/cuda/except/Error.cpp
+++ b/cxx/isce3/cuda/except/Error.cpp
@@ -25,18 +25,21 @@ static const char* cufftGetErrorString(const cufftResult err) {
             return "CUFFT_INVALID_SIZE";
         case CUFFT_UNALIGNED_DATA:
             return "CUFFT_UNALIGNED_DATA";
+#if defined(CUDART_VERSION) && CUDART_VERSION < 13000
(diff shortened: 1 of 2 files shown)


gmgunter (approved):
Thanks for the fix! LGTM, pending the CI checks passing

rtburns-jpl:
Oh, and for reference, the removed error codes are documented as deprecated by cuFFT in 12.9 and removed in 13.0 here: https://docs.nvidia.com/cuda/cufft/deprecated-functionality.html

I was not able to find any central documentation on the removal of the computeMode member of the device prop struct, but the CUDA samples repo made a similar change (to instead access it via cudaDevAttrComputeMode) in their changes for the 13.0 release, and noted in their changelog that it had been deprecated+removed: 
https://github.com/NVIDIA/cuda-samples/compare/v12.9...v13.0
https://github.com/NVIDIA/cuda-samples/blob/3f1c50965017932fc81e6d94a3fc9e04c105b312/CHANGELOG.md#cuda-130

PR #157, opened 2025-10-02
Add documentation builder github action

rtburns-jpl:
This action runs on all pull requests, and publishes the documentation on merge.

High level documentation uses mkdocs. C++ API documentation is built with doxygen, and Python API documentation is generated from its docstrings using pdoc. ISCE3 is built with CUDA support so that the pybind CUDA extension's API is also documented.

Currently, the Python API documentation appears very barebones because pdoc uses `__all__` to decide what to document, which we do not populate. Going forward, we will need to add `__all__` attributes to our Python package's submodules with all the members we want documented.


mkdocs.yml
```diff
@@ -5,7 +5,7 @@ nav:
   - Home: index.md
   - Installation: buildinstall.md
   - Python API: api/python/isce3/core.html
```
gmgunter: Not related to the changes in this PR, but I noticed that when I click on "Python API" link from the landing page, it takes me directly to the API docs for the `isce3.core` submodule, instead of the top-level `isce3` module.

I think it would make sense to change this to point users to the API docs for the top-level module. Otherwise, folks who aren't familiar with the API might be confused and think the library is a lot smaller than it is. WDYT?

```suggestion
  - Python API: api/python/isce3.html
```

gmgunter: Somewhat relatedly, we should probably find some way to document the API of the `nisar` package as well, if possible.

I suppose that would require another call to `pdoc`?

Then, I could imagine that we might want to include links to both the `isce3` API docs and the `nisar` API docs in this navigation sidebar. Maybe it would look something like this?

```suggestion
  - Python API:
      - isce3: api/python/isce3.html
      - nisar: api/python/nisar.html
```

gmgunter (approved):
Thanks, @rtburns-jpl! All the changes look great. I was able to run this docs build procedure locally and verify that the resulting HTML pages all look good.

hfattahi (approved):

PR #159, opened 2025-10-04
Fix dithered data processing for long observations

bhawkins:
Once upon a time, focus.py did the range processing for all pulses in an observation. This is obviously inefficient when we're only focusing a small subset of the input data, so we added the feature to read and process a subset of the raw data. However, that broke two things related to dithered data processing:
* The `resample` function assumed that the first pulse time was equivalent to the first time in the uniform output grid. Following the raw data subsetting change, that was only true when the subset includes the first pulse or other limited cases. It's no longer true in general.
* The `resample` function assumes the subswath annotations are aligned with the raw data. However, the raw data subsetting change neglected to properly subset this argument to `resample`.

This patch fixes the time offset by just using the `RadarGridParameters.sensing_times` attribute. It fixes the subswath subsetting issue by slicing it right away. That means the code diff actually doesn't touch `resample` but instead touches the two other places it's used, which is a slightly cleaner solution.

When debugging this I was getting annoyed having to figure out which debug file belonged to which channel, so I also added this info to the file names.


hfattahi (approved):
LGTM. This PR has been extensively tested.

Tyler-g-hudson (approved):
LGTM

PR #161, opened 2025-10-09
Fix bug in rubbersheet to process freq B interferogram

oberonia78:
The bug in rubbersheet introduces the error in processing frequency B. The radar_grid was defined from only the frequency A domain.


xhuang-jpl (approved):
LGTM, thank you @oberonia78

hfattahi (approved):
LGTM! The fix allows to process Frequency B only through InSAR which is nice for debugging.

PR #164, opened 2025-10-23
Add bridge algorithm to unwrap

oberonia78:
This PR adds the bridge functionality to the unwrap. 
For the fixed PRF interferogram, the phase jumps occurs along the subswath boundaries. To minimize the issues, this PR masked out the invalid areas (partially focused area) using mask layer, then applies the bridge algorithm.
image


hfattahi (commented):
Looks good to me. Would you please confirm if you have tested with latest constant PRF and Dithered data?

oberonia78:
Currently, there is no issue in running the bridging algorithm for dithered mode but we don't see the differences with/without bridge algorithm, because there is no gap between subswaths for now. @hfattahi

python/packages/isce3/unwrap/preprocess.py
```diff
@@ -376,3 +376,48 @@ def project_map_to_radar(cfg, input_data_path, freq):
 
     # stack to make whole then return
     return output_arrays
+
+
+def interpret_subswath_mask(subswath_mask, nodata=255):
```
xhuang-jpl: Thank you @oberonia78, this is a very helpful function, would you please making it a generic function to put it [here](https://github.com/isce-framework/isce3/blob/develop/python/packages/nisar/products/utils.py)

xhuang-jpl (approved):
Thank you @oberonia78 , LGTM, just a minor suggestion

python/packages/isce3/unwrap/preprocess.py
oberonia78: Good catch. I just moved it to `utils`.

hfattahi:
> Currently, there is no issue in running the bridging algorithm for dithered mode but we don't see the differences with/without bridge algorithm, because there is no gap between subswaths for now. @hfattahi

Sorry I meant to try with constant PRF data. Have you tested?

hfattahi (approved):
LGTM! Based on offline discussion, the PR is tested on Quad-pol data as well.

PR #167, opened 2025-10-23
Fix GCOV polarimetric symmetrization metadata

gshiroma:
This PR fixes issues with the GCOV polarimetric symmetrization metadata. It corrects the logic for storing the H5 datasets `polarimetricSymmetrizationApplied` and `algorithms/polarimetricSymmetrization`.

Previously, the flag polarimetricSymmetrizationApplied depended solely on the runconfig field `symmetrize_cross_pol_channels`, without verifying whether both cross-polarimetric channels were present in the input file. Additionally, the check for the existence of these polarization channels was incorrect when setting the `algorithms/polarimetricSymmetrization` dataset.


hfattahi (approved):
LGTM!

python/packages/nisar/products/writers/GcovWriter.py
```diff
@@ -575,9 +575,22 @@ def populate_processing_information(self):
             f'{parameters_group}/shadowMaskingApplied',
             False)
 
-        self.copy_from_runconfig(
+        # Add geocoding algorithm reference
+        flag_symmetrized_runconfig = self.cfg['processing']['input_subset'][
+            'symmetrize_cross_pol_channels']
+
+        flag_has_hv_and_vh = False
+        for _, pol_list in self.input_freq_pols_dict.items():
```
Tyler-g-hudson:
```suggestion
        for pol_list in self.input_freq_pols_dict.values():
```
Since you're not using the keys, you can just use the `.values()` method of `dict` to get only the set of values pointed to by all keys in the dictionary.

python/packages/nisar/products/writers/GcovWriter.py
```diff
@@ -575,9 +575,22 @@ def populate_processing_information(self):
...
+            if 'HV' in pol_list and 'VH' in pol_list:
+                flag_has_hv_and_vh = True
+                break
```
Tyler-g-hudson:
```suggestion
        flag_has_hv_and_vh = any(
            "HV" in pol_list and "VH" in pol_list
            for pol_list in self.input_freq_pols_dict.values()
        )
```
Python list comprehension gives us a faster and (imo) more quickly readable way to write this logic

Tyler-g-hudson (approved):
LGTM, couple of nitpicks

python/packages/nisar/products/writers/GcovWriter.py
gshiroma: Good suggestion. Thank you, @Tyler-g-hudson !

PR #172, opened 2025-10-29
Fix interferometry baseline

oberonia78:
This PR addresses the issue in the baseline. 

From the real NISAR interferometric pair, we found that the estimated baseline was higher than the expected. (expected : 117m, estimated :~ 2 km). The PR introduces the along-track baseline and compensate it while computing the perp and parallel baselines. 

From this PR, the estimated baseline is close to the expected and confirmed with real NISAR data. 
The PR was copied from internal [PR](https://github-fn.jpl.nasa.gov/isce-3/isce/pull/1930) with some modifications.

### DIFF
--- a/python/packages/nisar/workflows/baseline.py
+++ b/python/packages/nisar/workflows/baseline.py
@@ -350,78 +350,78 @@ def compute_baseline(ref_rngs,
         A component of the baseline perpendicular to the los vector
         from the reference sensor position to the target.
     """
-
+    info_channel = journal.info("baseline.compute_baseline")
     proj = isce3.core.make_projection(epsg_code)
-    meta_rows, meta_cols = ref_rngs.shape
 
     # Initialize output arrays
-    par_baseline_array = np.full((meta_rows, meta_cols),
-                                 np.nan, dtype=np.float32)
-    perp_baseline_array = np.full((meta_rows, meta_cols),
-                                  np.nan, dtype=np.float32)
+    par_baseline_array = np.full(
+        ref_rngs.shape, np.nan, dtype=np.float32)
+    perp_baseline_array = np.full(
+        ref_rngs.shape, np.nan, dtype=np.float32)
 
-    for row_ind in range(meta_rows):
-        for col_ind in range(meta_cols):
-            ref_azt = ref_azts[row_ind, col_ind]
-            ref_rng = ref_rngs[row_ind, col_ind]
-            sec_azt = sec_azts[row_ind, col_ind]
-            sec_rng = sec_rngs[row_ind, col_ind]
+    for row_ind, col_ind in np.ndindex(ref_rngs.shape):
+        ref_azt = ref_azts[row_ind, col_ind]
+        ref_rng = ref_rngs[row_ind, col_ind]
+        sec_azt = sec_azts[row_ind, col_ind]
+        sec_rng = sec_rngs[row_ind, col_ind]
 
-            x = coord_set[0, row_ind, col_ind]
-            y = coord_set[1, row_ind, col_ind]
-            h = coord_set[2, row_ind, col_ind]
+        x = coord_set[0, row_ind, col_ind]
+        y = coord_set[1, row_ind, col_ind]
+        h = coord_set[2, row_ind, col_ind]
 
-            lon, lat, h = proj.inverse(np.array([x, y, h]))
-            target_xyz = ellipsoid.lon_lat_to_xyz(
-                np.array([lon, lat, h]))
+        lon, lat, h = proj.inverse(np.array([x, y, h]))
+        target_xyz = ellipsoid.lon_lat_to_xyz(
+            np.array([lon, lat, h]))
 
-            if not np.isnan(ref_azt):
-                ref_xyz, ref_vel = ref_orbit.interpolate(ref_azt)
-                # get the sensor position at the sec_aztime
-                # on the secondary orbit
-                sec_xyz, _ = sec_orbit.interpolate(sec_azt)
-    
-                los_vec = target_xyz - ref_xyz
-                los_norm = np.linalg.norm(los_vec)
-                if los_norm == 0 or not np.isfinite(los_norm):
-                    continue
-                los_unit_vec = los_vec / los_norm
-                baseline_vec = sec_xyz - ref_xyz
+        if not np.isnan(ref_azt):
+            ref_xyz, ref_vel = ref_orbit.interpolate(ref_azt)
+            # get the sensor position at the sec_aztime
+            # on the secondary orbit
+            sec_xyz, _ = sec_orbit.interpolate(sec_azt)
 
-                # Along track components
-                vnorm = np.linalg.norm(ref_vel)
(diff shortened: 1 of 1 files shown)


python/packages/nisar/workflows/baseline.py
```diff
@@ -355,57 +355,65 @@ def compute_baseline(ref_rngs,
...
     # Initialize output arrays
-    perp_baseline_array = np.zeros([meta_rows, meta_cols])
-    par_baseline_array = np.zeros([meta_rows, meta_cols])
+    par_baseline_array = np.full((meta_rows, meta_cols),
+                                 np.nan, dtype=np.float32)
+    perp_baseline_array = np.full((meta_rows, meta_cols),
+                                  np.nan, dtype=np.float32)
 
     for row_ind in range(meta_rows):
         for col_ind in range(meta_cols):
```
Tyler-g-hudson:
```suggestion
    # Initialize output arrays
    par_baseline_array = np.full(
        ref_rngs.shape, np.nan, dtype=np.float32
    )
    perp_baseline_array = np.full(
        ref_rngs.shape, np.nan, dtype=np.float32
    )

    for row_ind, col_ind in np.ndindex(ref_rngs.shape):
```

I recently learned that nested for loops over arrays like this can be flattened in this manner. Saves some bookkeeping and line space, and keeps it clear at a glance that all of these structures are the same dimensions as `ref_rngs`. Optional for this PR but something useful for the future!

python/packages/nisar/workflows/baseline.py
```diff
@@ -355,57 +355,65 @@ def compute_baseline(ref_rngs,
...
+
+                # Along track components
+                vnorm = np.linalg.norm(ref_vel)
+                if vnorm > 0 and np.isfinite(vnorm):
+                    vhat = ref_vel / vnorm
+                    baseline_along = float(np.dot(baseline_vec, vhat))
+                    baseline_vec_comp = baseline_vec - baseline_along * vhat
+                else:
+                    # Fallback if velocity is degenerate
+                    baseline_vec_comp = baseline_vec
```
Tyler-g-hudson: What are the possible consequences of this fallback? How will users or debuggers know if a compensated baseline has been used vs. when one has not been used?
Do we have any means of reporting if velocity is degenerate for some portion of an image? Otherwise I feel like we could end up quietly including artifacts here where sections of an orbit are producing a compensated baseline and parts of it are quietly using baseline_vec.

Tyler-g-hudson (commented):
Some questions regarding the implications of this policy for falling back to `baseline_vec` as well as a recommendation for iterating over flattened for loops.

python/packages/nisar/workflows/baseline.py
```diff
@@ -355,57 +355,65 @@ def compute_baseline(ref_rngs,
...
             if not np.isnan(ref_azt):
                 ref_xyz, ref_vel = ref_orbit.interpolate(ref_azt)
                 # get the sensor position at the sec_aztime
                 # on the secondary orbit
                 sec_xyz, _ = sec_orbit.interpolate(sec_azt)
+    
+                los_vec = target_xyz - ref_xyz
+                los_norm = np.linalg.norm(los_vec)
+                if los_norm == 0 or not np.isfinite(los_norm):
+                    continue
```
xhuang-jpl: does that mean if the los is 0, both parallel and perpendicular baseline are NaNs? should not be 0?

python/packages/nisar/workflows/baseline.py
```diff
@@ -355,57 +355,65 @@ def compute_baseline(ref_rngs,
...
+                los_unit_vec = los_vec / los_norm
+                baseline_vec = sec_xyz - ref_xyz
+
+                # Along track components
+                vnorm = np.linalg.norm(ref_vel)
+                if vnorm > 0 and np.isfinite(vnorm):
```
xhuang-jpl: isn't the vnorm always positive since it is L2 norm?

python/packages/nisar/workflows/baseline.py
```diff
@@ -355,57 +355,65 @@ def compute_baseline(ref_rngs,
...
+                    vhat = ref_vel / vnorm
+                    baseline_along = float(np.dot(baseline_vec, vhat))
```
xhuang-jpl: why float() here? should not it be always float data type?

python/packages/nisar/workflows/baseline.py
```diff
@@ -355,57 +355,65 @@ def compute_baseline(ref_rngs,
...
-                # positive perpendicular baseline is defined
-                # at below to LOS vector
+                # Parallel (LOS) component using compensated baseline
+                parallel_baseline = float(np.dot(baseline_vec_comp,
+                                                 los_unit_vec))
+
+                # Perpendicular component magnitude using compensated baseline
+                perp_baseline_temp = float(
+                    np.linalg.norm(
+                        baseline_vec_comp - parallel_baseline * los_unit_vec))
```
xhuang-jpl: same here, the float() function might not be necessary

xhuang-jpl (commented):
Thank you @jungkyoJung, few questions about the computation.

python/packages/nisar/workflows/baseline.py
oberonia78: Yes right. So, this case is only for the `ref_vel` is not computed or filled with `nan` or `inf`. This case will not happen in actual situations.

python/packages/nisar/workflows/baseline.py
oberonia78: when los is zero, it means we cannot compute the baseline due to the several reasons. For example, if the orbit is out of the metadata field or geogrid of the metadata is not fully covered and geo2rdr failed, then `target_xyz` and `ref_xyz` could be 0 or invalid value. Then, baseline will have the NaN value

python/packages/nisar/workflows/baseline.py
oberonia78: Good point. Thanks @Tyler-g-hudson I changed the lines as you suggested.

python/packages/nisar/workflows/baseline.py
oberonia78: I added the `journal.info` to inform that the velocity is degenerate. However, this will be very rare, I guess.

python/packages/nisar/workflows/baseline.py
oberonia78: Yes. I'll delete float()

python/packages/nisar/workflows/baseline.py
oberonia78: Same here. I'll delete float().

oberonia78 (commented):
@xhuang-jpl @Tyler-g-hudson Thanks for the review the PR. I just answer your comments and change the code as you requested. Thanks again.

xhuang-jpl (approved):
Thank you @jungkyoJung for the clarification. LGMT.

Tyler-g-hudson (approved):
LGTM!

PR #65, opened 2025-07-21
Remove the InSAR workflow intermediate files 

xhuang-jpl:
This PR is to remove the InSAR workflow intermediate files under the scratch folder. The outputs and dependency of each module can be found [here](https://docs.google.com/spreadsheets/d/1zJoyuhzUGND3Qdnfxwp3weBdq-psHjWQHBw1d2o93Kg/edit?usp=drive_link).

PR #173, opened 2025-10-30
Add Polyfitting to compute the range and azimuth offsets for the fine resampling

xhuang-jpl:
the old version rubbersheet algorithm is to interpolate the filtered dense offsets to compute the range and azimuth offsets for the fine resampling, and this PR is to add an additional polyfitting algorithm to compute the range and azimuth offsets for the fine resampling.  To help you review this PR, the PPT is attached for the theory behind.

[Polynominal_Fitting.pptx](https://github.com/user-attachments/files/23247017/Polynominal_Fitting.pptx)


python/packages/nisar/workflows/rubbersheet.py
```diff
@@ -22,9 +24,193 @@
...
+                    prf=prf, rbw=rbw,
+                    abw=abw, rsr=rsr,)
+
+                print(f"Polyfitting Degree: {results["degree"]}")
+                print(f"Polyfitting CoefL: {results["coefL"]}")
+                print(f"Polyfitting CoefP: {results["coefP"]}")
+
+                info_channel.log(f"Polyfitting Degree: {results["degree"]}")
+                info_channel.log(f"Polyfitting CoefL: {results["coefL"]}")
+                info_channel.log(f"Polyfitting CoefP: {results["coefP"]}")
```
Tyler-g-hudson:
```suggestion
                polyfitting_degree = results["degree"]
                polyfitting_coefL = results["coefL"]
                polyfitting_coefP = results["coefP"]

                print(f"Polyfitting Degree: {polyfitting_degree}")
                print(f"Polyfitting CoefL: {polyfitting_coefL}")
                print(f"Polyfitting CoefP: {polyfitting_coefP}")

                info_channel.log(f"Polyfitting Degree: {polyfitting_degree}")
                info_channel.log(f"Polyfitting CoefL: {polyfitting_coefL}")
                info_channel.log(f"Polyfitting CoefP: {polyfitting_coefP}")
```
Our Python Minimum Requirements Github Actions workflow fails due to how these strings are structured, presumably because you've embedded a second set of quotes into the format strings and older python strings don't like that.

Tyler-g-hudson (commented):
The following tests fail on the minimum requirements build for this PR:
```
 The following tests FAILED:
	214 - test.python.pkg.nisar.workflows.insar (Failed)
	216 - test.python.pkg.nisar.workflows.bandpass_insar (Failed)
Errors while running CTest
	217 - test.python.pkg.nisar.workflows.ionosphere (Failed)
	220 - test.python.pkg.nisar.workflows.baseline (Failed)
```

The test `test_add_baseline` in `tests/python/packages/nisar/workflows/baseline.py:273` fails on the minimum requirements build with the following error:

```
 >                   baseline.add_baseline(
                        output_paths_gunw,
                        unit_test_params.ref_orbit,
                        unit_test_params.ref_orbit,
                        unit_test_params.ref_radargrid,
                        unit_test_params.ref_radargrid,
                        unit_test_params.ref_doppler,
                        unit_test_params.ref_doppler,
                        metadata_path_dict,
                        unit_test_params.geo2rdr_parameters,
                        use_gpu=False,
                        baseline_dir_path=unit_test_params.baseline_dir_path,
                        baseline_mode='top_bottom')

../../../../../../tests/python/packages/nisar/workflows/baseline.py:273: 
../../../../../../install/packages/nisar/workflows/baseline.py:542: in add_baseline
    dst_h5[metadata_path_dict["projection"]].attrs['epsg_code']
h5py/_objects.pyx:54: in h5py._objects.with_phil.wrapper
    ???
h5py/_objects.pyx:55: in h5py._objects.with_phil.wrapper
    ???
/home/runner/micromamba/envs/isce3/lib/python3.9/site-packages/h5py/_hl/group.py:360: in __getitem__
    oid = h5o.open(self.id, self._e(name), lapl=self._lapl)
h5py/_objects.pyx:54: in h5py._objects.with_phil.wrapper
    ???
h5py/_objects.pyx:55: in h5py._objects.with_phil.wrapper
    ???

>   ???
E   KeyError: "Unable to synchronously open object (object 'projection' doesn't exist)"

h5py/h5o.pyx:257: KeyError
```

I've addressed a number of the other errors due to a string formatting issue - please fix these test failures and ensure that the minimum requirements run succeeds

xhuang-jpl:
> The following tests fail on the minimum requirements build for this PR:
> 
> ```
>  The following tests FAILED:
> 	214 - test.python.pkg.nisar.workflows.insar (Failed)
> 	216 - test.python.pkg.nisar.workflows.bandpass_insar (Failed)
> Errors while running CTest
> 	217 - test.python.pkg.nisar.workflows.ionosphere (Failed)
> 	220 - test.python.pkg.nisar.workflows.baseline (Failed)
> ```
> 
> The test `test_add_baseline` in `tests/python/packages/nisar/workflows/baseline.py:273` fails on the minimum requirements build with the following error:
> 
> ```
>  >                   baseline.add_baseline(
>                         output_paths_gunw,
>                         unit_test_params.ref_orbit,
>                         unit_test_params.ref_orbit,
>                         unit_test_params.ref_radargrid,
>                         unit_test_params.ref_radargrid,
>                         unit_test_params.ref_doppler,
>                         unit_test_params.ref_doppler,
>                         metadata_path_dict,
>                         unit_test_params.geo2rdr_parameters,
>                         use_gpu=False,
>                         baseline_dir_path=unit_test_params.baseline_dir_path,
>                         baseline_mode='top_bottom')
> 
> ../../../../../../tests/python/packages/nisar/workflows/baseline.py:273: 
> _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
> ../../../../../../install/packages/nisar/workflows/baseline.py:542: in add_baseline
>     dst_h5[metadata_path_dict["projection"]].attrs['epsg_code']
> h5py/_objects.pyx:54: in h5py._objects.with_phil.wrapper
>     ???
> h5py/_objects.pyx:55: in h5py._objects.with_phil.wrapper
>     ???
> /home/runner/micromamba/envs/isce3/lib/python3.9/site-packages/h5py/_hl/group.py:360: in __getitem__
>     oid = h5o.open(self.id, self._e(name), lapl=self._lapl)
> h5py/_objects.pyx:54: in h5py._objects.with_phil.wrapper
>     ???
> h5py/_objects.pyx:55: in h5py._objects.with_phil.wrapper
>     ???
> _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
> 
> >   ???
> E   KeyError: "Unable to synchronously open object (object 'projection' doesn't exist)"
> 
> h5py/h5o.pyx:257: KeyError
> ```
> 
> I've addressed a number of the other errors due to a string formatting issue - please fix these test failures and ensure that the minimum requirements run succeeds

Thank you @Tyler-g-hudson , it passed all unit tests.

python/packages/nisar/workflows/rubbersheet.py
```diff
@@ -22,9 +24,197 @@
...
+        for freq, _, pol_list in get_cfg_freq_pols(cfg):
+            # Get the slant range and zero doppler time spacing
+            ref_radar_grid = ref_slc.getRadarGrid(freq)
+            ref_slant_range_spacing = ref_radar_grid.range_pixel_spacing
+            ref_zero_doppler_time_spacing = ref_radar_grid.az_time_interval
+
+            prf = ref_radar_grid.prf
+            rsr = isce3.core.speed_of_light/ref_slant_range_spacing * 0.5
+            rbw = ref_slc.getSwathMetadata(freq).processed_range_bandwidth
+            abw = ref_slc.getSwathMetadata(freq).processed_azimuth_bandwidth
```
Tyler-g-hudson: In general I'm not a fan of truncated variable names like this - in the interest of self-documenting code it's best to make these human-readable e.g. `range_bandwidth`

python/packages/nisar/workflows/rubbersheet.py
```diff
@@ -22,9 +24,197 @@
...
+                              ref_radar_grid.prf).astype(int)
+            off_rg_indices = ((off_slant_range - ref_slant_ranges[0])/
+                              ref_slant_range_spacing).astype(int)
+
+            # Produce ground track velocity for the frequency under processing
+            ground_track_velocity_file = get_ground_track_velocity_product(ref_slc,
+                                                                           off_slant_range,
+                                                                           off_zero_doppler_time,
+                                                                           dem_file,
+                                                                           rubbersheet_dir)
```
Tyler-g-hudson:
```suggestion
            ground_track_velocity_file = get_ground_track_velocity_product(
                ref_slc,
                off_slant_range,
                off_zero_doppler_time,
                dem_file,
                rubbersheet_dir,
            )
```
This is a war that I will probably never win, but these parameter lists that are way up to the right are more difficult to read. Also, it's best to use keyword arguments where possible so that if the order of function parameters changes in the function being called, the code is less likely to break or, if it does, it will at least give a more appropriate error.

python/packages/nisar/workflows/rubbersheet.py
```diff
@@ -22,9 +24,197 @@
...
+                lines, pixels = np.meshgrid(off_az_indices, off_rg_indices, indexing="ij")
+                culled_az_offsets, culled_rg_offsets =\
+                    offsets_polyfit.predict_offsets_batch(lines, pixels,
+                                                          results["coefL"],
+                                                          results["coefP"],
+                                                          results["degree"],
+                                                          minL, maxL, minP, maxP)
+
+                _write_to_disk(str(f'{out_dir}/culled_az_offsets'),culled_az_offsets)
+                _write_to_disk(str(f'{out_dir}/culled_rg_offsets'),culled_rg_offsets)
```
Tyler-g-hudson:
```suggestion
                _write_to_disk(str(f'{out_dir}/culled_az_offsets'), culled_az_offsets)
                _write_to_disk(str(f'{out_dir}/culled_rg_offsets'), culled_rg_offsets)
```

python/packages/nisar/workflows/rubbersheet.py
```diff
@@ -22,9 +24,197 @@
...
+                offset_az_prod = dst_h5[f'{pol_group_path}/alongTrackOffset']
+                offset_rg_prod = dst_h5[f'{pol_group_path}/slantRangeOffset']
+                offset_peak_prod = dst_h5[f'{pol_group_path}/correlationSurfacePeak']
+
+                # Assign cross-correlation peak
+                offset_peak_prod[...] = _open_raster(corr_peak_path)
+                # Convert the along track and slant range pixel offsets to meters
+                offset_az_prod[...] = \
+                    culled_az_offsets * ground_track_velocity \
+                    * ref_zero_doppler_time_spacing
```
Tyler-g-hudson:
```suggestion
                offset_az_prod[...] = (
                    culled_az_offsets
                    * ground_track_velocity
                    * ref_zero_doppler_time_spacing
                )
```
Multi-line formulae can be done like this, saves you the effort of putting a bunch of back slashes and is a little easier to follow at a glance

python/packages/isce3/math/offsets_polyfit.py
```diff
@@ -0,0 +1,508 @@
...
+
+    Returns
+    -------
+    int
+        The number of unique coefficients in the polynomial.
+    """
+    return (degree + 1) * (degree + 2) // 2
+
+
+def normalize(x, xmin, xmax):
```
bhawkins:
```suggestion
def normalize_to_unit_range(x, xmin=None, xmax=None):
```
This name is too generic for the `isce3.math` namespace. Also it would be more useful if the function could compute the extrema itself, though I can see why you might want the option not to.

python/packages/isce3/math/offsets_polyfit.py
```diff
@@ -0,0 +1,508 @@
...
+
+def polyfit_offsets(
+    data,degree=2,
+    crit_value=0.1,
+    max_iterations=50,
+    minL=None, maxL=None,
+    minP=None, maxP=None,
+    prf=None, abw=None,
+    rsr=None, rbw=None):
+
```
Tyler-g-hudson, line 200:
```suggestion
```

python/packages/isce3/math/offsets_polyfit.py
```diff
@@ -0,0 +1,508 @@
...
+    minL, maxL : float, optional
+        Optional bounds on acceptable *line* offsets (units of `dL`).
+        Points with `dL` outside ``[minL, maxL]`` are removed before fitting.
+    minP, maxP : float, optional
+        Optional bounds on acceptable *pixel* offsets (units of `dP`).
+        Points with `dP` outside ``[minP, maxP]`` are removed before fitting.
+    prf, abw, rsr, rbw : float, optional
+        SAR system parameters (pulse repetition frequency, azimuth bandwidth,
+        range sampling rate, range bandwidth) used, if provided, to derive
+        prior standard deviations for weighting the line/pixel residuals.
```
Tyler-g-hudson, line 236:
```suggestion
    prf, abw : float, optional
        SAR pulse repetition frequency and azimuth bandwidth, used if provided, to derive
        prior standard deviations for weighting the line residuals. If either is not given,
        a default prior standard deviation of 0.15/1.1 will be used for line residuals. 
    rsr, rbw : float, optional 
        SAR system range sampling rate and range bandwidth used, if provided, to derive
        prior standard deviations for weighting the pixel residuals. If either is not given,
        a default prior standard deviation of 0.1/1.1 will be used for pixel residuals. 
```
These pairs are used independently in the code, so should have independent entries in the docstring.

python/packages/isce3/math/offsets_polyfit.py
```diff
@@ -0,0 +1,508 @@
...
+          or by hard bounds.
+        - **degree** : int
+          Polynomial degree actually used.
+        - **design_nunk** : int
+          Number of unknown coefficients ``M``.
+    """
+
+    # Prior std for w-test
+    sigmaL = 0.15 / ((prf / abw) if ((prf is not None) and (abw is not None)) else 1.1)
+    sigmaP = 0.10 / ((rsr / rbw) if ((rsr is not None) and (rbw is not None)) else 1.1)
```
Tyler-g-hudson:
```suggestion
    sigmaL = 0.15 / ((prf / abw) if (None not in (prf, abw) else 1.1)
    sigmaP = 0.10 / ((rsr / rbw) if (None not in (rsr, rbw) else 1.1)
```
This is a shorter way of expressing the same logical expression

python/packages/isce3/math/offsets_polyfit.py
```diff
@@ -0,0 +1,508 @@
...
+
+    if None in (minL, maxL, minP, maxP):
+        minL, maxL = data[:, 1].min(), data[:, 1].max()
+        minP, maxP = data[:, 2].min(), data[:, 2].max()
```
Tyler-g-hudson: This doesn't seem like something the users and maintainers of this code are going to expect - all values given are overwritten if any value is None, currently. That should either be stated in the docstring or, preferably in my opinion, each input value should be handled independently like so:

```suggestion
    maxL = data[:, 1].max() if maxL is None else maxL
    minL = data[:, 1].min() if minL is None else minL
    maxP = data[:, 2].max() if maxP is None else maxP
    minP = data[:, 2].min() if minP is None else minP

    if minL >= maxL:
        raise ValueError(
            f"minL value {minL} greater than or equal to maxL value {maxL}. "
            "minL must be less than maxL."
        )
    if minP >= maxP:
        raise ValueError(
            f"minP value {minP} greater than or equal to maxP value {maxP}. "
            "minP must be less than maxP."
        )
```

python/packages/isce3/math/offsets_polyfit.py
```diff
@@ -0,0 +1,508 @@
...
+
+    Nunk, eps = ncoeffs(degree), 1e-8
+    removed_indices, iteration = [], 0
```
Tyler-g-hudson:
```suggestion
    num_coeffs = ncoeffs(degree)
    eps = 1e-8
    removed_indices = []
    iteration = 0
```
Are these variables related somehow? They're being assigned on the same line but don't seem to be similar values.

Also, there are a lot of abbreviated variable names e.g. `Nunk` and `eps` being thrown around here that make this code take longer to interpret, review, and maintain. You could either add comments everywhere that these short variables are assigned to explain what they are, or (preferably) give them a name that plainly says what they are, wherever possible.

python/packages/isce3/math/offsets_polyfit.py
```diff
@@ -0,0 +1,508 @@
...
+
+    Parameters
+    ----------
+    data : (N, 6) array_like
+        Input tie-point array where each row is:
+        ``[id, line, pixel, dL, dP, corr_peak]``.
+        - *id* : int-like identifier.
+        - *line, pixel* : coordinates of the tie point.
+        - *dL, dP* : measured offsets in line and pixel directions.
+        - *corr_peak* : correlation peak or quality metric.
```
Tyler-g-hudson: Why not just use six equal-length vectors?

python/packages/isce3/math/offsets_polyfit.py
```diff
@@ -0,0 +1,508 @@
...
+    sigmaP = 0.10 / ((rsr / rbw) if ((rsr is not None) and (rbw is not None)) else 1.1)
+
+    if None in (minL, maxL, minP, maxP):
+        minL, maxL = data[:, 1].min(), data[:, 1].max()
+        minP, maxP = data[:, 2].min(), data[:, 2].max()
+
+    Nunk, eps = ncoeffs(degree), 1e-8
+    removed_indices, iteration = [], 0
+
+    while True:
```
Tyler-g-hudson:
```suggestion
    for iteration in itertools.count():
```
You can kill the need for hand-incrementing your loop counter and maintain an infinite loop by using `itertools.count()`

python/packages/isce3/math/offsets_polyfit.py
```diff
@@ -0,0 +1,508 @@
...
+        A[i, :] = [l^(j-k) * p^k]  for all  0 <= j <= degree,  0 <= k <= j
+
+    where `l` and `p` are the normalized line and pixel coordinates.
+
+    Parameters
+    ----------
+    lines_n : array_like
+        Normalized line coordinates (independent variable 1).
+    pixels_n : array_like
+        Normalized pixel coordinates (independent variable 2).
```
bhawkins:
```suggestion
        Normalized pixel coordinates (independent variable 2).
        Same length as `lines_n`.
```

python/packages/isce3/math/offsets_polyfit.py
```diff
@@ -0,0 +1,508 @@
...
+
+    Returns
+    -------
+    A : ndarray of shape (N, M)
+        The design matrix, where `N` is the number of observations and
+        `M` is the number of polynomial coefficients, given by:
+
+        M = (degree + 1) * (degree + 2) / 2
+    """
+    N = lines_n.size
```
bhawkins:
```suggestion
    N = lines_n.size
    if pixels_n.size != N:
        raise ValueError("Coordinate vectors must be the same size")
```

python/packages/nisar/workflows/rubbersheet.py
```diff
@@ -22,9 +24,197 @@
...
+                    np.arange(corr_peak.size),
+                    np.repeat(az_polyfit_indices, len(rg_polyfit_indices)),
+                    np.tile(rg_polyfit_indices, len(az_polyfit_indices)),
+                    az_offsets.ravel(), rg_offsets.ravel(),
+                    corr_peak.ravel()
+                ])
+
+                # polyfitting the range and azimuth offsets
+                results = offsets_polyfit.polyfit_offsets(
+                    data,degree=rubbersheet_params['polyfitting']['degree'],
```
Tyler-g-hudson:
```suggestion
                    data,
                    degree=rubbersheet_params['polyfitting']['degree'],
```
Original construction looked as if the value at `rubbersheet_params['polyfitting']['degree']` was being split as a tuple and assigned to the keywords `data` and `degree`

Tyler-g-hudson (commented):
Mostly code style commentary, some fixes to bugs/deviations from the behavior users might expect from the polyfitting function

python/packages/isce3/math/offsets_polyfit.py
```diff
@@ -0,0 +1,508 @@
...
+    A = np.zeros((N, ncoeffs(degree)), dtype=float)
+    idx = 0
+    for j in range(degree + 1):
+        for k in range(j + 1):
+            A[:, idx] = (lines_n ** (j - k)) * (pixels_n ** k)
+            idx += 1
```
bhawkins: This could be done more efficiently with a variant of [Horner's method](https://en.wikipedia.org/wiki/Horner%27s_method). Basically you'd recursively compute powers of x, then recursively compute powers of y, then you'd populate the cross terms with products of those results. Not a big deal for small problems, I guess.

python/packages/isce3/math/offsets_polyfit.py
```diff
@@ -0,0 +1,508 @@
...
+    where `l` and `p` are the normalized line and pixel coordinates.
+
+    Parameters
+    ----------
+    lines_n : array_like
+        Normalized line coordinates (independent variable 1).
+    pixels_n : array_like
+        Normalized pixel coordinates (independent variable 2).
+    degree : int
+        The maximum total degree of the 2D polynomial.
```
bhawkins:
```suggestion
        The maximum total degree of the 2D polynomial.
        (Same in both dimensions)
```

python/packages/isce3/math/offsets_polyfit.py
```diff
@@ -0,0 +1,508 @@
+import numpy as np
```
bhawkins, line 3: There's some similar functionality in the module

https://github.com/isce-framework/isce3/blob/develop/python/packages/isce3/core/poly2d.py

Maybe we should put some of the polyfit functions in that module, and move the offsets-specific stuff to its own module.

python/packages/isce3/math/__init__.py
```diff
@@ -1,2 +1,3 @@
 from isce3.ext.isce3.math import *
 from .snap import snap, snap_datetime
+from .offsets_polyfit import polyfit_offsets, predict_offsets_batch, build_design_matrix, ncoeffs, normalize
```
bhawkins:
```suggestion
from . import offsets_polyfit
```
I don't think we should import all these names into `isce3.math`.

python/packages/isce3/math/offsets_polyfit.py
```diff
@@ -0,0 +1,508 @@
...
+    A = np.zeros((N, ncoeffs(degree)), dtype=float)
+    idx = 0
+    for j in range(degree + 1):
+        for k in range(j + 1):
+            A[:, idx] = (lines_n ** (j - k)) * (pixels_n ** k)
+            idx += 1
+    return A
+
+
+def cholesky_solve(N_mat, rhs):
```
bhawkins, line 135: We don't need this since we already depend on scipy which has a method for it:

https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.cho_solve.html#scipy.linalg.cho_solve

python/packages/isce3/math/offsets_polyfit.py
```diff
@@ -0,0 +1,508 @@
...
+    -------
+    x : ndarray
+        Solution vector or matrix satisfying `N_mat @ x = rhs`.
+    L : ndarray
+        Lower-triangular Cholesky factor of `N_mat`.
+    """
+
+    L = np.linalg.cholesky(N_mat)
+    y = np.linalg.solve(L, rhs)
+    x = np.linalg.solve(L.T, y)
```
bhawkins: The numpy solver doesn't claim to check for any structure in the matrix, so IMO we should use scipy routines since we already depend on scipy.
```suggestion
    L, is_lower = scipy.linalg.cholesky(N_mat, lower=True)
    x = scipy.linalg.cho_solve((L, is_lower), rhs)
```

python/packages/isce3/math/offsets_polyfit.py
```diff
@@ -0,0 +1,508 @@
...
+            raise ValueError("No sufficient points for the rubbersheet polyfitting")
+
+        # Weights
+        w = np.clip(data[:, 5], eps, 1.0)
+        Qy_diag = 1.0 / (w * w)
+        W = w[:, None]
+
+        # Design
+        Ln, Pn = normalize(lines, minL, maxL), normalize(pixels, minP, maxP)
+        A = build_design_matrix(Ln, Pn, degree)
```
bhawkins: It seems wasteful to recompute this every iteration. We could just compute it once and delete rows as we go, just like with `data`.

python/packages/isce3/math/offsets_polyfit.py
```diff
@@ -0,0 +1,508 @@
...
+    -------
+    A_inv : (M, M) ndarray
+        Inverse of the original matrix `A`.
+    """
+
+    Linv = np.linalg.inv(L)
+    return Linv.T @ Linv
+
+def polyfit_offsets(
+    data,degree=2,
```
Tyler-g-hudson, line 193:
```suggestion
    data,
    degree=2,
```

python/packages/isce3/math/offsets_polyfit.py
bhawkins:
```suggestion
        - *corr_peak* : correlation peak or quality metric in interval (0, 1].
        Measurements with larger values get more weight.
```
A bigger number means the measurement should get more weight, right? That seems to be the behavior below. It's not necessarily intuitive since often one would supply measurement variances (smaller is better).

python/packages/isce3/math/offsets_polyfit.py
```diff
@@ -0,0 +1,508 @@
...
+
+    # Prior std for w-test
+    sigmaL = 0.15 / ((prf / abw) if ((prf is not None) and (abw is not None)) else 1.1)
+    sigmaP = 0.10 / ((rsr / rbw) if ((rsr is not None) and (rbw is not None)) else 1.1)
+
+    if None in (minL, maxL, minP, maxP):
+        minL, maxL = data[:, 1].min(), data[:, 1].max()
+        minP, maxP = data[:, 2].min(), data[:, 2].max()
+
+    Nunk, eps = ncoeffs(degree), 1e-8
```
bhawkins:
```suggestion
    Nunk = ncoeffs(degree)
    eps = np.sqrt(np.finfo(float).eps)
```
I think 1e-8 is too small for your intended usage.

python/packages/isce3/math/offsets_polyfit.py
xhuang-jpl: good suggestion, I think it is fine that we use six equal length vectors, but we might need to check if their lengths are equal. Since the docstring is clear enough and they are equivalent, I will keep it as is.

python/packages/isce3/math/offsets_polyfit.py
```diff
@@ -0,0 +1,508 @@
...
+    """
+
+    l = 0.0 if maxL == minL else (line - minL) / (maxL - minL)
+    p = 0.0 if maxP == minP else (pixel - minP) / (maxP - minP)
+
+    row = []
+    for j in range(degree + 1):
+        for k in range(j + 1):
+            row.append((l ** (j - k)) * (p ** k))
+    return np.array(row, dtype=float)
```
bhawkins:
```suggestion
    return build_design_matrix(np.array(l), np.array(p), degree)[0]
```
Seems better not to duplicate this fiddly stuff.

python/packages/isce3/math/offsets_polyfit.py
xhuang-jpl: good suggestion!

python/packages/isce3/math/offsets_polyfit.py
```diff
@@ -0,0 +1,508 @@
...
+    l = np.ravel(l)
+    p = np.ravel(p)
+
+    Aidx, Bidx = _expo_pairs(degree)
+    ar = np.arange(degree + 1, dtype=np.int64)
+
+    l_pows = np.power(l[:, None], ar[None, :])
+    p_pows = np.power(p[:, None], ar[None, :])
+
+    return l_pows.take(Aidx, axis=1) * p_pows.take(Bidx, axis=1)
```
bhawkins: Ah this is sort of like what I was suggesting in my earlier comment on `build_design_matrix`. I don't see why we need both methods. IMO we should delete one (or implement one in terms of the other if you want the normalization to be optional).

tests/python/packages/isce3/math/offsets_polyfit.py
```diff
@@ -0,0 +1,114 @@
+
```
bhawkins, line 1: The name of this file should match the name of the module it's testing. This is especially confusing given the similar functions in isce3.core.polyfit2d.

python/packages/isce3/math/offsets_polyfit.py
```diff
@@ -0,0 +1,508 @@
...
+        Optional bounds on acceptable *pixel* offsets (units of `dP`).
+        Points with `dP` outside ``[minP, maxP]`` are removed before fitting.
+    prf, abw, rsr, rbw : float, optional
+        SAR system parameters (pulse repetition frequency, azimuth bandwidth,
+        range sampling rate, range bandwidth) used, if provided, to derive
+        prior standard deviations for weighting the line/pixel residuals.
+
+    Returns
+    -------
+    result : dict
```
bhawkins, line 240: A [namedtuple](https://docs.python.org/3/library/collections.html#collections.namedtuple) would be more convenient IMO.

python/packages/isce3/math/offsets_polyfit.py
```diff
@@ -0,0 +1,508 @@
...
+        W = w[:, None]
+
+        # Design
+        Ln, Pn = normalize(lines, minL, maxL), normalize(pixels, minP, maxP)
+        A = build_design_matrix(Ln, Pn, degree)
+
+        # Weighted normal equations
+        A_til = A * W
+        At = A_til.T
+        Nmat = At @ A_til
```
bhawkins, line 304: I'm not sure this is the most accurate way to solve the problem since squaring the design matrix could lose precision. It might be better to do `scipy.linalg.solve(A * W, rhs * W)`, though it may do something similar. Since your tests work okay, perhaps this isn't a big issue.

share/nisar/defaults/insar.yaml
```diff
@@ -89,6 +89,9 @@ runconfig:
 
         # TODO OPTIONAL - To setup type of worker
         worker:
+             # OPTIONAL - To remove the intermediate files. Default False
+            intermediate_files_removal_enabled: False
```
nemo794: Are you sure we want to keep the intermediate files? I thought there was discussion awhile back about needing to delete the intermediate files for disk space(?) reasons, but I could be misremembering.

python/packages/nisar/workflows/rubbersheet.py
```diff
@@ -22,9 +24,197 @@
...
+            pixel_offsets_path = f'{freq_group_path}/pixelOffsets'
+            geo_offset_dir = geo2rdr_offsets_path / 'geo2rdr' / f'freq{freq}'
+            rubbersheet_dir = scratch_path / 'rubbersheet_offsets' / f'freq{freq}'
+            off_slant_range = dst_h5[f'{pixel_offsets_path}/slantRange'][()]
+            off_zero_doppler_time = dst_h5[f'{pixel_offsets_path}/zeroDopplerTime'][()]
+
+            off_az_indices = ((off_zero_doppler_time - ref_az_times[0]) *
+                              ref_radar_grid.prf).astype(int)
+            off_rg_indices = ((off_slant_range - ref_slant_ranges[0])/
+                              ref_slant_range_spacing).astype(int)
```
bhawkins: It probably doesn't matter much given large images and small offsets, but some of the details of this aren't obvious to me. Do we want the surface as a function of the reference image domain or the secondary image domain? Why `floor` instead of `round` here?

python/packages/nisar/workflows/rubbersheet.py
```diff
@@ -22,9 +24,197 @@
...
+                                                 len(off_az_indices)-1,
+                                                 rubbersheet_params['polyfitting']['samples_along_azimuth'],
+                                                 dtype=int)
+                rg_sampled_indices = np.linspace(0,
+                                                 len(off_rg_indices)-1,
+                                                 rubbersheet_params['polyfitting']['samples_along_range'],
+                                                 dtype=int)
+
+                az_polyfit_indices = off_az_indices[az_sampled_indices]
+                rg_polyfit_indices = off_rg_indices[rg_sampled_indices]
```
bhawkins, line 144: As discussed in the ADT meeting today, I don't see why we should need to subsample the data. Let's just use it all if we can.

python/packages/nisar/workflows/insar.py
```diff
@@ -34,6 +39,12 @@ def run(cfg: dict, out_paths: dict, run_steps: dict):
     if run_steps['geo2rdr']:
         geo2rdr.run(cfg)
 
+    # Remove the rdr2geo scratch folder
```
bhawkins: Does all this cleanup stuff have anything to do with polyfitting?

share/nisar/defaults/insar.yaml
```diff
@@ -547,6 +550,18 @@ runconfig:
...
+                    # Flag to enable/disable polyfitting
+                    enabled: false
+                    # The degrees for the polyfitting
+                    degree: 2
+                    # Critical value for the w-test to remove the outliers
+                    critical_value: 0.1
+                    # Number of samples along the range in pixels for polyfitting
+                    samples_along_range: 50
+                    # Number of samples  along the azimuth in pixels for polyfitting
+                    samples_along_azimuth: 50
```
bhawkins:
```suggestion
```

share/nisar/schemas/insar.yaml
```diff
@@ -561,6 +562,17 @@ gaussian_filter_options:
...
+polyfitting_options:
+    enabled: bool(required=False)
+    # The degrees for the polyfitting
+    degree: num(min=1, required=False)
+    # Critical value for the w-test to remove the outliers
+    critical_value: num(min=0, required=False)
+    # Number of samples along the range in pixels for polyfitting
+    samples_along_range: num(min=0, required=False)
+    # Number of samples  along the azimuth in pixels for polyfitting
+    samples_along_azimuth: num(min=0, required=False)
```
bhawkins, line 574:
```suggestion
```

bhawkins (commented):
Thanks, this is a welcome feature. I think the only issue I really consider a blocker is subsampling the offset data. We work hard to compute all those offsets, so we should use them!

share/nisar/defaults/insar.yaml
xhuang-jpl: sorry for the misleading @nemo794, it is related to this PR #65 , I add them to this PR to help me to test them on the ISCALE where there are space issues, I think I will remove those from this PR after the testing.

Tyler-g-hudson (approved):
Brian's comments are good, once you've implemented and tested running with all offsets I think we're good to go. I'll mark an approval here so I'm not obstructing the merge when it's ready, but let's make certain that Brian is ready to merge before merging regardless of any other PR approval rules.

nemo794:
> Brian's comments are good, once you've implemented and tested running with all offsets I think we're good to go. I'll mark an approval here so I'm not obstructing the merge when it's ready, but let's make certain that Brian is ready to merge before merging regardless of any other PR approval rules.

@Tyler-g-hudson , could you please track this comment https://github.com/isce-framework/isce3/pull/173#discussion_r2543993643 before this PR merges? Thanks!

Tyler-g-hudson:
> > Brian's comments are good, once you've implemented and tested running with all offsets I think we're good to go. I'll mark an approval here so I'm not obstructing the merge when it's ready, but let's make certain that Brian is ready to merge before merging regardless of any other PR approval rules.
> 
> @Tyler-g-hudson , could you please track this comment [#173 (comment)](https://github.com/isce-framework/isce3/pull/173#discussion_r2543993643) before this PR merges? Thanks!

I will do that, thanks for alerting me

python/packages/isce3/math/offsets_polyfit.py
xhuang-jpl: Thanks

python/packages/isce3/math/offsets_polyfit.py
xhuang-jpl: good suggestion, to me, those functions are designed specifically for the offset polyfitting

python/packages/isce3/math/offsets_polyfit.py
xhuang-jpl: very good suggestion

python/packages/isce3/math/offsets_polyfit.py
xhuang-jpl: very good suggestion

python/packages/isce3/math/offsets_polyfit.py
xhuang-jpl: You are right, the bigger number means more weights, and has changed the docstring to make it clear

python/packages/isce3/math/offsets_polyfit.py
xhuang-jpl: I have deleted the old `build_design_matrix ` and only keep this function

tests/python/packages/isce3/math/offsets_polyfit.py
xhuang-jpl: have changed to offsets_polyfit.py

python/packages/isce3/math/offsets_polyfit.py
xhuang-jpl: this function has been deleted

python/packages/isce3/math/offsets_polyfit.py
xhuang-jpl: Yes, it still works well based on my testing

python/packages/nisar/workflows/rubbersheet.py
xhuang-jpl: Since the two images has been coarse co-registered, the polyfitted surface is always the function of the reference image. You are right, we should use `round` here.

python/packages/nisar/workflows/insar.py
xhuang-jpl: No, this is only for the testing when I use the ISCALE where there are space issues when we test the 40+5 MHZ insar data, and I will delete those after the testing in the PR.

xhuang-jpl:
> Thanks, this is a welcome feature. I think the only issue I really consider a blocker is subsampling the offset data. We work hard to compute all those offsets, so we should use them!

Thank you @bhawkins , I have addressed most of your comments to clean the code and remove some redundant functions  (e.g. build_design_matrix), however, I think we still need to keep the subsampling parameters, the reason is that we can always give a very large number of samples so that the polyfitting will use all the entire offsets, however, if the user generates  the offsets with very fine resolution, there might be some memory and performance (i.e. runtime) issues since our algorithm check and remove the outlier one by one. For now, I updated the default number of samples to be 1000 to force the polyfitting using as many samples as we can, but I still need to test the performance (i.e. the runtime). What do you think?

bhawkins:
That's a reasonable compromise for now. In the future, I think you could change the algorithm to discard some small fraction of outliers each iteration (say `max(1, 1/2500 * n)`), which I'd expect would have similar performance to discarding one point at a time for a 50*50 grid.

bhawkins (approved):
LGTM thanks

Tyler-g-hudson:
@nemo794 I've quickly confirmed that the intermediate file removal code has been removed.
@xhuang-jpl We're good to go with merging this as far as I'm concerned.

xhuang-jpl:
Thanks @bhawkins  and @Tyler-g-hudson. 

@Tyler-g-hudson , what about the unit test failure? It seems other PRs have similar failure.

xhuang-jpl:
> That's a reasonable compromise for now. In the future, I think you could change the algorithm to discard some small fraction of outliers each iteration (say `max(1, 1/2500 * n)`), which I'd expect would have similar performance to discarding one point at a time for a 50*50 grid.

Thanks @bhawkins , I have tested few cases with different number of samples, I make the default 200 by 200 for now, which has reasonable performance in terms of runtime and quality.

xhuang-jpl:
In addition, the PR has been tested for both CPU and GPU. Thanks all

Tyler-g-hudson:
> @Tyler-g-hudson , what about the unit test failure? It seems other PRs have similar failure.

Yes, that test is unstable, we'll need to figure out what's wrong with it eventually but it has nothing to do with your PR. You're good to merge.

PR #176, opened 2025-11-04
Support new corner reflector CSV format

bhawkins:
NASA and ISRO data processing teams agreed to some updates to the corner reflector CSV file specification. In particular, we added some validity codes to distinguish LSAR and SSAR usage, and we added a "Shape" column to distinguish triangular and square trihedrals. This PR implements support for these new features while retaining backwards compatibility.

When I went to implement the formula for a square trihedral (GO model), I was surprised that the one we have for triangular trihedrals didn't look quite right. All I can figure is that somebody misunderstood what was meant by "direction cosines" in the references. We ended up with the correct RCS for the region near the boresight, but the other branch in the formula was wrong. It's especially surprising because I could've sworn this came up earlier and Geoff had fixed it, but I guess not. Anyway, here are plots of the trihedral RCS vs angle before and after the fix in this PR (lines are contours at 3 dB intervals):

rcs_models

I added unit tests that check the boresight RCS against the simple formulas, and I added a unit test that verifies that a cut through the pattern is continuous. I also manually ran `nisarqa` on an RSLC image that I'd run before and verified that I got the same abscal factor.


tests/python/packages/isce3/cal/corner_reflector.py
```diff
@@ -291,3 +293,57 @@ def test_get_crs_in_polygon():
...
+        expected_rcs = get_peak_rcs_triangular(side_length, wavelength)
+    elif shape == "square":
+        expected_rcs = get_peak_rcs_square(side_length, wavelength)
+    else:
+        assert False, "invalid CR shape in unit test"
+    assert np.isclose(pow2db(rcs), pow2db(expected_rcs), rtol=0, atol=0.01)
+
+
+# There's a branch in the RCS vs angle model, so make sure the RCS is continuous
+
```
Tyler-g-hudson, line 331:
```suggestion
```

Tyler-g-hudson (approved):
Having read over all of this, it's hard to find fault with anything - LGTM!

bhawkins:
Need to add test file description to README file

python/packages/isce3/cal/corner_reflector.py
```diff
@@ -11,11 +12,14 @@
 
 import isce3
 
+class CRShape(str, Enum):
```
rad-eng-59, line 15: FYI, if necessary, in order to make the enum case insensitive (`str.casefold()`) to the string inputs, one can update `__missing__` classmethod in `Enum` as shown [here](https://stackoverflow.com/questions/76129550/how-to-make-case-insensitive-choices-using-pythons-enum-and-fastapi)

rad-eng-59 (approved):
LGTM. Thank you for the fix, update, and comprehensive test.

hfattahi (approved):
LGTM. We reviewed in a PR review session. Thanks @bhawkins !

PR #177, opened 2025-11-05
Fix the decimating offset for frequency B interferogram

oberonia78:
**Summary**
This PR fixes an issue in decimating the offset used to generate the frequency-\ B interferogram.

**Details**
The frequency B interferogram, which is required for ionospheric estimation, was previously generated from a decimated offset computed using the RUNW slant ranges of frequencies A and B.
However, because the RUNW product is multi-looked, its slant range array is shorter than that of the RSLC. This mismatch caused errors in the decimation process, leading to incorrect ionospheric estimates.

**Fix**
Instead of using the slant ranges from the RUNW product, this PR now reads the slant ranges directly from the RSLC, ensuring proper alignment and accurate ionospheric estimation.

### DIFF
--- a/python/packages/isce3/signal/interpolate_by_range.py
+++ b/python/packages/isce3/signal/interpolate_by_range.py
@@ -59,19 +59,13 @@ def decimate_freq_a_array(
     left_missing = min(left_missing, total_missing)
     right_missing = min(right_missing, total_missing - left_missing)
 
-    if left_missing > 0:
-        pad_left = np.zeros(
-            (decimated_array.shape[0], left_missing),
-            dtype=decimated_array.dtype
+    if left_missing > 0 or right_missing > 0:
+        decimated_array = np.pad(
+            decimated_array,
+            pad_width=((0, 0), (left_missing, right_missing)),
+            mode="constant",
+            constant_values=0,
         )
-        decimated_array = np.concatenate([pad_left, decimated_array], axis=1)
-
-    if right_missing > 0:
-        pad_right = np.zeros(
-            (decimated_array.shape[0], right_missing),
-            dtype=decimated_array.dtype
-        )
-        decimated_array = np.concatenate([decimated_array, pad_right], axis=1)
 
     return decimated_array
 
--- a/python/packages/isce3/signal/interpolate_by_range.py
+++ b/python/packages/isce3/signal/interpolate_by_range.py
@@ -30,16 +30,17 @@ def decimate_freq_a_array(
     spacing_side = slant_side[1] - slant_side[0]
 
     # make sure stride is at least 1
-    resampling_scale_factor = max(1,
-                                  int(np.round(spacing_side / spacing_main)))
+    resampling_scale_factor = max(
+        1, int(np.round(spacing_side / spacing_main))
+    )
 
     n_side = len(slant_side)
 
     # slice whatever overlaps (no shifting); then pad left/right as needed
     end_excl = min(width, first_index + n_side * resampling_scale_factor)
-    decimated_array = target_runw[:,
-                                  first_index:end_excl:resampling_scale_factor]
-
+    decimated_array = target_runw[
+        :, first_index:end_excl:resampling_scale_factor
+    ]
     # how many side samples fall outside main on each side?
     # (assumes increasing slant arrays)
     left_missing = int(
--- a/python/packages/isce3/signal/interpolate_by_range.py
+++ b/python/packages/isce3/signal/interpolate_by_range.py
@@ -1,26 +1,27 @@
 import numpy as np
 
+
 def decimate_freq_a_array(
         slant_main,
         slant_side,
         target_runw):
-    """decimate target_runw of main band to have same size with side band
-    assuming slant_main and slant_side are evenly spaced
+    """Decimate target_runw of main band to have same size as side band,
+    assuming slant_main and slant_side are evenly spaced.
 
     Parameters
     ----------
     slant_main : numpy.ndarray
-        slant range array of frequency A band
+        Slant range array of frequency A band
     slant_side : numpy.ndarray
-        slant range array of frequency B band
+        Slant range array of frequency B band
     target_runw : numpy.ndarray
-        RUNW array of frequency A band
-        width of target_runw should be same with length of slant_main
+        RUNW array of frequency A band.
+        Width of target_runw should be same as length of slant_main.
 
     Returns
     -------
     decimated_array : numpy.ndarray
-        decimated RUNW array
+        Decimated RUNW array with width == len(slant_side).
     """
     _, width = target_runw.shape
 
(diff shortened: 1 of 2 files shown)


python/packages/isce3/signal/interpolate_by_range.py
```diff
@@ -1,44 +1,80 @@
...
+    right_missing = int(
+        np.ceil(
+            max(0.0, (slant_side[-1] - slant_main[-1]) / spacing_side)
+        )
+    )
+
+    # clamp in case both sides miss (very long slant_side)
+    total_missing = max(0, n_side - decimated_array.shape[1])
+    left_missing = min(left_missing, total_missing)
+    right_missing = min(right_missing, total_missing - left_missing)
```
hfattahi, line 60: The sanity checks and being extra careful here to make sure that the decimated data is not shifted and is only decimated and the extra possible margins at left and right sides are zero padded is fine for completeness. However, I'm trying to understand what is the situation where left_missing and right_missing is non zero for NISAR. Can you elaborate if there was a specific need for these paddings ?

python/packages/isce3/signal/interpolate_by_range.py
```diff
@@ -1,44 +1,80 @@
...
     _, width = target_runw.shape
 
     first_index = np.argmin(np.abs(slant_main - slant_side[0]))
     spacing_main = slant_main[1] - slant_main[0]
     spacing_side = slant_side[1] - slant_side[0]
 
-    resampling_scale_factor = int(np.round(spacing_side / spacing_main))
+    # make sure stride is at least 1
+    resampling_scale_factor = max(1,
+                                  int(np.round(spacing_side / spacing_main)))
```
Tyler-g-hudson:
```suggestion
    resampling_scale_factor = max(
        1, int(np.round(spacing_side / spacing_main))
    )
```
Just a little stylistic recommendation, I'm not a big of these heavily indented argument lists because they condense all the parameters into a little chunk way over to the right.

python/packages/isce3/signal/interpolate_by_range.py
```diff
@@ -1,44 +1,80 @@
...
+
+    n_side = len(slant_side)
 
-    x_cand = np.arange(1, width + 1)
+    # slice whatever overlaps (no shifting); then pad left/right as needed
+    end_excl = min(width, first_index + n_side * resampling_scale_factor)
+    decimated_array = target_runw[:,
+                                  first_index:end_excl:resampling_scale_factor]
```
Tyler-g-hudson:
```suggestion
    decimated_array = target_runw[
        :, first_index : end_excl : resampling_scale_factor
    ]
```

python/packages/nisar/workflows/ionosphere.py
```diff
@@ -155,51 +155,27 @@ def decimate_freq_a_offset(iono_insar_cfg, original_dict):
...
         resample_type = 'fine'
     else:
         resample_type = 'coarse'
     decimated_offset_dir = offsets_dir
 
-    # Instantiate a RUNW object to get path to RUNW datasets
-    runw_obj = RUNWGroupsPaths()
+    # Instantiate a RSLC object to get slant range for frequency A and B
+    slc_swath_obj_freqa = isce3.product.Swath(ref_slc_path, 'A')
+    slc_swath_obj_freqb = isce3.product.Swath(ref_slc_path, 'B')
```
Tyler-g-hudson, line 173: A general and out-of-scope comment about the interface of functions like these - in keeping with the Single Responsibility Principle, it's best to split this into a function where one outer function acquires all of the information from any input files and then passes them as parameters into one or several inner functions that perform the actual computation.

The benefits of this are several, but the main ones that come to mind are that the resulting code is more portable and easier to test. In unit tests, it's much easier to test functionality like this if you don't have to provide an entire e.g. RSLC and runconfig, and can instead just throw together some spoofed Swath objects or similar (and potentially parametrize them to test many edge cases)

Tyler-g-hudson (approved):
LGTM! A few stylistic comments only.

python/packages/isce3/signal/interpolate_by_range.py
oberonia78: Recently, I've not observed any cases where either left_missing or right_missing are larger than 0. All cases were 0. So I don't have specific concerns in zero-padding but it is good to have it to avoid the issue when we found the issue in old RSLCs. If NISAR inputs always share the same slant-range coverage, those paddings are almost certainly never triggered.

python/packages/isce3/signal/interpolate_by_range.py
oberonia78: Thanks. @Tyler-g-hudson

python/packages/nisar/workflows/ionosphere.py
```diff
@@ -155,51 +155,27 @@ def decimate_freq_a_offset(iono_insar_cfg, original_dict):
...
+    ref_slc_path = original_dict['reference_rslc_file']
     if iono_insar_cfg['processing']['fine_resample']['enabled']:
         resample_type = 'fine'
     else:
         resample_type = 'coarse'
     decimated_offset_dir = offsets_dir
 
-    # Instantiate a RUNW object to get path to RUNW datasets
-    runw_obj = RUNWGroupsPaths()
+    # Instantiate a RSLC object to get slant range for frequency A and B
```
xhuang-jpl:
```suggestion
    # Instantiate a RSLC Swath object to get slant range for frequency A and B
```

python/packages/isce3/signal/interpolate_by_range.py
```diff
@@ -1,44 +1,81 @@
...
+    total_missing = max(0, n_side - decimated_array.shape[1])
+    left_missing = min(left_missing, total_missing)
+    right_missing = min(right_missing, total_missing - left_missing)
+
+    if left_missing > 0:
+        pad_left = np.zeros(
+            (decimated_array.shape[0], left_missing),
+            dtype=decimated_array.dtype
+        )
+        decimated_array = np.concatenate([pad_left, decimated_array], axis=1)
```
xhuang-jpl: Can we use the np.pad function instead?

https://numpy.org/devdocs/reference/generated/numpy.pad.html

python/packages/isce3/signal/interpolate_by_range.py
```diff
@@ -1,44 +1,81 @@
...
+
+    if right_missing > 0:
+        pad_right = np.zeros(
+            (decimated_array.shape[0], right_missing),
+            dtype=decimated_array.dtype
+        )
+        decimated_array = np.concatenate([decimated_array, pad_right], axis=1)
```
xhuang-jpl: same here

xhuang-jpl (approved):
Thank you @jungkyoJung , some suggestions using the np.pad for the padding. Others look good to me!

hfattahi (approved):

python/packages/isce3/signal/interpolate_by_range.py
oberonia78: very good point!

PR #180, opened 2025-11-12
Fix InSAR issue in Qual-pol (Splitspectrum issue)

oberonia78:
The issue happened when processing InSAR for Qual-pol RSLC. 
In `split_spectrum.py`, the new sub-band RSLC is created from original RSLC. Also, some metadata is updated for the input polarization. However, the broadcast error occurred because the target dataset `listOfPolarizations` in the sub-band files was copied with a fixed length of 4 ([b'VH' b'HV' b'HH' b'VV']), but previous code was trying to write 2 items (['HH','VV']). h5py won’t broadcast (2,) -> (4,).

Solution:
Delete the old dataset and recreate it with length = len(pol_list) and a string dtype. Do this for both low and high sub-band files right where you currently do data = dst_h5_low[...] / data[...] = pol_list.

### DIFF
--- a/python/packages/nisar/workflows/split_spectrum.py
+++ b/python/packages/nisar/workflows/split_spectrum.py
@@ -79,6 +79,15 @@ def prep_subband_h5(src_rslc_hdf5: str,
                             excludes=pols_excludes)
 
 
+def _write_pol_list(h5, dset_path, pol_list):
+    # ensure ASCII fixed/variable string dtype
+    # (2 chars is enough for HH/VV/VH/HV)
+    sdt = h5py.string_dtype(encoding='ascii', length=2)
+    if dset_path in h5:
+        del h5[dset_path]
+    h5.create_dataset(dset_path, data=np.array(pol_list, dtype=sdt))
+
+
 def run(cfg: dict):
     '''
     run split spectrum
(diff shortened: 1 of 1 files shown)


python/packages/nisar/workflows/split_spectrum.py
```diff
@@ -79,6 +79,15 @@ def prep_subband_h5(src_rslc_hdf5: str,
                             excludes=pols_excludes)
 
 
+def _write_pol_list(h5, dset_path, pol_list):
+    # ensure ASCII fixed/variable string dtype
+    # (2 chars is enough for HH/VV/VH/HV)
+    sdt = h5py.string_dtype(encoding='ascii', length=2)
+    if dset_path in h5:
+        del h5[dset_path]
+    h5.create_dataset(dset_path, data=np.array(pol_list, dtype=sdt))
```
xhuang-jpl, line 88: You can use the [to_bytes](https://github.com/isce-framework/isce3/blob/9895af8e41464f6f4b3ac9d56bfe27af2c240ac0/python/packages/nisar/products/utils.py#L8) function to convert the list to string bytes that we used in the InSAR writer.

xhuang-jpl (approved):
Thank you @oberonia78 , this PR LGTM!

hfattahi (approved):
LGTM!

PR #181, opened 2025-11-12
Fix the InSAR offsets products QA failure where there are multiple col pols

xhuang-jpl:
**The Issue**:
The default InSAR  populate ONLY ONE co-pol offset products (HH first, VV second) because of the runconfig parameter (process_single_co_pol_offset) [here](https://github.com/isce-framework/isce3/blob/9895af8e41464f6f4b3ac9d56bfe27af2c240ac0/share/nisar/defaults/insar.yaml#L105C13-L105C41).  The InSAR writer mis-used the polarizations from the function [get_cfg_freq_pols](https://github.com/isce-framework/isce3/blob/9895af8e41464f6f4b3ac9d56bfe27af2c240ac0/python/packages/nisar/workflows/helpers.py#L737) where the third returned parameters are the polarization from the offsets products, but it used the second return parameters. This triggers the inconsistency between the InSAR writer and offset products, which means that the VV offsets  will be created in the InSAR products, but not in the offsets products.

**The solution:**
Update the InSAR writer to use the right polarizations being consistent with the offsets products. 

**The caveat:**
 how we will populate the **listOfPolarizations**  in the RIFG, RUNW, and GUNW products? I mean the interferograms will have two co-polarizations, but the pixelOffsets will have one co-pol polarization, which might fail the QA?


Tyler-g-hudson (approved):
LGTM, at least in terms of logic

nemo794 (changes requested):
Hi @xhuang-jpl , Thanks for putting up this PR to force the conversation! You're correct in that this creates a bit of a mess with QA and with the product specs. I'm going to block the merging for now to prevent any accidents, until the team discusses and agrees upon the correct path forward. Thank you again!

hfattahi:
I believe we do not need this PR anymore as there has been a solution upstream at PCM level.

PR #183, opened 2025-11-13
Add an extra margin to the RSLC LUT extents

gshiroma:
This PR ensures that the RSLC LUTs completely cover the imagery extents. It also adds an extra margin so that, after geocoding with an interpolation algorithm (e.g., bicubic spline), the LUTs still fully cover the geocoded imagery extents.


python/extensions/pybind_isce3/core/Constants.cpp
```diff
@@ -20,4 +20,12 @@ void add_constants(py::module & core)
     core.attr("GLOBAL_MAX_HEIGHT") = py::float_(isce3::core::GLOBAL_MAX_HEIGHT);
     core.attr("WGS84_ELLIPSOID") = isce3::core::Ellipsoid();
     core.attr("SINC_HALF") = isce3::core::SINC_HALF;
+
+    // TODO agree on LUT postings.
+    core.attr("RSLC_CALIBRATION_SECTION_SAMPLING") = 50;
```
hfattahi: can you remind me what is the unit and where 50 comes from?

python/extensions/pybind_isce3/core/Constants.cpp
```diff
@@ -20,4 +20,12 @@ void add_constants(py::module & core)
...
+
+    // Extra margin added to RSLC LUTs to ensure that, after
+    // geocoding with an interpolation algoritm (e.g., bicubic spline),
+    // the LUTs fully cover the geocoded imagery extents.
+    core.attr("RSLC_LUTS_EXTRA_MARGIN_IN_PIXELS") = 11;
```
hfattahi: How confident are you 11 is magic number and would work everywhere?

python/packages/isce3/geometry/doppler.py
```diff
@@ -131,6 +131,12 @@ def make_doppler_lut_from_attitude(
     # computed.
     dem.compute_min_max_mean_height()
 
+    # crop az_time using orbit and attitude extents
+    min_time = max([orbit.start_time, attitude.start_time])
+    max_time = min([orbit.end_time, attitude.end_time])
+    az_time = az_time[(az_time > min_time) & (az_time < max_time)]
```
hfattahi: This is a good thing to do but I assume unrelated to the main purpose of the PR. Is that right? I mean in this PR you are trying to extend the doppler in range direction.

hfattahi (commented):
Thanks @gshiroma . Overall looks good to me. Few comments below to understand the default parameters you have used.

python/extensions/pybind_isce3/core/Constants.cpp
gshiroma: I copied it from the RSLC writer [here](https://github.com/isce-framework/isce3/blob/fb110024451235e43edf6a8ff8d0c92d6ddb7d95/python/packages/nisar/products/writers/SLC.py#L964).

python/extensions/pybind_isce3/core/Constants.cpp
gshiroma: This line assumes that half of the interpolation kernel size will not exceed 11 pixels, which is very conservative.

python/packages/isce3/geometry/doppler.py
gshiroma: We also add a margin in azimuth to ensure that the interpolated LUT covers the RSLC imagery in the azimuth direction as well. I do not recall the exact reason for adding this, but it was probably to fix some issue I encountered extending the Doppler Centroid LUT in the azimuth direction.

bhawkins: I don't like this approach. The API is `make_lut(t, r, *params)` so the result should have shape `(len(t), len(r))`, and if that's not possible then an exception should be thrown. If clipping is needed, it should be done on the calling side.

python/packages/nisar/workflows/focus.py
```diff
@@ -397,7 +397,13 @@ def make_doppler_lut(rawfiles: list[str],
...
     # Now do the actual calculations.
     wvl = isce3.core.speed_of_light / fc
-    epoch_in, t, r = get_total_grid(rawfiles, azimuth_spacing, range_spacing)
+
+    # Get grid for Doppler LUT with an extra margin to ensure that, after
+    # geocoding with an interpolation algoritm (e.g., bicubic spline),
+    # the LUT fully covers the geocoded imagery extents.
+    epoch_in, t, r = get_total_grid(
+        rawfiles, azimuth_spacing, range_spacing,
+        extra_margin_in_pixels=isce3.core.RSLC_LUTS_EXTRA_MARGIN_IN_PIXELS)
```
bhawkins: Maybe `extra_margin_in_pixels` should be configurable?

If time clipping is needed, it should go here. By the way, there's the `ephemeris_crop_pad` parameter [here](https://github.com/isce-framework/isce3/blob/60bbe2f2b2b0a8d99cf9bf647de6eda68e999966/share/nisar/defaults/focus.yaml#L147) that defaults to 10 seconds.  If that's not enough to accommodate the extra margin, then we should increase the default pad.

bhawkins (changes requested):

python/packages/nisar/workflows/focus.py
gshiroma: @bhawkins , I moved the azimuth cropping outside of `make_doppler_lut_from_attitude()`. 

If we consider that PRI is `0.000657894736842105`, the LUT az. time interval will be `50 * 0.000657894736842105 = 0.03289473684` seconds. A margin of `11` LUT az. pixels would represent `0.3618421053` seconds, which is much shorter than `ephemeris_crop_pad = 10` seconds, so I think that default is fine.

I can also expose `extra_margin_in_pixels` in the runconfig. I'll be implementing that soon.

python/packages/nisar/workflows/focus.py
```diff
@@ -409,6 +420,14 @@ def make_doppler_lut(rawfiles: list[str],
         t = [tmin, tmax]
 
     t = convert_epoch(t, epoch_in, epoch)
+
+    # crop the azimuth time using orbit and attitude extents
+    min_time = max([orbit.start_time, attitude.start_time])
+    max_time = min([orbit.end_time, attitude.end_time])
+
+    t = np.asarray(t)
+    t = t[(t > min_time) & (t < max_time)]
```
bhawkins, line 437:
```suggestion
    t = np.asarray(t)
    if np.any(t <= min_time):
        log.warning(f"Desired Doppler LUT start time is {min_time - t[0]} "
            "seconds before ephemeris start. Consider adjusting "
            "ephemeris_crop_pad or providing more orbit/attitude data.")
    if np.any(t >= max_time):
        log.warning(f"Desired Doppler LUT end time is {t[-1] - max_time} "
            "seconds after ephemeris end. Consider adjusting "
            "ephemeris_crop_pad or providing more orbit/attitude data.")
    t = t[(t > min_time) & (t < max_time)]
```
We should let the user know if things aren't working as expected and provide tips to help improve results.

bhawkins (approved):
Thanks for the updates! LGTM pending CI passing and at least one full-size RSLC run.

gshiroma:
> Thanks for the updates! LGTM pending CI passing and at least one full-size RSLC run.

Sounds good. Thank you, @bhawkins !

hfattahi (approved):
LGTM. Thanks @gshiroma

PR #190, opened 2025-11-20
Modify the RSLC processing information to record the L0B granule id

### LINKED ISSUE #184: Normalize the contents of the l0bGranules dataset in RSLC inputs
The RSLC product should probably record just the ancestor L0B granule id in 

```science.LSAR.RSLC.metadata.processingInformation.inputs.l0bGranules```

rather than the full path of the input file as it was seen on the ground processing system. This would make it consistent with the way the rest of the products interpret their corresponding entries in their `inputs`

aivazis:
The contents `science.LSAR.RSLC.metadata.processingInformation.inputs.l0bGranules` now reflect the granule id of the L0B that was focused, rather than the path to the input file in the processing system

### DIFF
--- a/.mm/isce3.cxx.tests
+++ b/.mm/isce3.cxx.tests
@@ -4,7 +4,7 @@
 # the C++ tests
 isce3.cxx.tests.stem := cxx/isce3
 isce3.cxx.tests.prerequisites := isce3.lib
-isce3.cxx.tests.extern := isce3.lib gdal hdf5 mpi eigen fftw pyre gtest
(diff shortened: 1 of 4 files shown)


bhawkins (approved):
I'm not bothered by the mm changes, and the path change seems like a good one.

cc @hfattahi might want to sneak this into the next release.

hfattahi (approved):
LGTM. Thanks!

hfattahi:
I'll merge this to make it to the release today.

PR #192, opened 2025-12-01
RX Channel Imbalances in EAP

rad-eng-59:
This PR computes frequency-dependent polarization-dependent RX channels imbalances and incorporate them into RX DBF process of *pattern.py* module as part EAP in RSLC workflow.

By adding this feature:
1. Residual RX channel imbalances, both phase and magnitude, after applying Caltone is taken into account prior to DBF which leads to proper EL-dependent radiometric correction when *EAP=ON* in RSLC science products.
2. Detect and remove undesired phase jumps due to delay/clock anomaly in one qFSP out of three showing up as an outlier. This will correct for undesired instrument-related phase jump within the swath between two passes in InSAR workflow.

**Disclaimer: Not yet fully tested over a pair of L0Bs with phase jump issue!**


python/packages/nisar/antenna/pattern.py
```diff
@@ -167,12 +182,27 @@ def __init__(self, raw: Raw, dem: DEMInterpolator,
...
+                raise ValueError(
+                    f'freq_band {freq_band} is out of range {freqs}!')
+            self.freq_band = freq_band
+        # get all polarization for a frequency band
+        self.txrx_pols = raw.polarizations[self.freq_band]
+        # comput all RX channel imbalances over all
+        # txrx pols of a desired frequency band.
+        # This RX imbalanced is basically LNA/CALTONE ratio!
+        # XXX perhaps Caltone frequency can be parsed from L0B DRT
+        # rather than provided as an input!
```
Tyler-g-hudson: Is resolving this within the scope of this PR, or would we need to wait until later?

python/packages/nisar/antenna/pattern.py
```diff
@@ -205,14 +235,31 @@ def __init__(self, raw: Raw, dem: DEMInterpolator,
...
+            rxpol = txrx_pol[1]
             if rxpol in rd_all:
                 continue
             rd_all[rxpol], wd_all[rxpol], wl_all[rxpol] = raw.getRdWdWl(
-                chan.freq_id, chan.pol)
+                self.freq_band, txrx_pol)
             self.finder[rxpol] = TimingFinder(self.pulse_times, rd_all[rxpol],
-                                            wd_all[rxpol], wl_all[rxpol])
+                                              wd_all[rxpol], wl_all[rxpol])
+            # XXX get range delay offset for band B in split-spectrum to
```
Tyler-g-hudson: Why is this explanatory comment marked with XXX?

python/packages/nisar/antenna/rx_channel_imbalance_helpers.py
```diff
@@ -0,0 +1,434 @@
...
+import numpy as np
+import h5py
+
+from nisar.products.readers.Raw import Raw
+from nisar.antenna import get_calib_range_line_idx, CalPath
+from isce3.core import speed_of_light
+
+
+@dataclass(frozen=True)
+class RX_CHANNEL_IMBALANCE_PRODUCT:
```
Tyler-g-hudson:
```suggestion
class RxChannelImbalanceProduct:
```
All-caps names like these are typically reserved for constants in Python, is there a reason why you named the class in this way?

python/packages/nisar/antenna/rx_channel_imbalance_helpers.py
```diff
@@ -0,0 +1,434 @@
...
+    max_amp_ratio: float
+        Max amplitude ratio used in peak normalizing `lna_caltone_ratio`.
+
+    """
+    lna_caltone_ratio: np.ndarray
+    ntap_dominant: np.ndarray
+    time_delays_sec: np.ndarray
+    max_amp_ratio: float
+
+    def __post_init__(self):
```
Tyler-g-hudson, line 38: Maybe check the array dtypes here to ensure they're complex/int/float respectively?

python/packages/nisar/antenna/rx_channel_imbalance_helpers.py
```diff
@@ -0,0 +1,434 @@
...
+        # XXX Size of all arrays must be 12 for L-band NISAR but
+        # not enforced due to failure of special cases such as unit test
```
Tyler-g-hudson, line 40: Maybe warn if this isn't the case?

python/packages/nisar/antenna/rx_channel_imbalance_helpers.py
```diff
@@ -0,0 +1,434 @@
...
+        frq_pols = {f: [txrx_pol] for f in frq_pols if txrx_pol in frq_pols[f]}
+
+    out = dict()
+    for freq_band in frq_pols:
+        for txrx_pol in frq_pols[freq_band]:
+            (lna_caltone_ratio, n_tap_dominant, time_delays, max_ratio
+             ) = compute_rx_channel_imbalance(
+                raw,
+                freq_band,
+                txrx_pol,
```
Tyler-g-hudson:
```suggestion
                raw=raw,
                freq_band=freq_band,
                txrx_pol=txrx_pol,
```
Slightly more future-proof and readable to always give your arguments as keyword arguments

python/packages/nisar/antenna/rx_channel_imbalance_helpers.py
```diff
@@ -0,0 +1,434 @@
...
+        Dominant tap number, a value within [1,3] over all 12 RXs.
+    time_delays: np.ndarray(float)
+        Time delays from the phase of qFSP outlier
+    max_ratio : float
+        Report peak power among all channels used for amplitude
+        normalization of RX channel imbalances.
+
+    """
+    lna_mean, n_tap_dominant = get_lna_cal_mean(
+        raw, freq_band, txrx_pol)
```
Tyler-g-hudson:
```suggestion
    lna_mean, n_tap_dominant = get_lna_cal_mean(raw, freq_band, txrx_pol)
```
Why split this line?

python/packages/nisar/antenna/rx_channel_imbalance_helpers.py
```diff
@@ -0,0 +1,434 @@
...
+                    # use RX pol!
+                    p_ds_i = (f'{p_qfsp}/CHIRP_CORRELATOR_I{n_tap}_'
+                              f'{txrx_pol[1]}{n_rx:02d}')
+                    p_ds_q = (f'{p_qfsp}/CHIRP_CORRELATOR_Q{n_tap}_'
+                              f'{txrx_pol[1]}{n_rx:02d}')
+                    # initialize the 3-D array, lines by 12 by 3
+                    if i_qfsp == nn == i_tap == 0:
+                        # XXX get caltype from the very first qFSP assuming
+                        # it is qFSP independent!
+                        p_type = f'{p_qfsp}/CP_CAL_TYPE_{txrx_pol[1]}{i_qfsp}'
```
Tyler-g-hudson: What are the implications of making this assumption? Is there a preferred method?

python/packages/nisar/antenna/rx_channel_imbalance_helpers.py
```diff
@@ -0,0 +1,434 @@
...
+                        # XXX Following Try/exception block is added to
+                        # support old sim L0B products lacking HRT!
```
Tyler-g-hudson: This is pretty deep into a nested FOR loop, is there some earlier point that this could be checked to quickly escape the function?

python/packages/nisar/antenna/rx_channel_imbalance_helpers.py
```diff
@@ -0,0 +1,434 @@
...
+    with h5py.File(raw.filename, mode='r', swmr=True) as f5:
+        # loop over three qfsp
+        for i_qfsp in range(3):
+            p_qfsp = f'{qfsp_path}{i_qfsp}'
+            # loop over 4 channels per qfsp:
+            for nn in range(4):
+                i_chn = nn + i_qfsp * 4
+                n_rx = i_chn + 1
+                # loop over 3 taps
+                for i_tap in range(3):
```
Tyler-g-hudson, line 276: A three-deep FOR loop like this can be avoided as follows, to save you some indenting:
```python
from itertools import product

#...

for i_qfsp, nn, i_tap in product(range(3), range(4), range(3)):
```

python/packages/nisar/antenna/rx_channel_imbalance_helpers.py
```diff
@@ -0,0 +1,434 @@
...
+        chp_cor, cal_type = parse_chirp_corr_from_hrt_qfsp(
+            raw,
+            txrx_pol
+        )
+    n_rxs = chp_cor.shape[1]
+    _, idx_byp, idx_lna, _ = get_calib_range_line_idx(cal_type)
+    if len(idx_lna) == 0:
+        warn('No LNA CAL to represent RX! Use BYPASS Cal instead!')
+        if len(idx_byp) == 0:
+            # XXX to avoid failure in unit test or very short L0B
```
Tyler-g-hudson, line 437: Why is this marked with #XXX, is this behavior undesirable? Should the unit tests be updated?

python/packages/nisar/antenna/rx_channel_imbalance_helpers.py
```diff
@@ -0,0 +1,434 @@
...
+
+
+def correct_lna_caltone_ratio_for_second_band(
+        lna_caltone_ratio: np.ndarray,
+        raw: Raw,
+        freq_band: str,
+        txrx_pol: str,
+        caltone_freq: float | None = None
+) -> Tuple[np.ndarray, np.ndarray]:
+    # Get calton efrequency from DRT if not provided
```
Tyler-g-hudson:
```suggestion
    # Get caltone frequency from DRT if not provided
```

python/packages/nisar/antenna/rx_channel_imbalance_helpers.py
```diff
@@ -0,0 +1,434 @@
...
+    if caltone_freq is None:
+        caltone_freq = parse_caltone_freq_from_drt(raw, txrx_pol)
+        warn(f'Caltone frequency is extracted from {txrx_pol[1]}-pol DRT '
+             f'-> {caltone_freq * 1e-6:.3f} (MHz)')
+    # XXX check if product from the second band so we can modify
```
Tyler-g-hudson: This appears to be a descriptive comment, why is it marked with `# XXX`?

python/packages/nisar/antenna/rx_channel_imbalance_helpers.py
```diff
@@ -0,0 +1,434 @@
...
+    # one of the qFSP is an outlier due to  ADC clock/delay issue
+    # check if there is delay anomaly among three qFSP
+    fc_a, _, _, _ = raw.getChirpParameters('A', txrx_pol[0])
+    # get diff of chirp (band=A) and caltone freq for delay detection
+    dif_chirp_caltone_freq = fc_a - caltone_freq
+    time_delay = _get_qfsp_delay_anomaly(
+        lna_caltone_ratio, dif_chirp_caltone_freq)
+    if _is_product_from_second_band(raw, freq_band, txrx_pol):
+        warn(f'correcting LNA/CALTONE for band={freq_band} and pol={txrx_pol}')
+        # if there is then get diff of frequency bands A dn B
```
Tyler-g-hudson:
```suggestion
        # if there is then get diff of frequency bands A and B
```

python/packages/nisar/antenna/rx_channel_imbalance_helpers.py
```diff
@@ -0,0 +1,434 @@
...
+        txrx_pol: str):
+    """
+    Determine whether the produt is avolable on both bands
+    and it is from the second band.
+    """
+    if freq_band == "B" and len(raw.frequencies) == 2:
+        if txrx_pol in raw.polarizations['A']:
+            return True
+        return False
+    return False
```
Tyler-g-hudson, line 529:
```suggestion
    if freq_band == "B" and len(raw.frequencies) == 2:
        if txrx_pol in raw.polarizations['A']:
            return True
    return False
```
Nitpick: Only one `return False` is necessary since the code will exit the conditional block anyway

python/packages/nisar/antenna/rx_channel_imbalance_helpers.py
```diff
@@ -0,0 +1,434 @@
...
+
+
+def _get_qfsp_delay_anomaly(
+        lna_caltone_ratio: np.ndarray,
+        dif_chirp_caltone_freq: float,
+        adc_clock: float = 240e6) -> np.ndarray:
+    """
+    get time delays for a qfSP with phase anomaly only for
+    12 channel NISAR L-band product.
+    For other case, it will be set to zero!
```
Tyler-g-hudson:
```suggestion
    If the product is a 12-channel NISAR L-band product,
    return the time delays for a qFSP with phase anomaly.
    Else, return zeros.
```

python/packages/nisar/antenna/rx_channel_imbalance_helpers.py
```diff
@@ -0,0 +1,434 @@
...
+        mean_all.append(np.nanmean(data_q1_q3))
+    return np.asarray(mean_all)
+
+
+def _check_if_zero(arr: np.ndarray, msg: str):
+    is_zero = np.isclose(arr, 0)
+    if is_zero.all():
+        # XXX to avoid unit test failure and old sim L0B
+        # a warning will be issued and all values will be set
+        # to unity!
```
Tyler-g-hudson, line 585: Similar concern to other `XXX` comments - is this behavior undesirable? Do we intend to change it later?

Tyler-g-hudson (commented):
Some code organization and documentation commentary

python/packages/nisar/antenna/rx_channel_imbalance_helpers.py
rad-eng-59: Nope if you want clarity and flexibility. This is one-time short computation by using telemetry not echo data.

python/packages/nisar/antenna/rx_channel_imbalance_helpers.py
rad-eng-59: Here we are not after python conciseness but clarity as being implemented and structured by L-band NISAR. See my previous response.

python/packages/nisar/antenna/rx_channel_imbalance_helpers.py
rad-eng-59: Requires attention for future updates of the several data and products used in unit tests and repos while making the update works with both current nominal science products as well as old non-nominal cases  (special cases) w/o failure but a simple warning. The same philosophy applies to all `XXX`.

python/packages/nisar/antenna/rx_channel_imbalance_helpers.py
rad-eng-59: See my previous response.

python/packages/nisar/antenna/rx_channel_imbalance_helpers.py
rad-eng-59: Not necessary. Float is a special case of complex if occurs for any reason and thus shall be supported.

python/packages/nisar/antenna/rx_channel_imbalance_helpers.py
rad-eng-59: I don't understand your question. Let's talk in person. I can't lay out some details here.

PR #195, opened 2025-12-11
Fix the ionosphere runconfig that will change the original runconfig

xhuang-jpl:
This PR is to fix the ionosphere script where the original runconfig get updated. Specifically, 

If `original_cfg` is a dictionary containing nested dictionaries, then .copy() creates a shallow copy and only the top-level dictionary is copied. However, the nested dictionaries are still shared between original_cfg and iono_insar_cfg. So modifying a nested value in iono_insar_cfg will also affect original_cfg, because they reference the same inner object.


oberonia78 (approved):
LGTM.

seongsujeong (approved):
LGTM

PR #196, opened 2025-12-19
`IH5Dataset`: Fix API changes to `Get/SetGeoTransform` methods in GDAL 3.12

scottstanie:
GDAL 3.12 introduced a breaking API change:
`GetGeoTransform()` and `SetGeoTransform()` methods now use a `GDALGeoTransform` class instead of raw double* arrays.

This patch adds conditional compilation to support both the new GDAL 3.12+ API and older versions.

From  https://gdal.org/en/stable/user/migration_guide.html

> Virtual methods [GDALDataset::GetGeoTransform()](https://gdal.org/en/stable/api/gdaldataset_cpp.html#_CPPv4NK11GDALDataset15GetGeoTransformER16GDALGeoTransform) (resp. [GDALDataset::SetGeoTransform()](https://gdal.org/en/stable/api/gdaldataset_cpp.html#_CPPv4N11GDALDataset15SetGeoTransformERK16GDALGeoTransform) have been modified to accept GDALGeoTransform& gt (resp. const GDALGeoTransform& gt) parameters instead of a pointer to 6 doubles. The new class [GDALGeoTransform](https://gdal.org/en/stable/api/gdaldataset_cpp.html#_CPPv416GDALGeoTransform) is a thin wrapper around a std::array<double, 6>. This change affects out-of-tree raster drivers.

### DIFF
--- a/cxx/isce3/io/IH5Dataset.cpp
+++ b/cxx/isce3/io/IH5Dataset.cpp
@@ -385,27 +385,56 @@ void * IH5Dataset::GetInternalHandle(const char *)
 /*                          GetGeoTransform()                           */
 /************************************************************************/
 
-CPLErr IH5Dataset::GetGeoTransform( double *padfGeoTransform )
+#if GDAL_VERSION_MAJOR >= 4 || (GDAL_VERSION_MAJOR == 3 && GDAL_VERSION_MINOR >= 12)
+// GDAL 3.12+ API using GDALGeoTransform class
+CPLErr IH5Dataset::GetGeoTransform( GDALGeoTransform &gt ) const
 {
-    memcpy( padfGeoTransform, adfGeoTransform, sizeof(double) * 6 );
+    gt = GDALGeoTransform(adfGeoTransform);
     if( bGeoTransformSet )
         return CE_None;
 
     return CE_Failure;
 }
+#else
+// GDAL < 3.12 API using double array
+CPLErr IH5Dataset::GetGeoTransform( double *padfTransform )
+{
+    memcpy(padfTransform, adfGeoTransform, sizeof(double) * 6);
+    if( bGeoTransformSet )
+        return CE_None;
+
+    return CE_Failure;
+}
+#endif
 
 /************************************************************************/
 /*                          SetGeoTransform()                           */
 /************************************************************************/
 
-CPLErr IH5Dataset::SetGeoTransform( double *padfGeoTransform )
-
+#if GDAL_VERSION_MAJOR >= 4 || (GDAL_VERSION_MAJOR == 3 && GDAL_VERSION_MINOR >= 12)
+// GDAL 3.12+ API using GDALGeoTransform class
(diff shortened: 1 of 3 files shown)


bhawkins (approved):
Thanks for the patch! LGTM

hfattahi:
Thanks @scottstanie looks good to me. Now it looks like one of the unit tests has a very strict threshold [here](https://github.com/isce-framework/isce3/blob/develop/tests/cxx/isce3/geocode/geocodeCov.cpp#L634). @gshiroma please take a look.

cxx/isce3/io/IH5Dataset.cpp
```diff
@@ -385,27 +385,56 @@ void * IH5Dataset::GetInternalHandle(const char *)
 /*                          GetGeoTransform()                           */
 /************************************************************************/
 
-CPLErr IH5Dataset::GetGeoTransform( double *padfGeoTransform )
+#if GDAL_VERSION_MAJOR >= 4 || (GDAL_VERSION_MAJOR == 3 && GDAL_VERSION_MINOR >= 12)
+// GDAL 3.12+ API using GDALGeoTransform class
+CPLErr IH5Dataset::GetGeoTransform( GDALGeoTransform &gt ) const
 {
-    memcpy( padfGeoTransform, adfGeoTransform, sizeof(double) * 6 );
+    gt = GDALGeoTransform(adfGeoTransform);
```
gshiroma: This is very minor (the original ISCE3 code behaves the same way), but we could copy  `adfGeoTransform` to `gt` (or  `padfTransform `) only when `bGeoTransformSet == true`, to avoid modifying the output on failure `CE_Failure` (e.g., with uninitialized values). Again this is very minor and free to ignore this comment.

gshiroma (approved):
Thank you for the patch, @scottstanie ! I noticed that the original ISCE3 `IH5Dataset::GetGeoTransform()` updates `padfTransform` (your `gt`) regardless of whether it returns `CE_Failure` or not. Maybe we could update the output only if `CE_None` (i.e., geotransform is valid), but I guess both ways are fine. So, there's no need to update your PR based of this. 

Feel free to address @hfattahi 's [comment](https://github.com/isce-framework/isce3/pull/196/#issuecomment-3726364709) to make the `GeocodeCov` unit test pass. You can change the threshold [here](https://github.com/isce-framework/isce3/blob/5b2d6f6057a706d3ff5cfe7342ced8a7933fb302/tests/cxx/isce3/geocode/geocodeCov.cpp#L634). Looking at the full log, it seems that `1.0e-12` is sufficient. But there's a chance you may need to relax the thresholds in the lines below too =/ . I’m also happy to handle this in a separate PR if you prefer. Thanks again.

scottstanie:
> Looking at the full log, it seems that 1.0e-12 is sufficient.
looks like it worked @gshiroma !

gshiroma:
> > Looking at the full log, it seems that 1.0e-12 is sufficient.
> > looks like it worked @gshiroma !

Awesome! Thank you, @scottstanie !

PR #210, opened 2026-01-30
Update the SM SAS Version to v0.4.0

xhuang-jpl:
This PR is to update the soil moisture SAS version to V0.4.0 where the historical SME2 products are brought back to replace the GCOV, which will save many runtime.


xhuang-jpl:
this PR is not needed anymore

PR #211, opened 2026-02-03
Fix Chirp Correlator & Cal Type for QP in NET, RSLC NESZ & EAP

rad-eng-59:
This PR add proper parser for chirp correlator and calibration type (HPA, LNA, BYPASS, and INVALID) in case of QP needed in noise estimator workflow (aka NET), RSLC NESZ and EAP correction workflow. Here are the key updates
1. Augment *Raw* module with proper chirp correlator and cal type parser in particular for QP while preserving the backward compatibility.  Also introduce polarization type class and functionality to the module.
2. Modify NET tool to handle QP properly. Previously half of the QP products are not generated due to missing noise-only range lines (sniffer pulses).
3. Update *pattern.py* to correctly parse cal type and chirp correlator needed to build TX TRM as part of synthesized TX pattern.
4. Slightly modify *focus.py* to support full QP NESZ product generation properly. Previously half of the QP products are filled with zeros.

Tested successfully for QP  `RCID=141` as well as DP `RCID=156`.

### DIFF
--- a/python/packages/nisar/antenna/pattern.py
+++ b/python/packages/nisar/antenna/pattern.py
@@ -6,7 +6,7 @@ import logging
 from nisar.mixed_mode.logic import PolChannelSet
 from nisar.products.readers.antenna import AntennaParser
 from nisar.products.readers.instrument import InstrumentParser
-from nisar.products.readers.Raw import Raw
+from nisar.products.readers.Raw import Raw, chirpcorrelator_caltype_from_raw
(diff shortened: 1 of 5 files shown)


rad-eng-59:
***Note that we want to review and merge this PR first prior to PR #192. Then, I can adjust the RX Imbalance module in the PR #192 to support QP and use L0B calibration  parsers from this PR.***

PR #213, opened 2026-02-09
Ensure that `_FillValue` and statistical attributes have the same data type as the correspondind layer

gshiroma:
This PR ensures that `_FillValue` and statistical attributes have the same data type as the corresponding layer.


bhawkins:
> This PR ensures that `_FillValue` and statistical attributes have the same data type as the corresponding layer.

That makes sense for `_FillValue`. Do you always want that for stats, though? If you compute the mean of an integer-valued image in floating point it might be preferable to report it that way.

gshiroma:
> > This PR ensures that `_FillValue` and statistical attributes have the same data type as the corresponding layer.
> 
> That makes sense for `_FillValue`. Do you always want that for stats, though? If you compute the mean of an integer-valued image in floating point it might be preferable to report it that way.

Thank you for your comment, @bhawkins. We do not compute statistics for the mask layer. Also, note that this code is implemented in `_get_attribute_dict()`, where the default value for the [`to_data_format_function` parameter](https://github.com/gshiroma/isce3/blob/5458d8b6ca6a1626e8561b2b86485fd10e6422f3/python/packages/nisar/products/writers/BaseL2WriterSingleInput.py#L56) is `lambda x: x`. This means that, by default, the data type of `_FillValue` and the statistical values is not modified. This function is (indirectly) called by `save_dataset()`, which is only used by GCOV and by the code that geocodes RSLC LUTs. So, I don’t see this concern occurring in the current products. But it's a good point. Even if that happened, there could be cases in which it could be preferable to compute statistics of integer datasets in float or integer.

For context, I added this feature after noticing that some statistical values were being saved as `float64` for `float32` datasets, and this PR fixes that problem.

bhawkins (approved):
Makes sense.

hfattahi (approved):
LGTM.

hfattahi:
@gshiroma let's merge this unless if you are waiting for something else?

PR #220, opened 2026-02-12
Fix bugs in Ionosphere Phase Uncertainty & connected components mask

oberonia78:
This PR fixes an issue in `IonospherePhaseUncertainty` derived from the `main_diff_ms_band` method.
The problem was in `estimate_iono_std`, where the division operation was incorrect. The operator / has been replaced with *.
image

In addition, there was a bug in `connected_components` used during mask building. This issue has also been fixed.

### DIFF
--- a/python/packages/isce3/atmosphere/main_band_estimation.py
+++ b/python/packages/isce3/atmosphere/main_band_estimation.py
@@ -421,6 +421,7 @@ class MainBandIonosphereEstimation(IonosphereEstimation):
             self,
             main_coh=None,
             side_coh=None,
+            diff_ms_coh=None,
             low_band_coh=None,
             high_band_coh=None,
             diff_low_high_coh=None,
@@ -437,6 +438,8 @@ class MainBandIonosphereEstimation(IonosphereEstimation):
             coherence of main-band interferogram
         side_coh : numpy.ndarray
             coherence of side-band interferogram
+        diff_ms_coh : numpy.ndarray
+            coherence of difference (main-side) interferogram
         low_band_coh : numpy.ndarray
             coherence of low subband interferogram
         high_band_coh : numpy.ndarray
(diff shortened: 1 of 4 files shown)


python/packages/isce3/atmosphere/main_band_estimation.py
```diff
@@ -458,9 +461,21 @@ def estimate_iono_std(
         sig_nondisp : numpy.ndarray
             phase standard deviation of non-dispersive
         """
+        if side_coh is None and diff_ms_coh is None:
+            raise ValueError(
+                "estimate_iono_std requires at least one of "
+                "`diff_ms_coh` or `side_coh` to be provided."
+            )
```
xhuang-jpl: Can you also use journal log here?

python/packages/isce3/atmosphere/main_band_estimation.py
```diff
@@ -718,7 +702,7 @@ def compute_unwrapp_error_main_diff_ms_band(
         (main_runw - (nondisp_array + disp_array)) / (2 * np.pi))
     term1 = f_diff / f0 * nondisp_array
     term2 = f_diff / f1 * disp_array
-    diff_unw_coeff = np.round(
+    diff_unw_coeff = -1 * np.round(
```
xhuang-jpl: where does the -1 come from?

xhuang-jpl (approved):
LGTM, thank you @oberonia78 , some minor comments.

hfattahi:
The bug fixing for estimating phase standard deviation looks good to me. However, the PR also seems to make some changes to the unwrap error correction which was not mentioned in the description. Was that intentional to modify the unwrap error correction in this PR as well? If yes, do you have some equations that you have worked for the unw error correction to check the implementation against and how much this part is tested?

python/packages/isce3/atmosphere/main_band_estimation.py
oberonia78: This change was removed @xhuang-jpl Thanks for pointing out

python/packages/isce3/atmosphere/main_band_estimation.py
oberonia78: journal was added. Good suggestion.

hfattahi (approved):
LGTM. Thanks.

PR #221, opened 2026-02-14
Fix RSLC valid data annotations

bhawkins:
The RSLC product contains datasets `/science/LSAR/RSLC/swaths/frequencyA/validSamplesSubSwath1` (and so on for additional subswaths) that communicate the valid data regions within the image data matrix. The current approach has a problems that manifest in two ways
1) Dithered data are marked as valid everywhere. Basically the L0B valid data mask is ignored because we fill the transmit gaps.
2) The mask for non-dithered data does not update in time. It is only valid at the midpoint of the observation.

Both arise from failing to consider that the receive window position can change along track. Since L0B zero-pads the data to a common starting range, the valid data region can vary significantly for long observations, and it's wrong to assume the full swath can be reconstructed in the dithered case.

Fortunately, the code already supports varying valid data regions as part of mixed-mode processing. This PR only changes the `Raw.getSubSwathBboxes` method to emit multiple lists of subswaths instead of just one per observation. It now uses the (RD, WD, WL) timing parameters to identify where the timing changes and update the subswath annotations.

One complexity for dithered data is that the TX gaps may intersect the start/end of the swath on any given pulse. However, one can assume no consecutive gaps (the whole point of dithering after all), so looking at extrema over two pulses is sufficient to determine the valid data extent.

Unfortunately I'm not able to satisfactorily test these changes because there are also outstanding issues with the L0B valid data mask. So marking this PR as draft for now.


bhawkins:
After some testing with the latest L0B updates, I noticed a small seam in the valid data mask between 10 s timing update boundaries. This was due to the reskew from the raw data geometry to the RSLC geometry. There are some more complicated ways to solve that problem, but a simple one was to sample the perimeter of the polygons more finely. Since I had to tweak it, I made the sample spacing a newly configurable parameter in the runconfig.

Folks at JPL can see the last two slides [here](https://jpl365prod-my.sharepoint.com/:p:/g/personal/brian_p_hawkins_jpl_nasa_gov/IQDn8S_7t-cqR5XYjTa8R37QAYGJnEM7EdMt-_th0Z5QBC8?e=PEJ7gR) where I ran a test on a dithered image with range-timing updates. You can see the mask tracks the range window change, and the edge of the mask corresponds closely to the inflection point in the backscatter power over water.

python/packages/nisar/products/readers/Raw/Raw.py
```diff
@@ -900,6 +936,33 @@ def getAttitude(self):
         return isce3.core.Attitude(old.time, qs, old.reference_epoch)
 
 
+def get_dwp_change_indices(rd, wd, wl):
```
rad-eng-59, line 1127: I think one should track the changes in RX window position that being used in L0B range line alignment. Basically, the variation in combination of `(RD1 + WD1) + RL-START`  as shown [here](https://github.jpl.nasa.gov/NISAR-L0B-STUFF/l0b_qfsp_issue/blob/65a141a28eac72d0532130ca0a65b17abf483d2c/utils/qfsp_delay_helpers.py#L551).
Given this function is put in the raw parser,  it's more relevant to either make it a method of the Raw class or a helper that uses raw for a desired frequency band and polarization similar to the new features recently added to the Raw class with function definition similar to [this](https://github.jpl.nasa.gov/NISAR-L0B-STUFF/l0b_qfsp_issue/blob/65a141a28eac72d0532130ca0a65b17abf483d2c/utils/qfsp_delay_helpers.py#L551).

python/packages/nisar/products/readers/Raw/Raw.py
```diff
@@ -713,71 +713,107 @@ def getSubSwaths(self, frequency='A', tx='H'):
...
+            # Get masked min and max valid sample for each pulse.
+            min_starts = np.min(starts, axis=0)
+            max_ends = np.max(ends, axis=0)
+
+            # Now replace subswaths with a single subswath with start/end
+            # so we can use same logic as fixed PRF below.
+            subswaths = np.vstack((min_starts, max_ends)).transpose().reshape(
+                (1, -1, 2))
+
+        changes = get_dwp_change_indices(rd, wd, wl)
```
rad-eng-59, line 961: This line and line # `752` could,be done in one method as part of Raw with frequency band and TxRx pol as inputs similar to [here](https://github.jpl.nasa.gov/NISAR-L0B-STUFF/l0b_qfsp_issue/blob/65a141a28eac72d0532130ca0a65b17abf483d2c/utils/qfsp_delay_helpers.py#L551). That is 
```python3
changes = self.get_dwp_change_indices(frequency, polarization)
```
By DWP variation, I mean `DWP - DWP.min()` where `DWP=(RD1 + WD1) + RL-START`

python/packages/nisar/products/readers/Raw/Raw.py
```diff
@@ -713,71 +713,107 @@ def getSubSwaths(self, frequency='A', tx='H'):
...
+        rd, wd, wl = self.getRdWdWl(frequency, polarization)
+
+        # Replace enormous fill values with number of samples.
+        subswaths = np.where(subswaths > nr, nr, subswaths)
+
+        # Replace last num_ignore pulses with previous value.
+        if (not is_dithered) and (num_ignore > 0):
+            # Avoid problems with trivially short observations, though this
+            # shouldn't ever happen.
+            num_ignore = min(num_ignore, nt)
```
rad-eng-59: perhaps it is worth to issue a warning whenever modifying an input parameter coming from runconfig.

python/packages/nisar/products/readers/Raw/Raw.py
```diff
@@ -713,71 +713,107 @@ def getSubSwaths(self, frequency='A', tx='H'):
...
+                (1, -1, 2))
+
+        changes = get_dwp_change_indices(rd, wd, wl)
+
+        # Append first and last pulses to generate pairs of constant DWP.
+        breaks = np.hstack(([0], changes, [grid.shape[0] - 1]))
+        bbox_lists = []
+        for ibreak in range(len(breaks) - 1):
+            ipulse0, ipulse1 = breaks[ibreak], breaks[ibreak + 1]
+            t0, t1 = times[ipulse0], times[ipulse1]  # one past end point
```
rad-eng-59, line 968: If `t1` is end time rather than stop time within a fixed DWP block then shouldn't be
```suggestion
            t0, t1 = times[ipulse0], times[ipulse1 - 1]  # one past end point
```

python/packages/nisar/products/readers/Raw/Raw.py
```diff
@@ -900,6 +936,33 @@ def getAttitude(self):
...
+        pulse (i - 1).
+    """
+    if not (rd.shape == wd.shape == wl.shape):
+        raise ValueError("shape mismatch among inputs")
+    if not rd.ndim == 2:
+        raise ValueError("expected 2D input data")
+    start = np.min(rd + wd, axis=1).astype(np.int64)
+    end = np.max(rd + wd + wl, axis=1).astype(np.int64)
+    location = np.vstack((start, end))
+    return np.where(np.any(np.diff(location, axis=1) != 0, axis=0))[0] + 1
```
rad-eng-59, line 1152: instead of `np.where(x != 0)` , one can use `np.nonzero(x)`

python/packages/nisar/products/readers/Raw/Raw.py
```diff
@@ -713,71 +713,107 @@ def getSubSwaths(self, frequency='A', tx='H'):
...
+                # If dithered peek ahead in case gap overlaps start or end of
+                # valid swath.  Only need to check one pulse ahead assuming
+                # dither sequence is correctly designed to avoid consecutive
+                # gaps.
+                if is_dithered:
+                    assert iswath == 0  # due to restructuring above
+                    assert ipulse0 < (nt - 1)  # from construction of breaks
+                    j0next = subswaths[iswath, ipulse0 + 1, 0]
+                    j1next = subswaths[iswath, ipulse0 + 1, 1]
+                    if j1next > j0next:
```
rad-eng-59, line 983: This check is unnecessary given masked array used in dithering case above.

python/packages/nisar/workflows/focus.py
```diff
@@ -1538,16 +1541,19 @@ def get_focused_sub_swaths(rawlist, out_chan, grid, orbit, doppler, dem, azres,
         raw_chan = find_overlapping_channel(raw, out_chan)
 
         freq = raw_chan.freq_id
-        bboxes = raw.getSubSwathBboxes(freq, epoch=orbit.reference_epoch)
-        raw_bbox_lists.append(bboxes)
+        bbox_lists = raw.getSubSwathBboxes(freq, epoch=orbit.reference_epoch,
+            num_ignore=20)  # make num_ignore configurable?
```
rad-eng-59: I would set this to max number of pulses in the air for the worst case, quad pol, that is something around 22 pulses. 
Alternatively, it would be better to use max PRF (fastest) and starting slant range to come up with a value and and add extra 4-5 pulses to it (suitable for 250 km swath )to be on the safe side by covering max pulses in the air for the far part of the swath. No need to bother user for it, IMO.

hfattahi: We also discussed in a PR review if a constant, it would be better to expose to run config.

hfattahi (commented):
We reviewed this at a PR review meeting and overall looks good to me pending @rad-eng-59 comments. We need to run few tests as well as discussed.

hfattahi:
@bhawkins also please rebase with develop so that we can start running some tests. Thanks.

python/packages/nisar/products/readers/Raw/Raw.py
bhawkins: No, we want one past the end like the comment says. Ultimately we want the polygons for consecutive time steps to share an edge.

python/packages/nisar/products/readers/Raw/Raw.py
bhawkins: I didn't know about `numpy.nonzero`, but I don't think that helps here since the `!= 0` is inside the `any` call, not the `where` call. I guess the `!= 0` might be superfluous for an `any` over integers, but I sort of prefer to make the boolean coercion explicit.

python/packages/nisar/products/readers/Raw/Raw.py
bhawkins: I guess that's true as long as there's at least one valid subswath in each rangeline. Now you've got me wondering what happens if a whole line is marked invalid...

bhawkins: I'm thinking if row `ibreak` has some valid data but `ibreak + 1` does not, then we'd also have to check `ibreak + 2` and `ibreak +3` (not just one more since we only have a guarantee about no _consecutive_ gaps). Of course that wouldn't work if we have lots of consecutive rows with no valid data, and we'd probably want to have a corresponding azimuth gap between the returned bboxes in that case. Hmm.

I'm thinking about punting on this for now and leaving the missing-rows case to a future PR.

python/packages/nisar/workflows/focus.py
bhawkins: Looking through the track-frame database, the largest starting range I see is `r0 = 919195.0449973388` m which occurs over the South Pole. The largest PRF I see in the mode table is `prf = 3086.4` Hz for quad-pol (double the per-transmit polarization PRF). Then If I calculate the number of pulses in the air,
```
print(r0 * 2 / c * prf)
18.92645069062936
```
So it seems like 20 is enough? But then you're saying I need to add the 5 more pulses that could arrive in the final receive window. So I'm thinking I'll make it configurable with a default of 25, and hopefully we'll never need to touch it.

bhawkins:
I've addressed everything except incorporating RL_START and RL_STOP. Maybe it's possible those could change while RD+WD+WL stay fixed.

python/packages/nisar/workflows/focus.py
bhawkins: Perhaps there could be a few more pulses in the air if the dither sequence begins with a shorter PRI than the average one. That doesn't seem to be the case for dither sequence 6, which is appears to be the one for the experimental dithered QP mode. But I guess maybe it's good to leave this configurable in case the mode table or observation plan changes somehow.

python/packages/nisar/products/readers/Raw/Raw.py
rad-eng-59: FYI, for the start, one simply needs the first channel  `start = rd[:, 0] + wd[:, 0] + RL_START` while for the end it shall be `end = rd[:, 0] + wd[:, 0] + RL_STOP`. The start used for range line alignment in L0B while the number of range bins is eventually determined by the `RL_STOP - RL_START` if I am not mistaken. The total number of valid range bins is set by `RL_STOP - RL_START` if I'm not mistaken.  A couple of notes per my understanding:
1. The RX channels are opened in ascending order. It would be very strange for RX # `6` to open prior to `1` and if so that is a bug in respective telemetry field which will be ignored here by using min/max. 
2. If the order of RD/WD/WL is assumed to be arbitrary then several functionalities in ISCE3 will fall apart. The low level L0 stored these fields in an ascending order of RX channels not randomly.
3. If the concern is the first active RX channel, then both L0B and L1 can use first index in the list of active RX channels as a the first RX channel index in the respective `RD/WD`. Although, I still think the L-band is designed to provide and use all 12 channel telemetries even when echo for some of RX channels are inactive like half-swath case.
4. WL for the last three channels can be very large beyond the number of valid range bins that contain echo. Using the largest `RD+WD+WL` can lead to more samples that actually exists.

python/packages/nisar/products/readers/Raw/Raw.py
rad-eng-59: Okay. Unless I have misunderstood this, I doubt it is possible for the instrument to have a rangeline with no valid data given limited duty cycle of the instrument per any feasible PRF.  Note that it is possible and feasible for the `changes` to be an empty container. For instance where the observation is too short or the geometry/topography barely changes within `20-30` seconds at the output sampling rate. I have run into a case before. However, that should not break your code here per how I am interpreting this loop.

python/packages/nisar/products/readers/Raw/Raw.py
rad-eng-59: So, you want the polygons to overlap at least one range line. I saw your comment but  misinterpreted your intention.

python/packages/nisar/workflows/focus.py
rad-eng-59: Yes, I'm good with any default value between `22-25`. Also, no objection to make it configurable although I doubt we would ever need to modify it.

rad-eng-59 (approved):
LGTM. No major comment or blocker. Thank you

bhawkins:
Oops, the warning I added in [28a4fe1](https://github.com/isce-framework/isce3/pull/221/commits/28a4fe1baa74fe16ab70c6d677ba3bbdbe69ce01) should've been in the else clause. Aside from some log spam it's harmless, but I'll fix it shortly.

bhawkins:
CI failure in `resample_slc`, the same one mentioned here https://github.com/isce-framework/isce3/pull/104#issuecomment-3207851395

bhawkins:
Re-ran CI and all tests passed 🤷

hfattahi:
I ran through a pair of acquisitions with this PR and so far the results are satisfactory. Here is a plot of the valid data mask overlayed on the RSLC data for frequency B:

Screenshot 2026-03-17 at 10 47 05 PM

A zoom into far range at the boundary of valid and invalid data, shows that partially focused data is correctly falling on the invalid part of the mask:

Screenshot 2026-03-17 at 10 46 02 PM

hfattahi (approved):
Thanks @bhawkins . LGTM. I extended the test to InSAR and confirm that the fix results in accurate valid data mask for interferogram and ionosphere as well.

jshimada47:
@bhawkins I'm getting an error that I haven't seen before with one of my test cases.  The test case happens to use an input L0B that's relatively old (Sep. 2025):
```
z2026/03/19 22:44:13.467000 - 2026-03-19 22:44:13.467, INFO, RSLC, focus, 999999, /opt/conda/lib/python3.12/site-packages/nisar/workflows/focus.py:1809, "computing valid swaths"
2026/03/19 22:44:13.483000 - Traceback (most recent call last):
2026/03/19 22:44:13.483000 -   File "<frozen runpy>", line 198, in _run_module_as_main
2026/03/19 22:44:13.483000 -   File "<frozen runpy>", line 88, in _run_code
2026/03/19 22:44:13.483000 -   File "/opt/conda/lib/python3.12/site-packages/nisar/workflows/focus.py", line 2248, in <module>
2026/03/19 22:44:13.483000 -     main(sys.argv[1:])
2026/03/19 22:44:13.483000 -   File "/opt/conda/lib/python3.12/site-packages/nisar/workflows/focus.py", line 2244, in main
2026/03/19 22:44:13.484000 -     focus(cfg, args.run_config_path)
2026/03/19 22:44:13.484000 -   File "/opt/conda/lib/python3.12/site-packages/nisar/workflows/focus.py", line 1810, in focus
2026/03/19 22:44:13.484000 -     valid_swaths = get_focused_sub_swaths(rawlist, chan, og, orbit,
2026/03/19 22:44:13.484000 -                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2026/03/19 22:44:13.484000 -   File "/opt/conda/lib/python3.12/site-packages/nisar/workflows/focus.py", line 1589, in get_focused_sub_swaths
2026/03/19 22:44:13.484000 -     bbox_lists = raw.getSubSwathBboxes(freq, epoch=orbit.reference_epoch,
2026/03/19 22:44:13.484000 -                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2026/03/19 22:44:13.484000 -   File "/opt/conda/lib/python3.12/site-packages/nisar/products/readers/Raw/Raw.py", line 939, in getSubSwathBboxes
2026/03/19 22:44:13.484000 -     subswaths[:, -num_ignore:, :] = subswaths[:, -num_ignore, :]
2026/03/19 22:44:13.484000 -     ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^
2026/03/19 22:44:13.484000 - ValueError: could not broadcast input array from shape (3,2) into shape (3,25,2)
2026/03/19 22:44:13.694000 - Error:
2026/03/19 22:44:13.694000 - Traceback (most recent call last):
2026/03/19 22:44:13.694000 -   File "/balmony/balmony/_shell_commands.py", line 49, in safe_run
2026/03/19 22:44:13.694000 -     result: CompletedProcess = run(
2026/03/19 22:44:13.694000 -                                ^^^^
2026/03/19 22:44:13.694000 -   File "/opt/conda/lib/python3.12/subprocess.py", line 571, in run
2026/03/19 22:44:13.694000 -     raise CalledProcessError(retcode, process.args,
2026/03/19 22:44:13.694000 - subprocess.CalledProcessError: Command '['/bin/bash', '-c', 'mamba run -n base  python -u -m nisar.workflows.focus runconfig.yaml --no-log-file']' returned non-zero exit status 1.
2026/03/19 22:44:13.694000 - Exception encountered when attempting to run command: /bin/bash -c 'mamba run -n base  python -u -m nisar.workflows.focus runconfig.yaml --no-log-file'
2026/03/19 22:44:13.694000 - EXIT CODE: 1

```
I'm not sure whether it's related to this PR.  I'm testing a new Docker image that has most of the R05.01.3 milestones (including this PR) with a newer base Docker image (OracleLinux 8.10 with patches through 3/1/2026) and a newer version of pysolid.  This error occurred about 30 seconds into execution and did not occur with the R05.01.2 Docker image.  Does your PR assume that the input L0B was generated with a recent version of the L0B SAS?  FYI, I'm running a few other test cases with October 2025 RRSDs.  So far they've all been running for at least 10 minutes.                                                                                                                                                                                                        

The two log files with this error are on nisar-adt-dev-3 under /scratch/jshimada/iscale/r05.01.2/newbase-pysolid/morocco_01/FAILED.

PR #226, opened 2026-02-19
Make isce3 work with numpy 2.x

piyushrpt:
- `isce3` currently is still pinned to numpy < 2, making it hard to use with most analytics tools that have moved on to newer versions
- This PR will be in draft state as we fix issues that we run into when working with isce3 in newer environments
- Current env used for this PR
    - OSX 26.3
    - numpy 2.4.2
    - GDAL 3.12.2
    - HDF5 2.0.0
    - h5py 3.15.1
- We are not able to test any of the nisar workflows as the test data is not open
- These changes ensure that unit tests work with newer setup

### DIFF
--- a/python/packages/isce3/io/gdal/gdal_raster.py
+++ b/python/packages/isce3/io/gdal/gdal_raster.py
@@ -20,10 +20,7 @@ GDALRasterT = TypeVar("GDALRasterT", bound="GDALRaster")
 
 def get_gdal_dtype(type: DTypeLike) -> int:
     """Returns the GDAL data type associated with a given NumPy dtype."""
-    try:
-        np_dtype = np.dtype(type)
-    except Exception:
-        np_dtype = np.dtype(type.dtype)
+    np_dtype = np.dtype(type)
 
     if np_dtype == np.complex128:
         return gdal.GDT_CFloat64
--- a/python/packages/isce3/io/gdal/gdal_raster.py
+++ b/python/packages/isce3/io/gdal/gdal_raster.py
@@ -20,7 +20,10 @@ GDALRasterT = TypeVar("GDALRasterT", bound="GDALRaster")
 
 def get_gdal_dtype(type: DTypeLike) -> int:
     """Returns the GDAL data type associated with a given NumPy dtype."""
-    np_dtype = np.dtype(type)
+    try:
+        np_dtype = np.dtype(type)
+    except Exception:
+        np_dtype = np.dtype(type.dtype)
 
     if np_dtype == np.complex128:
         return gdal.GDT_CFloat64
--- a/python/packages/nisar/cal/faraday_rotation_angle_slc.py
+++ b/python/packages/nisar/cal/faraday_rotation_angle_slc.py
@@ -387,14 +387,14 @@ class FaradayRotAngleSlc(ABC):
 
         # get range bins per block and the number of blocks of sr
         n_rgb_blk = round(sr_blk_size / self._rdr_grid.range_pixel_spacing)
-        n_blks_sr = int(np.ceil(np.diff(rgb_limit) / n_rgb_blk))
+        n_blks_sr = int(np.ceil(np.diff(rgb_limit).item() / n_rgb_blk))
 
         self.logger.info(f'Full block size in slant range -> {n_rgb_blk}')
         self.logger.info(f'Total number of range blocks -> {n_blks_sr}')
 
         # get azimuth bins per block and the number of blocks of azt
         n_azb_blk = round(azt_blk_size * self._rdr_grid.prf)
-        n_blks_azt = int(np.ceil(np.diff(azb_limit) / n_azb_blk))
+        n_blks_azt = int(np.ceil(np.diff(azb_limit).item() / n_azb_blk))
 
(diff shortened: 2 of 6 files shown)


hfattahi:
@piyushrpt is this PR still draft or you think ready to be reviewed.

piyushrpt:
I think it is ready to be reviewed - it works for external users for isce3. It would be good to run this internally on all NISAR workflows to make sure things dont break.

piyushrpt:
I have rebased this on to the latest to make sure all the tests pass.

python/packages/isce3/io/gdal/gdal_raster.py
```diff
@@ -20,7 +20,10 @@
 
 def get_gdal_dtype(type: DTypeLike) -> int:
     """Returns the GDAL data type associated with a given NumPy dtype."""
-    np_dtype = np.dtype(type)
+    try:
+        np_dtype = np.dtype(type)
+    except Exception:
+        np_dtype = np.dtype(type.dtype)
```
bhawkins: I'm sort of confused by this one as the `numpy.dtype` constructor docs ([link](https://numpy.org/doc/stable/reference/arrays.dtypes.html)) say

> Types with .dtype
> 
> Any type object with a dtype attribute: The attribute will be accessed and used directly. The attribute must return something that is convertible into a dtype object.

If we have to do something like this, I'd prefer if the error handling was a little more specific. However, it seems it can throw TypeError or ValueError depending on how it fails. Anyways, I think the following is a little cleaner:

```suggestion
    try:
        np_dtype = np.dtype(type)
    except (TypeError, ValueError):
        if hasattr(type, "dtype"):
            np_dtype = np.dtype(type.dtype)
        else:
            raise
```
Note that a potential one-liner `np_dtype = np.dtype(getattr(type, "dtype", type))` doesn't work because `np.dtype(np.float32.dtype)` throws a TypeError...

The CI failure occurs due the unit test [here](https://github.com/isce-framework/isce3/blob/1654419a349592171160262c26ae4f23d984c73f/tests/python/packages/isce3/io/gdal/gdal_raster.py#L152). If we change the attribute to be an actual dtype object (e.g., `dtype = np.dtype(np.complex64)`) then the test also passes. I'm wondering if we actually use this "object that has a dtype attribute" behavior with custom objects in isce3 or whether it's just for numpy objects that probably behave correctly across both versions. Or maybe we should update our objects in a compatible way. It looks like @Tyler-g-hudson wrote this stuff, so maybe he can chime in.

bhawkins (commented):
I appreciate the patches. I just have some questions for @Tyler-g-hudson about the `dtype` stuff.

python/packages/isce3/io/gdal/gdal_raster.py
piyushrpt: Yep. This was precisely due to the single unit test and I couldn't figure out the scope of the change - so tried to keep it minimal.

Tyler-g-hudson: Presently the only use of this function is for the GDALRaster class which is used in `resample_slc_v2`.

The function of this class is to provide a numpy-like interface to a GDAL raster which can be used interchangeably with numpy arrays for the purposes of reading and writing within the context of resampling a secondary SLC onto a reference grid.

The functionality is used once in `resample_slc.py` [here](https://github.com/isce-framework/isce3/blob/1654419a349592171160262c26ae4f23d984c73f/python/packages/isce3/image/v2/resample_slc.py#L100) - this line presumes the `out_dataset`, which could be an np.ndarray or a GDALRaster (or any other class that fulfills the `DatasetWriter` protocol,) has a numpy dtype.

piyushrpt: Happy to change that [unit test](https://github.com/isce-framework/isce3/blob/1654419a349592171160262c26ae4f23d984c73f/tests/python/packages/isce3/io/gdal/gdal_raster.py#L152) to just use `numpy.zeros` instead of a data class if that satisfies the use cases. Good to know that this only used in one place.

Tyler-g-hudson: I read through the code and more thoroughly understood what was going on - apologies for my rust, it's been a long time since I wrote this code.

The particular class-with-dtype functionality being tested in that test is never used within ISCE3, I'm scratching my head to figure out why I required that behavior to begin with except that it was an option in that version of numpy.

@bhawkins' intuition is right; we can safely get rid of that unit test rather than adding a check here, it doesn't effect anything within the codebase and the functionality it tests for is not intuitive for users.

Tyler-g-hudson:
```suggestion
    np_dtype = np.dtype(type)
```
recommend reverting this change since the test that necessitated it will also go away

piyushrpt: Done

piyushrpt: Test removed as it is not expected usage

bhawkins (approved):
LGTM

Tyler-g-hudson (approved):
LGTM, thanks!

PR #231, opened 2026-03-03
Fix Chirp Correlator and Cal Type in Raw, NET, RSLC EAP and NEB

rad-eng-59:
This PR is the merged combination of the two exiting PR #192 and PR #211 plus some minor improvements.
The main reason for the merge is that these two older PRs are fixing mainly the the same underlying issue with telemetry but for two different applications. Besides, this avoids merging conflict in future while placing the respective common helper functions under appropriate namespaces at once. Thus, once this PR is reviewed and ready to merge, those two older PRs can be safely removed. 

Here is the summary of key updates:
1. Improve Raw parser with helper functions to parse chirp correlator and cal type properly by adding a support for cross-pol products as well as full polarimetric products for Quad-pol case that is currently wrong or missing.
2.   Add a new module to compute residual RX channel imbalances, detect anomalies in received echoes (aka sample slip) where they are used as part of EAP in RSLC.
3. Update antenna pattern module to use frequency band and caltone for proper synthesis of RX DBF pattern. This requires caltone frequency parser incorporated in Raw L0B parser.
4. Update noise estimator module (NET) to handle QP properly with correct chirp correlator and cal type values. Also fix some potential biases in noise estimation due to dynamic CPI size in default MEE approach.
5. Update python API for noise estimator module used in RSLC to estimate NEB. The update adds support for QP w/ correct chirp correlator and cal type telemetry info and force CPI to stay as close as to the input value (or its maximum) for more robust noise power estimation across swath.


rad-eng-59:
@Tyler-g-hudson , FYI your valuable comments from #192 are already incorporated in this PR.

python/packages/nisar/products/readers/Raw/Raw.py
```diff
@@ -399,6 +403,140 @@ def getRangeLineIndex(self, frequency: str = 'A', tx: str = None):
             return f[path]["rangeLineIndex"][()]
 
 
+    def _parse_chirpcorrelator_from_hrt_qfsp(
+            self,
+            txrx_pol: str) -> np.ndarray | None:
+        """
+        Parse three-tap chirp correlator array with shape (lines, 12, 3)
+        as well ass cal type with shape (lines,) from HRT QFSP.
```
bhawkins:
```suggestion
```
Typo (🚫 🍑) and it looks like a different function returns the cal type.

python/packages/nisar/products/readers/Raw/Raw.py
```diff
@@ -399,6 +403,140 @@ def getRangeLineIndex(self, frequency: str = 'A', tx: str = None):
...
+                        p_ds_q = (f'{p_qfsp}/CHIRP_CORRELATOR_Q{n_tap}_'
+                                  f'{txrx_pol[1]}{n_rx:02d}')
+                        try:
+                            ds_i = f5[p_ds_i]
+                        except KeyError as err:
+                            warn(
+                                f'Missing dataset {p_ds_i} in {self.filename}.'
+                                f' Detailed error -> {err}'
+                            )
+                            return None
```
bhawkins: I think I prefer an exception instead of a special return value, but maybe I'll reconsider after seeing the usage. I guess it's a private method (leading underscore in name), so maybe that should be considered as well.

python/packages/nisar/products/readers/Raw/Raw.py
```diff
@@ -399,6 +403,140 @@ def getRangeLineIndex(self, frequency: str = 'A', tx: str = None):
...
+            p_qfsp = f'{qfsp_path}{i_qfsp}'
+            p_type = f'{p_qfsp}/CP_CAL_TYPE_{txrx_pol[1]}{i_qfsp}'
+            # XXX Following Try/exception block is added to
+            # support old sim L0B products lacking HRT!
+            try:
+                ds_cal_type = f5[p_type]
+            except KeyError as err:
+                warn(f'Missing dataset "{p_type}" in '
+                    f'"{self.filename}". Detailed error -> {err}')
+                return None
```
bhawkins: C style again

python/packages/nisar/products/readers/Raw/Raw.py
```diff
@@ -399,6 +403,140 @@ def getRangeLineIndex(self, frequency: str = 'A', tx: str = None):
...
+        elif txrx_pol not in pols:
+            raise ValueError(f'Available pols {pols} but got {txrx_pol}!')
+        rgl_idx_path = (f'{hrt_path}/tx{txrx_pol[0]}/rx{txrx_pol[1]}/'
+                        'RangeLine/RH_RANGELINE_INDEX')
+        with h5py.File(self.filename, mode='r', swmr=True) as f5:
+            try:
+                ds_rgl_idx = f5[rgl_idx_path]
+            except KeyError as err:
+                warn(f'Can not parse range line index from HRT. Error -> {err}')
+                return None
```
bhawkins: C style again

python/packages/nisar/products/readers/Raw/Raw.py
```diff
@@ -918,3 +1055,212 @@ def open_rrsd(filename) -> RawBase:
...
+
+
+def opposite_pol(pol: str) -> str:
+    """Get the oppsoite pol"""
+    if pol == 'H':
+        return 'V'
+    elif pol == 'V':
+        return 'H'
+    else:
+        return pol
```
bhawkins: I think we should handle L and R, and raise an exception for others.

python/packages/nisar/products/readers/Raw/Raw.py
```diff
@@ -918,3 +1055,212 @@ def open_rrsd(filename) -> RawBase:
...
+
+
+def chirpcorrelator_caltype_from_raw(
```
bhawkins, line 1282: Need something in the function name and/or description as to why this function is in the API in addition to the existing methods of Raw that seem to do the same thing.

Overall, this function is kind of a kludge. It seems cleaner to change the L0B spec to move the `chirpCorrelator` datasets down a level into the `txX/rxY` group, and then all this fiddly stuff becomes an L0B formatting problem. But I guess that'd take too much time to implement and test.

Edit: AFAICT this method should just replace the existing instance method. I guess the concern might be supporting simulations lacking the telemetry fields, but that could be handled by renaming the current method (e.g., to the private `_getSwathsChirpCorrelator`) and calling it as a fallback as done here. We'd also want to raise an exception if `len(txrx_pol) == 1` in order to catch any calls this PR might have missed.

python/packages/nisar/products/readers/Raw/Raw.py
```diff
@@ -918,3 +1055,212 @@ def open_rrsd(filename) -> RawBase:
...
+        raw: Raw,
+        txrx_pol: str
+) -> Tuple[np.ndarray, np.ndarray]:
+    """
+    Parse three-tap chirp correlator array with shape (lines, 12, 3)
+    as well ass cal type with shape (lines,) from Raw L0B for a certain
```
bhawkins:
```suggestion
    as well as cal type with shape (lines,) from Raw L0B for a certain
```
🚫 🍑

python/packages/nisar/products/readers/Raw/Raw.py
```diff
@@ -918,3 +1055,212 @@ def open_rrsd(filename) -> RawBase:
...
+
+def polarization_type_from_drt(raw: Raw) -> PolarizationTypeId:
+    """Get polarization ID and type from L0B DRT"""
+    pol_path = f'{raw.TelemetryPath}/DRT/MISC/CP_IFSW_POLARIZATION'
+    with h5py.File(raw.filename, mode='r', swmr=True) as f5:
+        try:
+            ds_pol = f5[pol_path]
+        except KeyError:
+            warn(f'Missing dataset "{pol_path}" in "{raw.filename}"')
+            id_pol = 6
```
bhawkins:
```suggestion
            id_pol = PolarizationTypeId.none
```
Prefer to avoid copying magic values.

python/packages/nisar/products/readers/Raw/Raw.py
```diff
@@ -918,3 +1055,212 @@ def open_rrsd(filename) -> RawBase:
...
+    chp_cor = raw._parse_chirpcorrelator_from_hrt_qfsp(txrx_pol=txrx_pol)
+    cal_type = raw._parse_caltype_from_hrt_qfsp(txrx_pol=txrx_pol)
+    # XXX if the respective field does not exist then use co-pol under
+    # swath in L0B for the sake of backward compatibility
+    if chp_cor is None or cal_type is None:
+        freq_band = [f for f in raw.frequencies if
+                     txrx_pol in raw.polarizations[f]][0]
+        chp_cor = raw.getChirpCorrelator(freq_band, txrx_pol[0])
+        cal_type = raw.getCalType(freq_band, txrx_pol[0])
+        return chp_cor, cal_type
```
bhawkins, line 1315:
```suggestion
        return chp_cor, cal_type
    elif polarization_type_from_drt(raw) == PolarizationTypeId.none:
        raise ValueError("unable to determine whether data are quad pol")
```
Edge case due to hiding the `KeyError` in `polarization_type_from_drt`. Stuff like this is why I think it'd be better to just throw exceptions.

python/packages/nisar/products/readers/Raw/Raw.py
```diff
@@ -918,3 +1055,212 @@ def open_rrsd(filename) -> RawBase:
...
+def is_raw_quad_pol(raw: Raw) -> bool:
+    """Determine whether raw L0B is Quad or not"""
+    return polarization_type_from_drt(raw) == PolarizationTypeId.quad
+
+
+def first_tx_pol_for_quad(raw: Raw) -> str:
+    """Get first TX polarization, H or V, from only Quad pol product"""
+    if not is_raw_quad_pol(raw):
+        raise ValueError('Not a quad pol!')
+    idx_rgl = raw._parse_rangeline_index_from_hrt()[0]
```
bhawkins: The indexing `[0]` will fail whenever the function returns `None`, so we'll never see the logic that follows. A try/except would be more natural and probably would've avoided this problem.

python/packages/nisar/products/readers/Raw/Raw.py
```diff
@@ -918,3 +1055,212 @@ def open_rrsd(filename) -> RawBase:
...
+    # if not in HRT parse single-pol version from swath path
+    if idx_rgl is None:
+        idx_rgl_h = raw.getRangeLineIndex('A', 'H')[0]
+        idx_rgl_v = raw.getRangeLineIndex('A', 'V')[0]
+        if idx_rgl_v < idx_rgl_h:
+            return 'V'
+        return 'H'
+    else:  # odd range line is V pol first and even is H pol first!
+        return {0: 'H', 1: 'V'}.get(idx_rgl % 2)
+
```
bhawkins, line 1253: I'm confused about why we actually need two algorithms here. If they don't give the same result, doesn't that mean the L0B is wrong anyway? If possible it seems better to just use the `raw.getRangeLineIndex` method since that doesn't depend on the NISAR LSAR even/odd rule.

python/packages/nisar/products/readers/Raw/Raw.py
```diff
@@ -918,3 +1055,212 @@ def open_rrsd(filename) -> RawBase:
...
+        cal_type = raw.getCalType(freq_band, txrx_pol[0])
+        return chp_cor, cal_type
+    # Quad pol case
+    if is_raw_quad_pol(raw):
+        tx_pol_first = first_tx_pol_for_quad(raw)
+        if txrx_pol[0] == tx_pol_first:
+            chp_cor = chp_cor[::2]
+            cal_type = cal_type[::2]
+        else:  # the second TX pol
+            # get data from the opssoite TX pol
```
bhawkins:
```suggestion
            # get data from the opposite TX pol
```
Need some finger exercises https://en.wikipedia.org/wiki/Typing_game

python/packages/nisar/products/readers/Raw/Raw.py
```diff
@@ -918,3 +1055,212 @@ def open_rrsd(filename) -> RawBase:
...
+            x_pol = opposite_pol(txrx_pol[0]) + txrx_pol[1]
+            chp_cor_x, cal_type_x = chirpcorrelator_caltype_from_raw(
+                raw, txrx_pol=x_pol)
+            # if co-pol get HPA value from same TX but
+            # fill in LNA/BYP from oppsoite TX
```
bhawkins:
```suggestion
            # fill in LNA/BYP from opposite TX
```

python/packages/nisar/products/readers/Raw/Raw.py
```diff
@@ -918,3 +1055,212 @@ def open_rrsd(filename) -> RawBase:
...
+    if is_raw_quad_pol(raw):
+        tx_pol_first = first_tx_pol_for_quad(raw)
+        if txrx_pol[0] == tx_pol_first:
+            chp_cor = chp_cor[::2]
+            cal_type = cal_type[::2]
+        else:  # the second TX pol
+            # get data from the opssoite TX pol
+            x_pol = opposite_pol(txrx_pol[0]) + txrx_pol[1]
+            chp_cor_x, cal_type_x = chirpcorrelator_caltype_from_raw(
+                raw, txrx_pol=x_pol)
```
bhawkins:
```suggestion
            x_pol = opposite_pol(txrx_pol[0]) + txrx_pol[1]
            assert x_pol != txrx_pol
            chp_cor_x, cal_type_x = chirpcorrelator_caltype_from_raw(
                raw, txrx_pol=x_pol)
```
I guess this is safe because we already know `is_raw_quad_pol(raw)` is True, and there's no circular version of QP. Otherwise there could be infinite recursion, for example when `opposite_pol("L")` returns `"L"`. Another case where an exception makes things easier IMO.

python/packages/nisar/products/readers/Raw/Raw.py
```diff
@@ -918,3 +1055,212 @@ def open_rrsd(filename) -> RawBase:
...
+                chp_cor = chp_cor[1::2]
+                cal_type = cal_type[1::2]
+                _, idx_byp, idx_lna, _ = get_calib_range_line_idx(cal_type_x)
+                chp_cor[idx_byp] = chp_cor_x[idx_byp]
+                chp_cor[idx_lna] = chp_cor_x[idx_lna]
+                cal_type[idx_byp] = CalPath.BYPASS
+                cal_type[idx_lna] = CalPath.LNA
+            else:  # x-pol product
+                chp_cor = chp_cor_x
+                cal_type = cal_type_x
```
bhawkins, line 1339: Don't we need to stride by two? Otherwise xpol has double the length and all the indexing is wrong.

python/packages/nisar/products/readers/Raw/Raw.py
```diff
@@ -918,3 +1055,212 @@ def open_rrsd(filename) -> RawBase:
...
+    # set x-pol HPA to INVALID given they are the mix of
+    # LNA from co-pol and HPA from x-pol!
+    if txrx_pol in ('HV', 'VH'):
+        idx_hpa, _, _, _ = get_calib_range_line_idx(cal_type)
+        if idx_hpa.size > 0:
+            warn(f'Set HPA cal type for x-pol {txrx_pol} to INVALID!')
```
bhawkins:
```suggestion
            log.info(f'Set HPA cal type for x-pol {txrx_pol} to INVALID!')
```
This doesn't seem like it should be a warning given that it's the expected behavior.

python/packages/nisar/products/readers/Raw/Raw.py
```diff
@@ -918,3 +1055,212 @@ def open_rrsd(filename) -> RawBase:
...
+           f'{txrx_pol[1]}')
+    with h5py.File(raw.filename, mode='r', swmr=True) as f5:
+        try:
+            ds_caltone_phase = f5[c_p]
+        except KeyError:
+            warn(f'Missing path "{c_p}" in L0B! Caltone frequency will '
+                 f'be set to {default} (Hz)')
+            return default
+        else:
+            i_cal = np.median(ds_caltone_phase[()]).astype(int)
```
bhawkins, line 1389:
```suggestion
            i_cal = np.median(ds_caltone_phase[()]).astype(int)
            if i_cal < 2**16:
                warn("CALTONE_PHASE_STEP seems too small, caltone frequency may be invalid")
```
I'm pretty sure I've had L0B files where not all bits were parsed correctly. That's already been fixed going forward, but a warning like this might help if someone happens to try running with old data.

python/packages/nisar/products/readers/Raw/Raw.py
```diff
@@ -918,3 +1055,212 @@ def open_rrsd(filename) -> RawBase:
...
+    """
+    # default caltone if dataset is not available (Hz)
+    default = 1214.883e6
+    # frequency of local oscillator (Hz)
+    lo = 1200e6
+    # ADC clock (Hz)
+    clock = 240e6
+    c_p = (f'{raw.TelemetryPath}/DRT/MISC/CP_IFSW_CALTONE_PHASE_STEP_'
+           f'{txrx_pol[1]}')
+    with h5py.File(raw.filename, mode='r', swmr=True) as f5:
```
bhawkins, line 1381: All these module functions that actually open and parse the underlying HDF5 file seem like they should be `Raw` instance methods.

python/packages/nisar/products/readers/Raw/Raw.py
```diff
@@ -918,3 +1055,212 @@ def open_rrsd(filename) -> RawBase:
...
+    """
+    # check if band is B and it is split spectrum
+    if freq_band == 'B' and len(raw.frequencies) == 2:
+        pols = raw.polarizations
+        # check if this is sequential transmit
+        if txrx_pol in pols['A']:
+            sr_b = raw.getRanges('B', txrx_pol[0])
+            sr_a = raw.getRanges('A', txrx_pol[0])
+            delay = 2 * (sr_b.first - sr_a.first) / speed_of_light
+            return delay
```
bhawkins, line 1440: The spec technically allows having different zero-padding on A and B, compensated by different validSampleSubSwath annotations, so ideally we'd compare the first _valid_ slantRange in A and B. However, I think that case is pretty obscure and it's fine to just document the assumption with a comment.

python/packages/nisar/antenna/rx_channel_imbalance_helpers.py
```diff
@@ -0,0 +1,489 @@
...
+@dataclass(frozen=True)
+class RxChannelImbalanceProduct:
+    """
+    RX channel imbalance product extracted from LNA/CALTONE ratio
+    for a certain frequency band and polarization.
+
+    Attributes
+    ----------
+    lna_caltone_ratio: np.ndarray(complex)
+        Peak-normalized complex LNA/CALTONE ratio over all RXs
```
bhawkins:
```suggestion
        Peak-normalized complex LNA/CALTONE ratio over all RX channels (typically 12)
```
A few extra words would help distinguish "receive channels" from "received echoes."

python/packages/nisar/antenna/rx_channel_imbalance_helpers.py
```diff
@@ -0,0 +1,489 @@
...
+    Compute 12 complex RX channel imbalance based on LNA/CALTONE ratio
+    for a desired frequency band and TR polarization.
+
+    Also report the dominant tap number our of 3 for LNA three-tap
+    correlator as well as detected relative time delays for all RX channels
+    for debugging purposes.
+
+    Parameters
+    ----------
+    aw : Raw
```
bhawkins:
```suggestion
    raw : Raw
```

python/packages/nisar/antenna/rx_channel_imbalance_helpers.py
```diff
@@ -0,0 +1,489 @@
...
+
+def compute_rx_channel_imbalance(
+        raw: Raw,
+        freq_band: str,
+        txrx_pol: str,
+        caltone_freq: float | None = None,
+) -> Tuple[np.ndarray, np.ndarray, np.ndarray, float]:
+    """
+    Compute 12 complex RX channel imbalance based on LNA/CALTONE ratio
+    for a desired frequency band and TR polarization.
```
bhawkins:
```suggestion
    Compute 12 complex RX channel imbalance based on LNA/CALTONE ratio
    for a desired frequency band and TR polarization, averaged over all
    records in the raw data file.
```

I think it's worth pointing out this is averaged over the whole L0B.  (By the way, that seems like a smart choice in order to avoid discontinuities at frame boundaries, for example.)

python/packages/nisar/antenna/rx_channel_imbalance_helpers.py
```diff
@@ -0,0 +1,489 @@
...
+        1-D array of size equal to number of channels representing
+        the mean across range lines.
+
+    """
+    # or simply np.nanmean(data, axis=0)
+    d = np.sort(np.abs(data), axis=0)
+    q1_all, q3_all = np.percentile(d, q=[perc, 100 - perc], axis=0)
+    mean_all = []
+    for cc, (q1, q3) in enumerate(zip(q1_all, q3_all)):
+        q1, q3 = np.sort([q1, q3])
```
bhawkins, line 466:
```suggestion
```
Aren't these already sorted, by definition?

python/packages/nisar/antenna/rx_channel_imbalance_helpers.py
```diff
@@ -0,0 +1,489 @@
...
+        lna_mean_tap3[nn] = _mean_2d(lna_cal)
+    # get dominat taps
+    abs_lna_mean_tap3 = abs(lna_mean_tap3)
+    idx_lna_taps = np.nanargmax(abs_lna_mean_tap3, axis=0)
+    amp_lna_mean = np.zeros(n_rxs)
+    for nn in range(n_rxs):
+        amp_lna_mean[nn] = abs_lna_mean_tap3[idx_lna_taps[nn], nn]
+    _check_if_zero(amp_lna_mean, msg=f'{txrx_pol[0]}-pol LNA Cal')
+    # get the phase part at a fixed common tap rather than dominant one
+    phs_lna_mean = np.angle(lna_mean_tap3[1])
```
bhawkins, line 237: It seems kinda strange to have different assumptions for amplitude and phase. Since I assume the peak usually will be at index 1 anyway, I guess I'm not really concerned about it, though.

python/packages/nisar/antenna/rx_channel_imbalance_helpers.py
```diff
@@ -0,0 +1,489 @@
...
+    Returns
+    -------
+    bool
+        True if the product band/pol represent the second frequency band
+        in split-spectrum case, otherwise false.
+    """
+    if freq_band == "B" and len(raw.frequencies) == 2:
+        if txrx_pol in raw.polarizations['A']:
+            return True
+    return False
```
bhawkins:
```suggestion
    return (freq_band == "B" and len(raw.frequencies) == 2
        and txrx_pol in raw.polarizations['A'])
```
Python guarantees short-circuit logic, so it's safe to write as a single expression.

python/packages/nisar/antenna/rx_channel_imbalance_helpers.py
```diff
@@ -0,0 +1,489 @@
...
+        caltone_freq = caltone_frequency_from_raw(raw, txrx_pol)
+        warn(f'Caltone frequency is extracted from {txrx_pol[1]}-pol DRT '
+             f'-> {caltone_freq * 1e-6:.3f} (MHz)')
+    # Check if product from the second band so we can modify
+    # the results from the first band only if there is a
+    # relative delay offset in one of qFSP vs others, that is
+    # one of the qFSP is an outlier due to  ADC clock/delay issue
+    # check if there is delay anomaly among three qFSP
+    if _is_product_from_second_band(raw, freq_band, txrx_pol):
+        warn(f'correcting LNA/CALTONE for band={freq_band} and pol={txrx_pol}')
```
bhawkins: At least to me, it doesn't seem like we should issue runtime warnings for fully-expected behavior. I'd prefer to just use the logging module for this.

python/packages/nisar/antenna/rx_channel_imbalance_helpers.py
```diff
@@ -0,0 +1,489 @@
...
+        1-D array of relative time delays of three qFSP (12 RX channels)
+        in seconds.
+
+    """
+    # Get caltone frequency from DRT if not provided
+    if caltone_freq is None:
+        caltone_freq = caltone_frequency_from_raw(raw, txrx_pol)
+        warn(f'Caltone frequency is extracted from {txrx_pol[1]}-pol DRT '
+             f'-> {caltone_freq * 1e-6:.3f} (MHz)')
+    # Check if product from the second band so we can modify
```
bhawkins:
```suggestion
    # Loopback cal is only ever measured on one sub-band.
    # Check if product from the second band so we can modify
```
An extra sentence of context is helpful.

python/packages/nisar/antenna/rx_channel_imbalance_helpers.py
```diff
@@ -0,0 +1,489 @@
...
+    return the time delays for a qFSP with phase anomaly.
+    Else, return zeros.
+
+    Parameters
+    ----------
+    lna_caltone_ratio : np.ndarray(complex)
+        1-D array of complex LNA/CALTONE ratio with size equals
+        to the number of RX channels
+    dif_chirp_caltone_freq: float
+        Frequency difference between chirp and caltone in Hz.
```
bhawkins:
```suggestion
        Frequency difference between chirp and caltone (fc - f_cal) in Hz.
```
It'd be nice to make sign as explicit as possible. It also might be worth noting that the algorithm doesn't work for fc==fcal, which Shaffer has floated as a possibility (in an only half-serious way).

python/packages/nisar/noise/noise_estimation_from_raw.py
```diff
@@ -456,7 +462,11 @@ def est_noise_power_from_raw(
                 f'Number of range blocks is smaller than min {n_rg_blk_min}'
             )
     logger.info(f'Number of range blocks -> {num_rng_block}')
-
+    # check if product is quad
+    is_quad_pol = is_raw_quad_pol(raw)
+    if is_quad_pol:
```
bhawkins, line 467:
```suggestion
    if is_raw_quad_pol(raw):
```
I don't see this used elsewhere, so need to store the result.

python/packages/nisar/noise/noise_estimation_from_raw.py
```diff
@@ -575,12 +588,20 @@ def est_noise_power_from_raw(
                 # calculate approximate ENBW for relatively white noise!
                 enbw = enbw_from_raw(raw, freq_band, txrx_pol[0])
                 logger.info(f'Approximate ENBW in (MHz) -> {enbw * 1e-6}')
+                # check if quad pol per telemetry then use the oppsoite
```
bhawkins:
```suggestion
                # check if quad pol per telemetry then use the opposite
```

python/packages/nisar/antenna/beamformer.py
```diff
@@ -609,16 +615,27 @@ def form_pattern(self, pulse_time, slant_range, channel_adj_factors=None):
...
+            if len(channel_adj_factors) == num_chanl:
+                # check if the correction factor is zero for all
+                # active channels
+                cor_fact = np.asarray(channel_adj_factors)[
+                    self.active_channel_idx]
+                if np.isclose(abs(cor_fact).max(), 0):
+                    raise ValueError('"channel_adj_factors" are zeros for all '
+                                     'active RX channels!')
+                rx_wgt *= cor_fact[:, None]
+            elif len(channel_adj_factors) == 1:  # fixed scalar
```
bhawkins, line 627: Not sure I understand the use case for a scalar `channel_adj_factors`.

python/packages/nisar/antenna/pattern.py
```diff
@@ -205,14 +236,31 @@ def __init__(self, raw: Raw, dem: DEMInterpolator,
         # Loop over all freqs & pols since some RX pols may be found only on
         # freq B (e.q. the QQP case).  Assume RD/WD/WL are the same for all
```
bhawkins:
```suggestion
        # Loop over all pols. Assume RD/WD/WL are the same for all
```
This comment is outdated after making the class specific to a single sub-band.

python/packages/nisar/antenna/pattern.py
```diff
@@ -205,14 +236,31 @@ def __init__(self, raw: Raw, dem: DEMInterpolator,
...
+            # correct RDs @ `self.fs_win` used in forming RX DBF pattern.
+            # This will account for delay after onboard DBF due to
+            # both the first pulsewidth in sequential TX chirps
+            # as well as the difference in filter group delays.
+            tm_delay = range_delay_sequential_tx_from_raw(
+                raw, self.freq_band, txrx_pol)
+            tm_delay += delay_ofs_dbf
+            n_samp_delay = round(tm_delay * self.fs_win)
+            if n_samp_delay != 0:
+                warn(
```
bhawkins:
```suggestion
                log.info(
```
This is expected behavior and shouldn't emit a warning IMO.

python/packages/nisar/workflows/focus.py
```diff
@@ -1909,9 +1916,14 @@ def temp(suffix):
...
             if cfg.processing.is_enabled.eap:
+                # XXX Due to a bug in respective DRT of some L0B products
+                # (CRID=05007), caltone.frequency is extrated from runconfig
+                # otherwise, it shall be set to None to be determined from DRT!
                 antpat = AntennaPattern(raw, dem, antparser,
                                         instparser, orbit, attitude,
-                                        el_lut=el_lut)
+                                        el_lut=el_lut,
+                                        freq_band=frequency,
+                                        caltone_freq=cfg.processing.caltone.frequency)
```
bhawkins: I wonder if we should make the DBF time offset parameter configurable, maybe via the INT_CAL file?

bhawkins (commented):
Wow! You've done an admirable job sorting out all the vagaries of the loopback cal data. Honestly, I'm not sure I understand every detail and might've made some stupid comments as a result.

I ended up writing down a lot of comments while reading through the changes, but I think there's only one that I consider a blocker. It's the one to do with the crosspol indexing, and I'll go add some emoji to make it stand out. It may just be a misunderstanding on my part.

python/packages/nisar/products/readers/Raw/Raw.py
bhawkins: 🏁 🏁 🏁 This is the one that most worries me 🏁 🏁 🏁

rad-eng-59: Yes we do and  it is happening but in a bit hidden way. Note that for the cross-pol, the TX polarization gets flipped back to the one that being transmitted first via recursive call to the function and then simply half of the array is being eventually used with strides of `2`. I have tested it over Qaud-pol w/o any issue.

python/packages/nisar/products/readers/Raw/Raw.py
rad-eng-59: Clarify the assumption behind this logic

python/packages/nisar/workflows/focus.py
rad-eng-59: Add  `delay_ofs_dbf=-2.1474e-6` and put some notes here

python/packages/nisar/products/readers/Raw/Raw.py
rad-eng-59: The intention was to handle those old and/or simulated NISAR products and test cases. This behavior is also documented. I'll add more for clarification of the intention.

python/packages/nisar/products/readers/Raw/Raw.py
rad-eng-59: The explicit None is for clarity and readability rather than using implicit None return for such scenario where we have different return type. This doesn't make it non-pythonic given this is exceptional scenario where None is returned under the special circumstance, in this case, for missing filed value in L0B. So, I think its existence is more help than harm.

python/packages/nisar/products/readers/Raw/Raw.py
rad-eng-59: ditto (see previous rational)

python/packages/nisar/products/readers/Raw/Raw.py
rad-eng-59: Already discussed the rationals behind this decision in a meeting. Assumed resolved unless otherwise.

python/packages/nisar/products/readers/Raw/Raw.py
rad-eng-59: We are modifying what's being parsed from input L0B because the content is partially incorrect. Eventually, we need to update L0B spec, processor and algorithm to get us the right chirp values and cal types directly in L0B products under `swaths` by taking into account exceptions and complexity of the L-band firmware.

python/packages/nisar/products/readers/Raw/Raw.py
rad-eng-59: Already discussed the reason in a meeting on why we are using external helper functions rather than methods as part of raw class.

python/packages/nisar/antenna/rx_channel_imbalance_helpers.py
rad-eng-59: No, not if the input is not sorted in ascending order.

python/packages/nisar/antenna/rx_channel_imbalance_helpers.py
rad-eng-59: We have to be very careful on how to get the residual channel imbalance from three-tap correlator that goes through various paths.
That assumption about second index being dominant is not always correct. The peak can be either of taps. It occurs in several cases I have analyzed. 
For magnitude part of imbalance we care about getting the max among all three taps for each channel because they can have different delay which is irrelevant to the gain imbalance.
On the other hand,, the phase imbalance we are after should corresponds to the same index over all channels of the same qFSP whether that index is the peak or not to get the phase for the same exact lag in chirp correlations that share the same chirp reference but go through different paths. Here the delay difference among channels will be captured by the same index as part of phase imbalance.

python/packages/nisar/antenna/rx_channel_imbalance_helpers.py
rad-eng-59: It actually works perfectly for band='A'.
If the center frequency of chirp and caltone is the same then band=`A` does NOT qFSP issue, no SNR loss, no InSAR phase artifacts because Caltone corrects the phase at center frequency of the TX chirp. Now, frequency band B won't be corrected because, we do not have any measurement of band "B". This is result of missing capability in the instrument not the faulty algorithm. 
If Shaffer has a more robust algorithm, please share it.

python/packages/nisar/noise/noise_estimation_from_raw.py
rad-eng-59: That is not true. `is_quad_pol` is used again at line `565`

python/packages/nisar/products/readers/Raw/Raw.py
rad-eng-59: We need to be very careful or we will break exiting test cases and workflows relying on old L0B products. This so-called `None` is documented in CMTT, btw. I didn't make this up.  Please check it out.

python/packages/nisar/antenna/beamformer.py
rad-eng-59: A fixed overall scalar for all channels rather than repeating every single value in an array.

python/packages/nisar/products/readers/Raw/Raw.py
rad-eng-59: Renamed the respective function to `opposite_linear_pol` to handle bad values like this.

python/packages/nisar/products/readers/Raw/Raw.py
rad-eng-59: Already explained the rational in the meeting. let's talk if still confusing. 
Fixed a potential bug and add notes for clarity

python/packages/nisar/products/readers/Raw/Raw.py
rad-eng-59: Already handle this by renaming the function to `opposite_linear_pol`

python/packages/nisar/products/readers/Raw/Raw.py
rad-eng-59: I have fixed the bug.

python/packages/nisar/antenna/beamformer.py
bhawkins: Yeah, I just mean we already have two other ways to get a constant scaling into the processor, which is arguably too many already.

python/packages/nisar/antenna/rx_channel_imbalance_helpers.py
bhawkins: The first line of the function sorts the data, though.

python/packages/nisar/products/readers/Raw/Raw.py
bhawkins: I understand the intent. My point is that requiring the user to check return values is error-prone and not Pythonic. A mandatory user-side if/else block on the return value is worse than a try/catch block since neglecting the former could cause all sorts of weird behavior, while neglecting the latter halts the program immediately. I'd agree that it might be nice to raise a `NotImplementedError` here with a nice message rather than the h5py `KeyError`. Anyways, I don't consider this a blocking issue.

python/packages/nisar/products/readers/Raw/Raw.py
bhawkins: Isn't the Pythonic way to handle an "exceptional scenario" an... Exception? 😆

python/packages/nisar/products/readers/Raw/Raw.py
bhawkins: I should've spelled this out more clearly. I understand that "none" is a documented polarization type in the CMTT. However, with this implementation, there are now two cases that return `PolarizationTypeId.none`: 1) the L0B data actually says it's "none", and 2) the code failed to parse the telemetry (e.g., old simulations). Your assumption seems to be that case 2 has already returned because other telemetry fields should also be missing.

However, there's an edge case where somebody adds the chirp correlator telemetry to the simulator but forgets to add the polarization type telemetry. This code never detects that case and may produce incorrect results. So we should either change this function to detect the edge case explicitly per the suggestion above (as you seem to advocate), or change polarization_type_from_drt to throw an exception when it can't do its job.

python/packages/nisar/products/readers/Raw/Raw.py
rad-eng-59: My hands are tight here because of other external factors such as exiting ISCE3 test sample files w/ old formats, broken L0B spec, etc.  I had to compromise to avoid a bunch of failures.  Also, None is a valid return value not an exception as described in the document. I'm a big fan of try and exception but that won't here.  Try your approach and  see how things will fall apart.

python/packages/nisar/products/readers/Raw/Raw.py
bhawkins: Yeah, as we discussed, `chp_cor_x` is already decimated by two since it's the result of a call to `chirpcorrelator_caltype_from_raw`.

python/packages/nisar/products/readers/Raw/Raw.py
rad-eng-59: Nope, because `None` is a valid return to handle special circumstances as explained in the docstring. Perhaps, there are `numpy`/`scipy` functions that return two different types of output depending on the input.

python/packages/nisar/products/readers/Raw/Raw.py
bhawkins: Just to document the discussion, the main reasons I remember are schedule pressure and the difficulty in coordinating last-minute API changes with L0B development that also has to use this code. I think, in principle, there's no strong objection to refactoring in the future if somebody wants to take that on.

bhawkins (approved):
Thanks for the quick revisions! The thing I was worried about turned out to be a red herring. For the sake of completeness, I documented a few minor gripes, but they shouldn't be considered blockers. LGTM.

python/packages/nisar/products/readers/Raw/Raw.py
bhawkins: Let's consider two designs from the user perspective.

Design A:
```
cc = _parse_chirpcorrelator_from_hrt_qfsp(txrx_pol)
if cc is None:
    # handle L0Bs with missing telemetry
```
Design B:
```
try:
    cc = _parse_chirpcorrelator_from_hrt_qfsp(txrx_pol)
except NotImplementedError:
    # handle L0Bs with missing telemetry
```
How does Design A handle "test sample files w/ old formats, broken L0B spec, etc." better than B? If you're saying that our unit tests or end-to-end tests won't pass with Design B, then it sounds like we have a lot of places that are missing the necessary logic to avoid incorrect behavior. I actually doubt that's the case since this is a new function, and all the usages should be contained in this PR.

python/packages/nisar/products/readers/Raw/Raw.py
rad-eng-59: Not all these modules can be part of Raw because it requires fundamental  design change of L0B spec (that ship has already sailed), change in all exiting L0B products, updates of isce3 sample files and test suites, breaking external tools that rely on Raw, etc. Besides, there exists a big difference in function arguments/interface. Trust me, I did tried to incorporate them into Raw as it seemed easier at the first glance until I did see the main blockers while implementing it.    Then, I have decided to introduce these helper functions inside raw module given there is no harm to define helper functions that relies on Raw parser as supplements rather than use them as friendly or actual methods of the raw without breaking the interface.
Ideally, they can be part of Raw class if we redesign L0B and update the products.

python/packages/nisar/antenna/beamformer.py
rad-eng-59: I believe there is another reason related to a isce3 test suite and sample product that breaks w/o this conditional block.

python/packages/nisar/products/readers/Raw/Raw.py
rad-eng-59: No point of arguing here. We need an offline meeting for this.

python/packages/nisar/products/readers/Raw/Raw.py
rad-eng-59: We can issue a warning but NOT raise an exception as it will break under existing L0B products. If users provide **NISAR L-band** L0B product that doesn't meet the spec then several unexpected things can happen. This function doesn't check and verify the L0B spec. I believe none of our functions do under ISCE3.

hfattahi:
We reviewed this PR and agreed that in few places where a function returns None, perhaps the better approach is to raise an exception.

rad-eng-59:
> We reviewed this PR and agreed that in few places where a function returns None, perhaps the better approach is to raise an exception.

Well, there were two flavors discussed in the meeting. I have gone with the latter one which requires minimum update w/o breaking any code while preserve consistency among functions with handling of missing datasets in L0B products. I have tested it over a short QP and DP. Let me know if there is still objection so I can provide the correct updates to Alice in regards to updating L0B process. Note that I'm trying to keep two separate repos, this one and [external one](https://github.jpl.nasa.gov/NISAR-L0B-STUFF/l0b_qfsp_issue/tree/l0b-qfsp), consistent and synced.

python/packages/nisar/products/readers/Raw/Raw.py
rad-eng-59: I have just updated the polarization ID to return None rather than `6` when the respective dataset is missing in old or partial NISAR-like L0B product. This none is different than CMTT none.

python/packages/nisar/products/readers/Raw/Raw.py
```diff
@@ -918,3 +1080,310 @@ def open_rrsd(filename) -> RawBase:
...
+    raw : nisar.products.readers.Raw
+        L0B raw parser object
+
+    Returns
+    --------
+    bool
+        True if the L0B product is quad pol otherwise False.
+
+    """
+    return polarization_type_from_raw(raw) == PolarizationTypeId.quad
```
bhawkins:
```suggestion
    poltype = polarization_type_from_raw(raw)
    if poltype is None:
        raise ValueError("L0B does not contain polarization telemetry "
            "so cannot determine whether it is quad pol")
    return poltype
```
When there's a special return value, not checking for it is almost always a bug. Raising an exception here is another way to address the edge case I identified in `chirpcorrelator_caltype_from_raw`.

rad-eng-59: This doesn't make sense here. This function return a boolean not a pol type to decide whether it is quad pol or not.   If the respective field  is missing in L0B, then the return pol type value will be None and this logic leads to False, that is it won't be quad pol. The warn message already issued due to nonexistent key.

bhawkins: I understand the logic. IMO it's incorrect behavior for `is_raw_quad_pol` to return False when the telemetry is missing. It could still be QP (for example, there could be four images in the file). "I don't know" isn't the same as "I know it's false."

Next you're going to say, okay fine, let's have this function return `None` to indicate that. And then I'll say okay great, now we have to check for `None` in `chirpcorrelator_caltype_from_raw` and we'll finally have correct C-style error handling in Python. 😜

rad-eng-59: I have just replaced return with raise and guess what, isce3 tests like the noise estimator (NET) , focus.py, fail, thanks to simulated and/or old L0B products. If the warning message for missing field along with the dcostring of the function are not clear to the user then I am not sure how propagating exception over several module will be more useful for the user whom runs functions blindfoldly. Any suggestion?

bhawkins: Great! I think it's good when the code issues an error when there's a mismatch between the logic and the input data. I'll build this PR with the suggestion above, run the tests, and see if I can come up with a good suggestion for any failures.

rad-eng-59: I just did some changes to accommodate your suggestions except for this function with a reasonable exception.. Let me test it  and upload the commit and see if you like it or not. Give me a few minutes plz.

rad-eng-59:
@bhawkins, @hfattahi  plz check out the new commit and let me know what you think.

python/packages/nisar/products/readers/Raw/Raw.py
bhawkins: LOL I stepped away for dinner and you'd already changed everything by the time I got back. You're quick!

bhawkins (approved):
LGTM thanks!

hfattahi (approved):
LGTM. Thanks @rad-eng-59 for this important addition. I have tested InSAR workflow with RSLCs generated with this PR and confirm the significant improvement in phase and offsets between pairs acquired with different sampling characteristics.

rad-eng-59:
@bhawkins , @hfattahi , sorry for the last minute update. I have simply added three lines to better support old or sim L0B products including the edge case  for determining quad pol when its respective field/id is missing in L0B . Please check out the new commit and let me know if you both are okay with that prior to merging. Thanks

bhawkins (approved):
The latest tweak looks fine to me.

python/packages/nisar/workflows/focus.py
```diff
@@ -1909,9 +1916,22 @@ def temp(suffix):
...
+                # flexible configuration over wide range of L0B products.
+                # XXX the intrument-related delay offset used in DBF process
+                # shall be eventually obtained from instrument INT CAL HDF5
+                # once the respective product spec is updated (delay_ofs_dbf)!
+                # The default value is suitable for NISAR L-band instrument.
                 antpat = AntennaPattern(raw, dem, antparser,
                                         instparser, orbit, attitude,
-                                        el_lut=el_lut)
+                                        el_lut=el_lut,
+                                        freq_band=frequency,
```
bhawkins, line 1932:
```suggestion
                                        freq_band=channel_in.freq_id,
```
I'm thinking we want to use the frequency of the input data here, right? The `frequency` variable here is the output frequency label (`channel_out.freq_id`), which may be different in the case of QD or mixed mode 77 MHz & a 5 MHz sideband.

bhawkins: Implemented in #244

PR #234, opened 2026-03-12
Fix unwrapping error in differential interferogram between A and B

oberonia78:
**Issue**

Phase jumps were observed in the ionosphere layer due to unwrapping errors (reported by Marin).
image

**Root Cause**
image

- During the computation of the differential interferogram between frequency A and B, the phase difference is directly calculated between the two bands without mask.
- During the unwrapping step, the subswath mask from frequency B only is applied to remove partially focused regions (in `main_diff_ms_band` method and when preprocessing is enabled in `unwrap.py`).
- In some areas, the locations of partially focused regions differ between frequency A and B. As a result, pixels that are invalid in frequency A but valid in frequency B remain in the differential interferogram. These residual pixels introduce phase inconsistencies that propagate into the unwrapping step, leading to phase jumps (above figure).

**Solution**
image
- When computing differential interferogram between A and B, we read both subswaths from A and B and mask out the partially focused areas. 

**Updated Ionosphere**
image

RSLCs used for this analysis
reference_rslc_file: NISAR_L1_PR_RSLC_006_148_A_022_2005_QPDH_A_20251202T105655_20251202T105730_X05009_N_F_J_001.h5
secondary_rslc_file: NISAR_L1_PR_RSLC_007_148_A_022_2005_QPDH_A_20251214T105655_20251214T105730_X05009_N_F_J_001.h5

### DIFF
--- a/python/packages/nisar/workflows/ionosphere.py
+++ b/python/packages/nisar/workflows/ionosphere.py
@@ -540,7 +540,9 @@ def compute_differential_phase(
                     first_data_block = _to_complex_if_needed(first_data_block)
                     second_data_block = _to_complex_if_needed(second_data_block)
 
-                    # Optional resampling of FIRST to SECOND grid
+                    # Resample first dataset (frequency A / low subband) to
+                    # match the grid of the second dataset (frequency B / high subband).
+                    # This ensures pixel-wise alignment before phase differencing.
                     if resampling_flag:
                         first_data_block = decimate_freq_a_array(
                             main_slant,
@@ -553,6 +555,17 @@ def compute_differential_phase(
                                 first_mask_block)
 
                     if subswath_mask_enabled:
+                        # Interpretation of input datasets depends on the
+                        # processing context:
+                        # - For main_diff_ms_band:
+                        #     first  → frequency A
+                        #     second → frequency B
+                        # - For main_diff_low_high_subband:
+                        #     first  → low subband
+                        #     second → high subband
+                        # Each mask provides:
+                        #   - reference_valid: valid pixels in reference region
+                        #   - secondary_valid: valid pixels in secondary region
                         first_reference_valid, first_secondary_valid, _ = \
                             interpret_subswath_mask(first_mask_block)
                         second_reference_valid, second_secondary_valid, _ = \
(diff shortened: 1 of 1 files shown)


python/packages/nisar/workflows/ionosphere.py
```diff
@@ -505,9 +546,31 @@ def _to_complex_if_needed(arr):
...
+                            (~second_reference_valid) |
+                            (~second_secondary_valid)
+                        )
+                    else:
+                        invalid = None
 
                     # Compute the differential phase
                     diff_phase = first_data_block * np.conj(second_data_block)
+                    if invalid is not None:
+                        diff_phase = diff_phase.copy()
```
hfattahi: Why is this copy needed?

python/packages/nisar/workflows/ionosphere.py
```diff
@@ -505,9 +546,31 @@ def _to_complex_if_needed(arr):
...
+                            first_mask_block = decimate_freq_a_array(
+                                main_slant,
+                                side_slant,
+                                first_mask_block)
+
+                    if subswath_mask_enabled:
+                        first_reference_valid, first_secondary_valid, _ = \
+                            interpret_subswath_mask(first_mask_block)
+                        second_reference_valid, second_secondary_valid, _ = \
+                            interpret_subswath_mask(second_mask_block)
```
hfattahi, line 572: I assume first_reference_valid and first_secondary_valid are the frequency A reference and secondary subswath valid mask and second_reference_valid, second_secondary_valid are the freqB subswath for reference and secomdary acquisitions respectively. Better naming would be helpful. Or at least put a comment and clarify.

hfattahi (commented):
Thanks @oberonia78 for the PR. It looks good to me. Please see my minor comment below.

python/packages/nisar/workflows/ionosphere.py
oberonia78: Thanks @hfattahi Depending on how this function is used, their meaning could be different. For the `main_diff_ms_band`, your assumption is correct. However, they could be `low_subband` and `high_subband` for the `main_diff_low_high_subband` method. I'll add more better comment.

hfattahi: Oh I see. Thanks for the explanation.

python/packages/nisar/workflows/ionosphere.py
oberonia78: No. It is not needed. I've just deleted.

hfattahi (approved):
LGTM. Thanks @oberonia78

Tyler-g-hudson (commented):
LGTM!

Tyler-g-hudson (approved):

PR #235, opened 2026-03-15
RFI Detection and Mitigation for Dithered PRF mode

hb6688:
This PR makes the following changes:

1. Separately computes sample covariance matrix for dithered-PRF and constant-PRF modes of operations
2. Skip threshold blocks when there are not enough valid samples for dithered-PRF mode

### DIFF
--- a/python/packages/isce3/signal/compute_evd_cpi.py
+++ b/python/packages/isce3/signal/compute_evd_cpi.py
@@ -102,9 +102,9 @@ def compute_evd_tb(
     mask_valid : np.ndarray bool, [num_pulses x num_rng_samples], optional
         Valid-sample mask with same shape as raw_data. Required if
         prf_dither_mode=True.
-    off_diag_overlap_ratio : float, optional
+    off_diag_overlap_ratio : float, optional, default = 0.25
         Minimum overlap ratio used by gap exclusion covariance estimation
-    diag_valid_ratio : float, optional
+    diag_valid_ratio : float, optional, default = 0.20
         Minimum fraction of valid samples required to compute a diagonal term
         in the sample covariance matrix entry R_ii.
     min_ev_valid_idx: int, optional
@@ -220,10 +220,10 @@ def compute_evd(
         True indicates valid samples. False indicates invalid samples or gaps.
         This is used only when prf_dither_mode=True.
         If None, an all-True mask is created.
-    off_diag_overlap_ratio : float, default=0.1
+    off_diag_overlap_ratio : float, default=0.25
         Minimum fraction of overlapping valid range samples required to compute
         an off-diagonal covariance term R_ij when prf_dither_mode=True.
-    diag_valid_ratio : float, default=0.05
+    diag_valid_ratio : float, default=0.20
         Minimum fraction of valid range samples required to compute
         a diagonal covariance term R_ii when prf_dither_mode=True.
 
@@ -286,10 +286,10 @@ def compute_gap_exclusion_cov(
     mask_valid_cpi: (num_pulses, num_rng_samples) bool array, optional
         True indicates valid samples. False indicates invalid samples or gaps.
         If None, an all true boolean mask is created. All samples are assumed to be valid.
-    off_diag_overlap_ratio: float
+    off_diag_overlap_ratio: float, default = 0.25
         Minimum fraction of overlapping valid range samples required to compute
         an off-diagonal term in the sample covariance matrix entry R_ij.
-    diag_valid_ratio : float, optional
+    diag_valid_ratio : float, optional, deffault = 0.20
         Minimum fraction of valid samples required to compute a diagonal term in the
         sample covariance matrix entry R_ii.
         
--- a/python/packages/isce3/signal/compute_evd_cpi.py
+++ b/python/packages/isce3/signal/compute_evd_cpi.py
@@ -78,7 +78,6 @@ def compute_evd_tb(
     raw_data: np.ndarray,
     *,
     cpi_len: int=16,
-    apply_gap_exclusion: bool=False,
     mask_valid: np.ndarray=None,
     off_diag_overlap_ratio: float=0.25,
     diag_valid_ratio: float=0.20,
@@ -98,13 +97,10 @@ def compute_evd_tb(
         Raw data to be processed
     cpi_len: int, optional
         Number of slow-time pulses within a CPI, default=16
-    apply_gap_exclusion: bool, optional, default = True
-        If True, CPI sample covariance matrix will be computed differently by excluding the
-        invalid data gaps.
     mask_valid : np.ndarray bool, [num_pulses x num_rng_samples], optional
-        Valid-sample mask with same shape as raw_data.
-        if apply_gap_exclusion is False, then mask_valid will be ignored, and standard
-        sample covariance matrix will be performed.
+        Valid-sample mask with same shape as raw_data.  If provided, CPI sample
+        covariance matrix will be computed differently by excluding the invalid
+        data gaps.
     off_diag_overlap_ratio : float, optional, default = 0.25
         Minimum overlap ratio used by gap exclusion covariance estimation
     diag_valid_ratio : float, optional, default = 0.20
@@ -162,12 +158,8 @@ def compute_evd_tb(
             "Since Python uses 0-based indexing, the maximum valid index is cpi_len - 1."
         )
 
-    # Verify that mask_valid is provided if apply_gap_exclusion is True
-    if apply_gap_exclusion and mask_valid is None:
-        raise ValueError("mask_valid must be provided if apply_gap_exclusion is True")
-
     # Verify TB Mask shape
-    if mask_valid.shape != raw_data.shape:
+    if (mask_valid is not None) and (mask_valid.shape != raw_data.shape):
         raise ValueError(f"Valid TB mask shape {mask_valid.shape} != TB data shape {raw_data.shape}")
     
     # Output Eigenvalues and Eigenvectors
@@ -181,10 +173,9 @@ def compute_evd_tb(
         slice_gen(num_pulses, cpi_len, combine_rem=False)
     ):
         data_cpi = raw_data[cpi_slow_time]
-        mask_valid_cpi = mask_valid[cpi_slow_time]
+        mask_valid_cpi = None if mask_valid is None else mask_valid[cpi_slow_time]
         eig_val_sort, eig_vec_sort = compute_evd(
             data_cpi,
-            apply_gap_exclusion=apply_gap_exclusion,
             mask_valid_cpi=mask_valid_cpi,
             off_diag_overlap_ratio=off_diag_overlap_ratio,
             diag_valid_ratio=diag_valid_ratio,
@@ -206,7 +197,6 @@ def compute_evd_tb(
 def compute_evd(
     raw_data: np.ndarray,
     *,
-    apply_gap_exclusion: bool = False,
     mask_valid_cpi: np.ndarray = None,
     off_diag_overlap_ratio: float = 0.25,
     diag_valid_ratio: float = 0.20,
@@ -217,19 +207,16 @@ def compute_evd(
     ----------
     raw_data : array-like complex [num_pulses x num_rng_samples]
         Raw data to be processed.
-    apply_gap_exclusion : bool, default=False
-        If True, CPI sample covariance matrix will be computed differently by excluding the
-        invalid data gaps.
     mask_valid_cpi : (num_pulses, num_rng_samples) bool array, optional
         True indicates valid samples. False indicates invalid samples or gaps.
-        This is used only when apply_gap_exclusion=True.
-        If None, an all-True mask is created.
+        If provided, CPI sample covariance matrix will be computed differently
+        by excluding the invalid data gaps.
     off_diag_overlap_ratio : float, default=0.25
         Minimum fraction of overlapping valid range samples required to compute
-        an off-diagonal covariance term R_ij if apply_gap_exclusion is True.
+        an off-diagonal covariance term R_ij if mask is provided.
     diag_valid_ratio : float, default=0.20
         Minimum fraction of valid range samples required to compute
-        a diagonal covariance term R_ii when apply_gap_exclusion is True.
+        a diagonal covariance term R_ii when mask is provided.
 
     Returns
     -------
@@ -247,10 +234,7 @@ def compute_evd(
     # Application in Narrow-Band Interference Suppression for SAR”, IEEE Geoscience 
     # and Remote Sensing Letters, vol. 4, no. 1, pp. 76,2007.
 
-    if apply_gap_exclusion:
-        if mask_valid_cpi is None:
-            raise ValueError("mask_valid_cpi must be provided when apply_gap_exclusion is True.")
-
+    if mask_valid_cpi is not None:
         mask_valid_cpi = mask_valid_cpi.astype(bool, copy=False)
 
         if mask_valid_cpi.shape != raw_data.shape:
--- a/python/packages/isce3/signal/compute_evd_cpi.py
+++ b/python/packages/isce3/signal/compute_evd_cpi.py
@@ -47,7 +47,6 @@ def slice_gen(total_size: int, batch_size: int, combine_rem: bool=True) -> Itera
             stop_idx = start_idx + batch_size
             yield slice(start_idx, stop_idx)
 
-
 def eigen_decomp_sort(cov_matrix):
     """Perform Eigenvaule Decomposition of Sample Covariance Matrix which is assumed 
     to be Hermitian, and sort Eigenvalues in descending order.  Re-arrange column 
@@ -76,47 +75,13 @@ def eigen_decomp_sort(cov_matrix):
 
     return eig_val_sort, eig_vec_sort
 
-def compute_evd(
-    raw_data: np.ndarray,
-):
-    """Perform Eigenvalue Decomposition along axis 0.
-
-    Parameters
-    ------------
-    raw_data: array-like complex [num_pulses x num_rng_samples]
-        raw data to be processed
-
-    Returns
-    --------
-    eig_val_sort: 1D array of float, same length as the number of rows of input matrix
-        Eigenvalues sorted in descending order
-    eig_vec_sort: 2D array of complex, same shape as input matrix
-        column vector Eigenvectors sorted based on index of sorted Eigenvalues
-    """
-
-    num_rng_samples = raw_data.shape[1]
-
-    # The raw_data is not necessarily zero-mean when it is corrupted by RFI.
-    # If so, estimated sample covariance matrix cov_cpi should be called 
-    # sample correlation  matrix instead.  The reference below demonstrates
-    # this concept and notation.
-
-    # F. Zhou, R. Wu, M. Xing, and Z. Bao, “Eigensubspace-Based Filtering With 
-    # Application in Narrow-Band Interference Suppression for SAR”, IEEE Geoscience 
-    # and Remote Sensing Letters, vol. 4, no. 1, pp. 76,2007.
-
-    cov_cpi = np.matmul(raw_data, np.conj(raw_data).transpose()) / num_rng_samples
-    eig_val_sort, eig_vec_sort = eigen_decomp_sort(cov_cpi)
-
-    return eig_val_sort, eig_vec_sort
-
 def compute_evd_tb(
     raw_data: np.ndarray,
     cpi_len: int=16,
-    prf_dither_mode: bool=False,
+    prf_dither_mode: bool=True,
     mask_valid: np.ndarray=None,
-    off_diag_overlap_ratio: float=0.1,
-    diag_valid_ratio: float=0.05,
+    off_diag_overlap_ratio: float=0.25,
+    diag_valid_ratio: float=0.20,
     min_ev_valid_idx: int=10,
     rx_dynamic_range_db: float=-50.0,
 ):
@@ -133,7 +98,7 @@ def compute_evd_tb(
         Raw data to be processed
     cpi_len: int, optional
         Number of slow-time pulses within a CPI, default=16
-    prf_dither_mode: bool, optional
+    prf_dither_mode: bool, optional, default = True
         If True, use gap-aware covariance estimation
     mask_valid : np.ndarray bool, [num_pulses x num_rng_samples], optional
         Valid-sample mask with same shape as raw_data. Required if
@@ -179,14 +144,14 @@ def compute_evd_tb(
     if num_rng_samples < rng_samples_min:
         raise ValueError(
             "Minimum number of samples in a range block to estimate Sample Covariance"
-            f" Matrix is {rng_samples_min}! Current number of samples per range block"
-            f" is {num_rng_samples}!"
+            f" Matrix is {rng_samples_min} per Reed-Mallet-Brennan Rule."
+            f"Current number of samples per range block is {num_rng_samples}!"
         )
 
     # Verify Total number of pulses is greater than CPI length
     if num_pulses < cpi_len:
         raise ValueError(
-            f"Coherent Processing Interval length exceeds total number of pulses {num_pulses}!"
+            f"Coherent Processing Interval length of {cpi_len} exceeds total number of pulses {num_pulses}!"
         )
 
     if min_ev_valid_idx >= cpi_len:
@@ -198,6 +163,10 @@ def compute_evd_tb(
     if prf_dither_mode and mask_valid is None:
         raise ValueError("mask_valid must be provided when prf_dither_mode=True")
 
+    # Verify TB Mask shape
+    if mask_valid.shape != raw_data.shape:
+        raise ValueError(f"Valid TB mask shape {mask_valid.shape} != TB data shape {raw_data.shape}")
+    
     # Output Eigenvalues and Eigenvectors
     eig_val_sort_array = np.zeros([num_cpi, cpi_len], dtype="f4")
     eig_vec_sort_array = np.zeros((num_cpi, cpi_len, cpi_len), dtype="complex64")
@@ -209,17 +178,14 @@ def compute_evd_tb(
         slice_gen(num_pulses, cpi_len, combine_rem=False)
     ):
         data_cpi = raw_data[cpi_slow_time]
-
-        if not prf_dither_mode:
-            eig_val_sort, eig_vec_sort = compute_evd(data_cpi)
-        else:
-            mask_valid_cpi = mask_valid[cpi_slow_time]
-            eig_val_sort, eig_vec_sort = compute_evd_gap(
-                data_cpi,
-                mask_valid_cpi=mask_valid_cpi,
-                off_diag_overlap_ratio=off_diag_overlap_ratio,
-                diag_valid_ratio=diag_valid_ratio,
-            )
+        mask_valid_cpi = mask_valid[cpi_slow_time]
+        eig_val_sort, eig_vec_sort = compute_evd(
+            data_cpi,
+            prf_dither_mode=prf_dither_mode,
+            mask_valid_cpi=mask_valid_cpi,
+            off_diag_overlap_ratio=off_diag_overlap_ratio,
+            diag_valid_ratio=diag_valid_ratio,
+        )
 
         # Verify if the eigenvalue of CPI at index defind by  noise_ev_idx is meaningful
         eig_val_sort_abs = np.maximum(np.abs(eig_val_sort), 1e-30)
@@ -234,37 +200,41 @@ def compute_evd_tb(
 
     return eig_val_sort_array, eig_vec_sort_array, tb_is_valid
 
-def compute_evd_gap(
+def compute_evd(
     raw_data: np.ndarray,
     *,
+    prf_dither_mode: bool = True,
     mask_valid_cpi: np.ndarray = None,
-    off_diag_overlap_ratio: float = 0.1,
-    diag_valid_ratio: float = 0.05, 
+    off_diag_overlap_ratio: float = 0.25,
+    diag_valid_ratio: float = 0.20,
 ):
     """Perform Eigenvalue Decomposition along axis 0.
 
     Parameters
-    ------------
-    raw_data: array-like complex [num_pulses x num_rng_samples]
-        raw data to be processed
-    mask_valid_cpi: (num_pulses, num_rng_samples) bool array, optional
+    ----------
+    raw_data : array-like complex [num_pulses x num_rng_samples]
+        Raw data to be processed.
+    prf_dither_mode : bool, default=True
+        If True, use gap-exclusion covariance for dithered-PRF data.
+        If False, use the standard sample covariance matrix.
+    mask_valid_cpi : (num_pulses, num_rng_samples) bool array, optional
         True indicates valid samples. False indicates invalid samples or gaps.
-        If None, an all true boolean mask is created. All samples are assumed to be valid.
-    off_diag_overlap_ratio: float
+        This is used only when prf_dither_mode=True.
+        If None, an all-True mask is created.
+    off_diag_overlap_ratio : float, default=0.1
         Minimum fraction of overlapping valid range samples required to compute
-        an off-diagonal term in the sample covariance matrix entry R_ij.
-    diag_valid_ratio : float, optional
-        Minimum fraction of valid samples required to compute a diagonal term in the
-        sample covariance matrix entry R_ii.
+        an off-diagonal covariance term R_ij when prf_dither_mode=True.
+    diag_valid_ratio : float, default=0.05
+        Minimum fraction of valid range samples required to compute
+        a diagonal covariance term R_ii when prf_dither_mode=True.
 
     Returns
-    --------
-    eig_val_sort: 1D array of float, same length as the number of rows of input matrix
-        Eigenvalues sorted in descending order
-    eig_vec_sort: 2D array of complex, same shape as input matrix
-        column vector Eigenvectors sorted based on index of sorted Eigenvalues
+    -------
+    eig_val_sort : 1D array of float
+        Eigenvalues sorted in descending order.
+    eig_vec_sort : 2D array of complex
+        Column eigenvectors sorted to match eig_val_sort.
     """
-
     # The raw_data is not necessarily zero-mean when it is corrupted by RFI.
     # If so, estimated sample covariance matrix cov_cpi should be called 
     # sample correlation  matrix instead.  The reference below demonstrates
@@ -274,30 +244,37 @@ def compute_evd_gap(
     # Application in Narrow-Band Interference Suppression for SAR”, IEEE Geoscience 
     # and Remote Sensing Letters, vol. 4, no. 1, pp. 76,2007.
 
-    if mask_valid_cpi is None:
-        mask_valid_cpi = np.ones(raw_data.shape, dtype=bool)
+    if prf_dither_mode:
+        if mask_valid_cpi is None:
+            mask_valid_cpi = np.ones(raw_data.shape, dtype=bool)
+        else:
+            mask_valid_cpi = mask_valid_cpi.astype(bool, copy=False)
 
-    if mask_valid_cpi.shape != raw_data.shape:
-        raise ValueError(f"mask shape {mask_valid_cpi.shape} != data shape {raw_data.shape}")
-    
-    cov_cpi = compute_gap_exclusion_cov(
-        raw_data, 
-        mask_valid_cpi=mask_valid_cpi, 
-        off_diag_overlap_ratio=off_diag_overlap_ratio,
-        diag_valid_ratio=diag_valid_ratio,
-    )
+        if mask_valid_cpi.shape != raw_data.shape:
+            raise ValueError(
+                f"CPI mask shape {mask_valid_cpi.shape} != CPI data shape {raw_data.shape}"
+            )
+
+        cov_cpi = compute_gap_exclusion_cov(
+            raw_data,
+            mask_valid_cpi=mask_valid_cpi,
+            off_diag_overlap_ratio=off_diag_overlap_ratio,
+            diag_valid_ratio=diag_valid_ratio,
+        )
+    else:
+        num_rng_samples = raw_data.shape[1]
+        cov_cpi = (raw_data @ raw_data.conj().T) / num_rng_samples
 
     eig_val_sort, eig_vec_sort = eigen_decomp_sort(cov_cpi)
 
     return eig_val_sort, eig_vec_sort
 
-
 def compute_gap_exclusion_cov(
     data: np.ndarray,
     *,
     mask_valid_cpi: np.ndarray = None,
-    off_diag_overlap_ratio: float = 0.2,
-    diag_valid_ratio: float = 0.15,
+    off_diag_overlap_ratio: float = 0.25,
+    diag_valid_ratio: float = 0.20,
 ):
     """
     Compute a gap-excluded slow-time sample covariance matrix.
@@ -341,6 +318,21 @@ def compute_gap_exclusion_cov(
     # Sample Covariance Matrix
     min_valid_off_diag = max(1, int(np.ceil(off_diag_overlap_ratio * num_rng_samples)))
     min_valid_diag = max(1, int(np.ceil(diag_valid_ratio * num_rng_samples))) 
+
+    rng_samples_min = 2 * num_pulses
+
+    if min_valid_off_diag < rng_samples_min:
+        warnings.warn(f"""
+            Minimum number of samples required per pulse to estimate sample covariance matrix
+            is {rng_samples_min}. The number of valid overlapping off-diagonal samples is
+            {min_valid_off_diag}.
+        """)
+
+    if min_valid_diag < rng_samples_min:
+        warnings.warn(f"""
+            Minimum number of samples required per pulse to estimate sample covariance matrix
+            is {rng_samples_min}. The number of valid diagonal samples is {min_valid_off_diag}.
+        """)
 
     # Zero-out invalid samples
     x_valid =  data * mask_valid_cpi
--- a/python/packages/isce3/signal/compute_evd_cpi.py
+++ b/python/packages/isce3/signal/compute_evd_cpi.py
@@ -186,11 +186,11 @@ def compute_evd_tb(
             diag_valid_ratio=diag_valid_ratio,
         )
 
-        # Verify if the eigenvalue of CPI at index defind by  noise_ev_idx is meaningful
+        # Verify if the eigenvalue of CPI at index defind by min_ev_valid_idx is meaningful
         eig_val_sort_abs = np.maximum(np.abs(eig_val_sort), 1e-30)
         noise_ev_norm_db = 10 * np.log10(eig_val_sort_abs[min_ev_valid_idx] / eig_val_sort_abs[0])
 
-        if noise_ev_norm_db < -rx_dynamic_range_db:
+        if -noise_ev_norm_db > rx_dynamic_range_db:
             tb_is_valid = False
             break
 
@@ -251,7 +251,7 @@ def compute_evd(
 
         if mask_valid_cpi.shape != raw_data.shape:
             raise ValueError(
-                f"CPI mask shape {mask_valid_cpi.shape} != CPI data shape {raw_data.shape}"
+                f"Valid CPI mask shape {mask_valid_cpi.shape} != CPI data shape {raw_data.shape}"
             )
 
         cov_cpi = compute_gap_exclusion_cov(
@@ -297,6 +297,11 @@ def compute_gap_exclusion_cov(
     -------
     cov : (num_pulses, num_pulses) complex64
         Gap-excluded sample covariance matrix.
+
+    Notes
+    -------
+    The number of range samples per purlse must be equal or larger than 2 x number of pulses 
+    to be processed in order to have a reliable sample covariance matrix estimate.
     """
 
     num_pulses, num_rng_samples = data.shape
@@ -305,7 +310,7 @@ def compute_gap_exclusion_cov(
         mask_valid_cpi = np.ones(data.shape, dtype=bool)
 
     if mask_valid_cpi.shape != data.shape:
-        raise ValueError(f"mask shape {mask_valid_cpi.shape} != data shape {data.shape}")
+        raise ValueError(f"CPI mask shape {mask_valid_cpi.shape} != CPI data shape {data.shape}")
 
     if not (0.0 <= off_diag_overlap_ratio <= 1.0):
         raise ValueError("off_diag_overlap_ratio must be between 0 and 1.")
@@ -318,6 +323,11 @@ def compute_gap_exclusion_cov(
     min_valid_off_diag = max(1, int(np.ceil(off_diag_overlap_ratio * num_rng_samples)))
     min_valid_diag = max(1, int(np.ceil(diag_valid_ratio * num_rng_samples))) 
 
+    # The number of range samples per purlse must be equal or larger than 2 x number of pulses
+    # to be processed in order to have a reliable sample covariance matrix estimate.
+    # However, sometimes we would like to explore the trade-off betweeen less number of samples
+    # to estimate sample covariance matrix and reduction of RFI artifact. Hence only a warning
+    # is raised.
     rng_samples_min = 2 * num_pulses
 
     if min_valid_off_diag < rng_samples_min:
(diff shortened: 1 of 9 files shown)


python/packages/isce3/signal/compute_evd_cpi.py
```diff
@@ -168,3 +170,247 @@ def compute_evd_tb(
...
+    for i in range(num_pulses):
+        # Compute Diagonal terms
+        valid_pulse = mask_valid_cpi[i]
+        num_valid_samples_pulse = int(valid_pulse.sum())
+
+        if num_valid_samples_pulse >= min_valid_diag:
+            pulse_i = data[i, valid_pulse]
+            cov[i, i] = np.vdot(pulse_i, pulse_i) / num_valid_samples_pulse
+        else:
+            cov[i, i] = 0.0
```
rad-eng-59:
```suggestion
```

python/packages/isce3/signal/compute_evd_cpi.py
```diff
@@ -168,3 +170,247 @@ def compute_evd_tb(
...
+        Minimum fraction of valid samples required to compute a diagonal term in the
+        sample covariance matrix entry R_ii.
+        
+    Returns
+    -------
+    cov : (num_pulses, num_pulses) complex64
+        Gap-excluded sample covariance matrix.
+    """
+
+    num_pulses, num_rng_samples = data.shape
```
rad-eng-59, line 312: Add a check to see if `x` percent of total number of samples is above min required CPI length per theory and if not issue a warning.

FYI, for fixed PRF, make sure to check the number of valid samples (fixed over all CPI pulses) meets the min requirement and they are not all zeros.

python/packages/isce3/signal/compute_evd_cpi.py
```diff
@@ -168,3 +170,247 @@ def compute_evd_tb(
...
+    eig_vec_sort_array: 3D array of complex with dimension [num_cpi x cpi_len x cpi_len]
+        Sorted column vector Eigenvectors of all CPIs based on index of sorted Eigenvalues
+    tb_is_valid : bool
+        False if any CPI in the threshold block does not have enough usable
+        eigenvalues for noise_ev_idx
+    """
+
+    # compute number of CPIs
+    num_pulses, num_rng_samples = raw_data.shape
+    num_cpi = num_pulses // cpi_len
```
rad-eng-59, line 137:
```suggestion
    if cpi_len > num_pulses:
        raise ValueError(f'"cpi_len" {cpi_len} is larger than number of pulses {num_pulses}!')
    num_cpi = num_pulses // cpi_len
```

python/packages/isce3/signal/compute_evd_cpi.py
```diff
@@ -168,3 +170,247 @@ def compute_evd_tb(
...
+        )
+
+        # Count number of usable eigenvalues of CPI and compare with chosen noise_ev_idx.
+        # Since Python uses 0-based indexing, a CPI must have at least noise_ev_idx + 1 usable
+        # eigenvalues.
+        # Abnormal small Eigenvalues in general occur only at the tail of the sorted
+        # Eigenvalues.  Zero Eigenvalues are replaced by a small epsilon, 1e-30 to 
+        # ensure log values of such are greater than zero.
+        eig_val_sort_db = 10*np.log10(np.maximum(np.abs(eig_val_sort), 1e-30))
+        usable_rank = int(np.count_nonzero(eig_val_sort_db > 0))
```
rad-eng-59: Adding some user-defined rank computation per max expected dynamic range of eigen values.
```suggestion
        # do we need to peak normalize this?
        eig_val_norm = np.copy(eig_val_sort)
        if eig_val_sort[0] > eig_val_sort[-1]:
            eig_val_norm /=  eig_val_sort[0]
        # set to max possible dynamic range (16 bits  ~ 96 dB ~ 1e-9) or better to ENOB + floating-point precision ( < 8 bit ~ 50 dB ~ 1e-5)
        usable_rank = np.sum(~np.isclose(eig_val_norm, 0, atol=1e-9))        
```

python/packages/isce3/signal/compute_evd_cpi.py
```diff
@@ -168,3 +170,247 @@ def compute_evd_tb(
...
+
+        if usable_rank < noise_ev_idx + 1:
```
rad-eng-59: Replace  `noise_ev_idx` with `min_cpi_rank` and then before for loop add error check `min_cpi_rank <= cpi_len` and issue a warning if not ` 1 < min_cpi_rank < cpi_len - 1` (recommended range in docstring).
```suggestion
        if usable_rank < min_cpi_rank:
```

python/packages/isce3/signal/compute_evd_cpi.py
```diff
@@ -168,3 +170,247 @@ def compute_evd_tb(
...
+
+    return eig_val_sort_array, eig_vec_sort_array, tb_is_valid
+
+
+def compute_evd_gap(
+    raw_data: np.ndarray,
+    *,
+    mask_valid_cpi: np.ndarray = None,
+    off_diag_overlap_ratio: float = 0.1,
+    diag_valid_ratio: float = 0.05,
```
rad-eng-59: At the runconfig, it is assumed that the `num_range_sample_block >= 250`, then here are the suggested default percentage. 
```suggestion
    off_diag_overlap_ratio: float = 0.25,
    diag_valid_ratio: float = 0.2, 
```

python/packages/isce3/signal/compute_evd_cpi.py
```diff
@@ -168,3 +170,247 @@ def compute_evd_tb(
...
+    rng_samples_min = 2 * cpi_len
+
+    # Verify number of range samples in raw data is greater than minimum needed
+    # to estimate Sample Covariance Matrix
+    if num_rng_samples < rng_samples_min:
+        raise ValueError(
+            "Minimum number of samples in a range block to estimate Sample Covariance"
+            f" Matrix is {rng_samples_min}! Current number of samples per range block"
+            f" is {num_rng_samples}!"
+        )
```
Tyler-g-hudson, line 151: This block assumes that a user with no knowledge of RFI mitigation is going to understand the significance of these numbers and how they are calculated. In general, it's most helpful to the user to be able to identify at a glance:
1. What code is throwing this error (e.g. start the ValueError string with something like 'RFI detection:')
2. How the errored values were acquired. In this case, it looks like these numbers are coming from the shape of the data and the CPI length; Some quick instruction regarding setting an appropriate CPI length for this data would be useful here e.g. CPI length must be no more than half of the number of range pixels.

python/packages/isce3/signal/compute_evd_cpi.py
```diff
@@ -168,3 +170,247 @@ def compute_evd_tb(
...
+
+    # Verify Total number of pulses is greater than CPI length
+    if num_pulses < cpi_len:
+        raise ValueError(
+            f"Coherent Processing Interval length exceeds total number of pulses {num_pulses}!"
+        )
```
Tyler-g-hudson, line 157:
```suggestion
        raise ValueError(
            f"Coherent Processing Interval length (cpi_len) exceeds total number of pulses {num_pulses}!"
        )
```

python/packages/isce3/signal/compute_evd_cpi.py
```diff
@@ -168,3 +170,247 @@ def compute_evd_tb(
...
+
+    # F. Zhou, R. Wu, M. Xing, and Z. Bao, “Eigensubspace-Based Filtering With 
+    # Application in Narrow-Band Interference Suppression for SAR”, IEEE Geoscience 
+    # and Remote Sensing Letters, vol. 4, no. 1, pp. 76,2007.
+
+    if mask_valid_cpi is None:
+        mask_valid_cpi = np.ones(raw_data.shape, dtype=bool)
+
+    if mask_valid_cpi.shape != raw_data.shape:
+        raise ValueError(f"mask shape {mask_valid_cpi.shape} != data shape {raw_data.shape}")
```
Tyler-g-hudson:
```suggestion
        raise ValueError(f"Valid CPI mask shape {mask_valid_cpi.shape} != data shape {raw_data.shape}")
```

python/packages/isce3/signal/compute_evd_cpi.py
```diff
@@ -168,3 +170,247 @@ def compute_evd_tb(
...
+        Gap-excluded sample covariance matrix.
+    """
+
+    num_pulses, num_rng_samples = data.shape
+
+    if mask_valid_cpi is None:
+        mask_valid_cpi = np.ones(data.shape, dtype=bool)
+
+    if mask_valid_cpi.shape != data.shape:
+        raise ValueError(f"mask shape {mask_valid_cpi.shape} != data shape {data.shape}")
```
Tyler-g-hudson:
```suggestion
        raise ValueError(f"Valid CPI mask shape {mask_valid_cpi.shape} != data shape {data.shape}")
```

python/packages/isce3/signal/rfi_detection_evd.py
```diff
@@ -285,8 +337,11 @@ def rfi_detect_evd_tb(
     num_cpi, cpi_len = eig_val_sort_array.shape
 
     # Ensure detection threshold is a positive value
-    if detect_threshold <= 0:
-        raise ValueError("Detection threshold must be a positive value!")
+    if (not np.isfinite(detect_threshold)) or (detect_threshold <= 0):
+        #raise ValueError("Detection threshold must be a positive value!")
```
Tyler-g-hudson: It looks like this ValueError is no longer needed
```suggestion
```

python/packages/isce3/signal/rfi_process_evd.py
```diff
@@ -139,23 +167,43 @@ def run_slow_time_evd(
             "Max number of deg. of freedom must be less than number of pulses in a CPI."
         )
 
+    # Verify noise_ev_idx
+    if noise_ev_idx >= cpi_len:
+        raise ValueError(
+            f"noise_ev_idx must be less than {cpi_len - 1}, got {noise_ev_idx}."
+    )
```
Tyler-g-hudson: The logic here and the error statement say two different things; the error output says it needs to be less than `cpi_len - 1` but the logic only checks if it's less than `cpi_len`.
```suggestion
    if noise_ev_idx >= (cpi_len - 1):
        raise ValueError(
            f"noise_ev_idx must be less than cpi_len - 1: {cpi_len - 1}, got {noise_ev_idx}."
    )
```

python/packages/isce3/signal/rfi_process_evd.py
```diff
@@ -139,23 +167,43 @@ def run_slow_time_evd(
...
+
+    # Verify Mask shape
+    if mask_valid.shape != raw_data.shape:
+        raise ValueError(f"mask shape {mask_valid.shape} != data shape {raw_data.shape}")
```
Tyler-g-hudson:
```suggestion
        raise ValueError(f"Valid CPI mask shape {mask_valid.shape} != data shape {raw_data.shape}")
```

Tyler-g-hudson (commented):
I looked over your error strings and added some commentary on how to make them more descriptive. In one of them, I think there is also a small logic error which could lead to an edge case.

python/packages/isce3/signal/compute_evd_cpi.py
```diff
@@ -168,3 +170,247 @@ def compute_evd_tb(
...
+        # Compute Diagonal terms
+        valid_pulse = mask_valid_cpi[i]
+        num_valid_samples_pulse = int(valid_pulse.sum())
+
+        if num_valid_samples_pulse >= min_valid_diag:
+            pulse_i = data[i, valid_pulse]
+            cov[i, i] = np.vdot(pulse_i, pulse_i) / num_valid_samples_pulse
+        else:
+            cov[i, i] = 0.0
+
```
hb6688, line 373: Improve the runtime by vectoring the sample covariance matrix construction.  Use as special type with numpy or scipy function for optimization.

python/packages/isce3/signal/compute_evd_cpi.py
hb6688: @Tyler-g-hudson, The comments I made above this error statement describes the reference to this error message.  I also made it a bit more descriptive.

hb6688: @Tyler-g-hudson, I actually derived of this computation based on my own understanding of the referenced equation

python/packages/isce3/signal/compute_evd_cpi.py
hb6688: @Tyler-g-hudson , I combined some of the functions from compute_evd_cpi.py so these error messages are no longer there.

hb6688:
@rad-eng-59 @Tyler-g-hudson @bhawkins @hfattahi , please take a look at newly updated RFI ST-EVD algorithm. I made several changes:

1. Vectorized gap-exclusion sample covariance estimation
2. Updated minimum usable eigenvalue check with rx_dynamic_range = -50 dB as an input parameter
3. Addressed the comments from Hirad and Tyler
4. Combined fixed-PRF and constant-PRF computation into one single compute_evd function
5. Updated focus.py with reading of subswath mask and mode of operation.

I believe we should try to deliver this important update for the 04/16 release, and I hope you can facilitate this delivery.

Thank you

python/packages/isce3/signal/compute_evd_cpi.py
```diff
@@ -74,62 +74,59 @@ def eigen_decomp_sort(cov_matrix):
...
-        raw data to be processed
+        Raw data to be processed
     cpi_len: int, optional
-        Number of slow-time pulses within a CPI, default=32
+        Number of slow-time pulses within a CPI, default=16
+    prf_dither_mode: bool, optional, default = True
+        If True, use gap-aware covariance estimation
+    mask_valid : np.ndarray bool, [num_pulses x num_rng_samples], optional
+        Valid-sample mask with same shape as raw_data. Required if
+        prf_dither_mode=True.
```
bhawkins: If I were writing this, I'd probably just have the mask argument. Since the mask is optional, the flag can be derived from it as `prf_dither_mode = mask_valid is not None`. They're not really independent parameters, right?

Edit: After looking at the rest of the code, I see that mask is always populated. So this change would involve changing a lot of things up the call chain, which may not be worth the time it'd take.

python/packages/isce3/signal/compute_evd_cpi.py
```diff
@@ -74,62 +74,59 @@ def eigen_decomp_sort(cov_matrix):
...
+        Minimum fraction of valid samples required to compute a diagonal term
+        in the sample covariance matrix entry R_ii.
+    min_ev_valid_idx: int, optional
+        Eigenvalue index used by threshold estimation to estimate the slow-time minimum
+        Eigenvalue slope. This parameter is also used to validate that the threshold block
+        has enough usable eigenvalues for robust sample covaraince estimation of a CPI.
+    rx_dynamic_range_db: int, optional
+        radar platform receiver dynamic range, e.g. -50 dB. This is applied as a threshold
+        to determine if the Eigenvalue under test is meaningfully signficant. If the
+        Eigenvalue under test is less than this threshold, it will be viewed as unusable.
```
bhawkins: A negative dynamic range doesn't make sense. I think we should make it positive and update the logic accordingly. Also I don't see why it's documented as int rather than float.

python/packages/nisar/workflows/focus.py
```diff
@@ -1120,6 +1127,17 @@ def process_rfi(cfg: Struct, raw_data: np.ndarray,
     # Mitigate in place unless user wants a debug file to compare raw and
     # mitigated data.  This means you'd need to run the workflow twice to find
     # a bug specific to in-place vs out-of-place processing.
+
+    num_pulses = raw_data.shape[0]
+    mask_valid = np.zeros(raw_data.shape, dtype=bool)
+    pulse_idx = np.arange(num_pulses)
+
+    # Read Sub-Swath Mask for all pulses of the raw data block
+    for imask, ipulse in enumerate(pulse_idx):
```
bhawkins: This doesn't look right. As written, `imask` and `ipulse` are always the same. Should they be? In the context of the main processing loop, I see
```Python
            swaths = raw.getSubSwaths(channel_in.freq_id, tx=pol[0])
            swaths = swaths[:, pulse_begin:pulse_end, :]
```
So yes, it's already been sliced to match the raw data block, and you could just do `for i in range(num_pulses):` here. I suggest making that simplification so that I don't second-guess the code every time I read it.

I'm also not excited about allocating the whole mask in memory. We don't do that anywhere else since it's basically just as easy to construct it for each processing block. However, for a 50k x 50k image it's still < 3 GB, so maybe that's not worth worrying about.

share/nisar/defaults/focus.yaml
```diff
@@ -249,7 +249,19 @@ runconfig:
...
+                    num_cpi_per_threshold_block: 15
+                    # minimum fraction of samples required for cross-variance terms
+                    # of sample covariance matrix
+                    off_diag_overlap_ratio: 0.25
+                    # minimum fraction of samples required for covariance terms
+                    # of sample covariance matrix
+                    diag_valid_ratio: 0.20
+                    # This ratio will be used to determine the minimum number of valid Eigenvalues
+                    # required for a CPI. min_ev_valid_idx = min_valid_ev_ratio * cpi_len
+                    min_valid_ev_ratio: 0.65
```
bhawkins: It's a little confusing that these are all called `*_ratio` but are ratios of different things. Maybe you could call this one `normalized_minimum_rank` instead? Not a big deal.

bhawkins (commented):
Thanks. It looks like this patch correctly implements the changes we talked about. I just have a few comments, mostly about the new interfaces.

python/packages/isce3/signal/compute_evd_cpi.py
```diff
@@ -146,25 +143,241 @@ def compute_evd_tb(
...
+        mask_valid_cpi = mask_valid[cpi_slow_time]
+        eig_val_sort, eig_vec_sort = compute_evd(
+            data_cpi,
+            prf_dither_mode=prf_dither_mode,
+            mask_valid_cpi=mask_valid_cpi,
+            off_diag_overlap_ratio=off_diag_overlap_ratio,
+            diag_valid_ratio=diag_valid_ratio,
+        )
+
+        # Verify if the eigenvalue of CPI at index defind by  noise_ev_idx is meaningful
```
Tyler-g-hudson:
```suggestion
        # Verify if the eigenvalue of CPI at index defined by eig_val_sort is meaningful
```
Looks like you've renamed a variable here without modifying this comment

python/packages/isce3/signal/compute_evd_cpi.py
Tyler-g-hudson: Looks like you've edited other values per this set of suggestions but not this one

Tyler-g-hudson (commented):
Just covering some of the modifications made due to my earlier review, a couple of small documentation comments.

python/packages/isce3/signal/compute_evd_cpi.py
```diff
@@ -146,25 +143,241 @@ def compute_evd_tb(
...
+        an off-diagonal term in the sample covariance matrix entry R_ij.
+    diag_valid_ratio : float, optional, deffault = 0.20
+        Minimum fraction of valid samples required to compute a diagonal term in the
+        sample covariance matrix entry R_ii.
+        
+    Returns
+    -------
+    cov : (num_pulses, num_pulses) complex64
+        Gap-excluded sample covariance matrix.
+    """
```
rad-eng-59, line 310:
```suggestion

    Notes
    -------
    The number of range samples must be equal or larger than number of pulses for
    the input data to have a reliably calculated covariance matrix.
    """
```

python/packages/isce3/signal/compute_evd_cpi.py
```diff
@@ -146,25 +143,241 @@ def compute_evd_tb(
...
+    if mask_valid_cpi.shape != data.shape:
+        raise ValueError(f"mask shape {mask_valid_cpi.shape} != data shape {data.shape}")
+
+    if not (0.0 <= off_diag_overlap_ratio <= 1.0):
+        raise ValueError("off_diag_overlap_ratio must be between 0 and 1.")
+
+    if not (0.0 <= diag_valid_ratio <= 1.0):
+        raise ValueError("diag_valid_ratio must be between 0 and 1.")
+
+    # Minimum Samples required to compute diagonal and off-diagonal terms of
```
rad-eng-59, line 326: Check if the input data shape meet minimum requirements for reliable cov matrix calculation and then raise error (better) or warning like partially valid block.
```suggestion
    if num_rng_samples < 2 * num_pulses:
         warnings.warn("Not enough range samples for forming sample covariance
             matrix reliably!")    
    # Minimum Samples required to compute diagonal and off-diagonal terms of
```

python/packages/isce3/signal/compute_evd_cpi.py
```diff
@@ -74,62 +74,59 @@ def eigen_decomp_sort(cov_matrix):
...
-
 def compute_evd_tb(
     raw_data: np.ndarray,
-    cpi_len=32,
+    cpi_len: int=16,
+    prf_dither_mode: bool=True,
+    mask_valid: np.ndarray=None,
+    off_diag_overlap_ratio: float=0.25,
+    diag_valid_ratio: float=0.20,
+    min_ev_valid_idx: int=10,
```
rad-eng-59, line 85: I thought we talked about this both in the meeting and in my earlier comments. We agreed on either of two options:
1. Set the default to None (optional) ad set it based on `cpi_len` internally by default if not provided.
2. Change name and the definition slightly to use fraction or percentage of the `cpi_len`, a value within `[0, 1]`  instead with default set to `0.7`.

python/packages/isce3/signal/compute_evd_cpi.py
```diff
@@ -74,62 +74,59 @@ def eigen_decomp_sort(cov_matrix):
...
+        prf_dither_mode=True.
+    off_diag_overlap_ratio : float, optional, default = 0.25
+        Minimum overlap ratio used by gap exclusion covariance estimation
+    diag_valid_ratio : float, optional, default = 0.20
+        Minimum fraction of valid samples required to compute a diagonal term
+        in the sample covariance matrix entry R_ii.
+    min_ev_valid_idx: int, optional
+        Eigenvalue index used by threshold estimation to estimate the slow-time minimum
+        Eigenvalue slope. This parameter is also used to validate that the threshold block
+        has enough usable eigenvalues for robust sample covaraince estimation of a CPI.
```
rad-eng-59, line 116: Check out my earlier comment on how to default or redefine `min_ev_valid_idx` per `cpi_lenght`
How would this value being set given it is optional? 
Please note that when you say optional you need to explain how its value is being determined in the absence of the input.

python/packages/isce3/signal/compute_evd_cpi.py
```diff
@@ -146,25 +143,241 @@ def compute_evd_tb(
...
         )
 
     # Verify Total number of pulses is greater than CPI length
     if num_pulses < cpi_len:
         raise ValueError(
-            f"Coherent Processing Interval length exceeds total number of pulses {num_pulses}!"
+            f"Coherent Processing Interval length of {cpi_len} exceeds total number of pulses {num_pulses}!"
+        )
+
+    if min_ev_valid_idx >= cpi_len:
```
rad-eng-59, line 159: The docstring says, this is optional, then what value is being compared with `cpi_len`? Confusing docstring. See my earlier comments on `min_ev_valid_idx`

python/packages/isce3/signal/compute_evd_cpi.py
```diff
@@ -146,25 +143,241 @@ def compute_evd_tb(
...
+            mask_valid_cpi=mask_valid_cpi,
+            off_diag_overlap_ratio=off_diag_overlap_ratio,
+            diag_valid_ratio=diag_valid_ratio,
+        )
+
+        # Verify if the eigenvalue of CPI at index defind by  noise_ev_idx is meaningful
+        eig_val_sort_abs = np.maximum(np.abs(eig_val_sort), 1e-30)
+        noise_ev_norm_db = 10 * np.log10(eig_val_sort_abs[min_ev_valid_idx] / eig_val_sort_abs[0])
+
+        if noise_ev_norm_db < rx_dynamic_range_db:
```
rad-eng-59: Dynamic range must be a positive value and thus, this should be `>`!
```suggestion
        if noise_ev_norm_db > rx_dynamic_range_db:
```

python/packages/isce3/signal/compute_evd_cpi.py
```diff
@@ -74,62 +74,59 @@ def eigen_decomp_sort(cov_matrix):
...
+        Minimum overlap ratio used by gap exclusion covariance estimation
+    diag_valid_ratio : float, optional, default = 0.20
+        Minimum fraction of valid samples required to compute a diagonal term
+        in the sample covariance matrix entry R_ii.
+    min_ev_valid_idx: int, optional
+        Eigenvalue index used by threshold estimation to estimate the slow-time minimum
+        Eigenvalue slope. This parameter is also used to validate that the threshold block
+        has enough usable eigenvalues for robust sample covaraince estimation of a CPI.
+    rx_dynamic_range_db: int, optional
+        radar platform receiver dynamic range, e.g. -50 dB. This is applied as a threshold
```
rad-eng-59: What is the default value? It says opional but doesn't say how this will be set internally when not provided.
Also, dynamic range is positive value.

```suggestion
    rx_dynamic_range_db: float, default=50.0
        radar platform receiver dynamic range in dB, e.g. 50 dB. This is applied as a threshold
```

python/packages/isce3/signal/rfi_detection_evd.py
rad-eng-59: remove default values. The optioanl keyword args can not appear prior to positional args.

```suggestion
   num_max_trim: int
       Number of large-value outliers to be trimmed in slow-time minimum Eigenvalues.
   num_min_trim: int
       Number of small-value outliers to be trimmed in slow-time minimum Eigenvalues
```

python/packages/isce3/signal/rfi_detection_evd.py
```diff
@@ -72,6 +78,24 @@ def rfi_detect(
         time. Hence the standard (STD) deviation of multiple dominant EVs across slow time 
         defined by this parameter are compared. The one with the maximum STD is used for RFI
         Eigenvalue first difference computation.
+    off_diag_overlap_ratio : float, optional
+        Minimum overlap ratio used by gap exclusion covariance estimation
+    diag_valid_ratio : float, optional
+        Minimum fraction of valid samples required to compute a diagonal term in the
+        sample covariance matrix entry R_ii.
```
rad-eng-59: Are these really optional? What is the defaulted value? Please fix the doctring

python/packages/isce3/signal/rfi_detection_evd.py
```diff
@@ -72,6 +78,24 @@ def rfi_detect(
...
+        If True, L0B acquisition is of PRF Dithering mode. Sample Covariance Matrix
+        is computed differently by excluding the invalid data gaps.
+    min_ev_valid_idx: int
+        Eigenvalue index used by threshold estimation to estimate the slow-time minimum
+        Eigenvalue slope. This parameter is also used to validate that the threshold block
+        has enough usable eigenvalues for robust sample covaraince estimation of a CPI.
+    rx_dynamic_range_db: int, optional
+        radar platform receiver dynamic range, e.g. -50 dB. This is applied as a threshold
+        to determine if the Eigenvalue under test is meaningfully signficant. If the
+        Eigenvalue under test is less than this threshold, it will be viewed as unusable.
```
rad-eng-59: Please fix the doctring. See my earlier comments. The value must be positive and if it is optional with default value must appear after positional args with proper description.

python/packages/isce3/signal/rfi_detection_evd.py
```diff
@@ -112,14 +162,14 @@ def rfi_detect(
...
-    return rfi_cpi_flag_array, eig_vec_sort_array
-
+    return rfi_cpi_flag_array, eig_vec_sort_array, fig_merit_detect_tb
 
 def threshold_estimate_evd(
     eig_val_sort_array,
     num_max_trim=0,
     num_min_trim=0,
     max_num_rfi_ev=2,
+    min_ev_valid_idx=10,
```
rad-eng-59, line 171: As discussed several times, either default shall be determined by the size of `eig_val_sort_array` internally if not provided or rename this var to use fraction of the number of eigen values (size of the eigen vector) with a default set to a value within `[0, 1]`. Please check out earlier similar comments. Please also fix the docstring.

python/packages/isce3/signal/rfi_process_evd.py
```diff
@@ -67,8 +73,25 @@ def run_slow_time_evd(
...
     mitigate_enable: bool, default=False
         Enable mitigation
+    prf_dither_mode: bool
+        If True, L0B acquisition is of PRF Dithering mode. Sample Covariance Matrix
+        is computed differently by excluding the invalid data gaps.
+    min_valid_ev_ratio: float
+        This ratio will be used to determine the minimum number of valid Eigenvalues
+        required for a CPI. min_ev_valid_idx = min_valid_ev_ratio * cpi_len
+    rx_dynamic_range_db: int, optional
+        radar platform receiver dynamic range, e.g. -50 dB. This is applied as a threshold
```
rad-eng-59:
```suggestion
    rx_dynamic_range_db: float, default=50
        radar platform receiver dynamic range in dB, e.g. 50 dB. This is applied as a threshold
```

python/packages/isce3/signal/rfi_process_evd.py
```diff
@@ -67,8 +73,25 @@ def run_slow_time_evd(
...
+        Minimum fraction of valid samples required to compute a diagonal term in the
+        sample covariance matrix entry R_ii.
     mitigate_enable: bool, default=False
         Enable mitigation
+    prf_dither_mode: bool
+        If True, L0B acquisition is of PRF Dithering mode. Sample Covariance Matrix
+        is computed differently by excluding the invalid data gaps.
+    min_valid_ev_ratio: float
+        This ratio will be used to determine the minimum number of valid Eigenvalues
+        required for a CPI. min_ev_valid_idx = min_valid_ev_ratio * cpi_len
```
rad-eng-59:
```suggestion
        required for a CPI. min_ev_valid_idx = min_valid_ev_ratio * cpi_len.
        Must be a value within (0,1].
```

python/packages/isce3/signal/rfi_process_evd.py
```diff
@@ -124,6 +149,16 @@ def run_slow_time_evd(
                 " as the input data"
             )
 
+    # Verify min_valid_ev_ratio
+    if min_valid_ev_ratio >= 1:
```
rad-eng-59: I thought the correct range is `(0, 1]`, that is ful cpi length, full rank, can be considered as min number of valid eigens. Am I mistaken?
```suggestion
    if min_valid_ev_ratio >= 1 or min_valid_ev_ratio  <= 0:
```

python/packages/isce3/signal/compute_evd_cpi.py
hb6688: @bhawkins , you are correct that prf_dither_mode and mask_valid are NOT independent parameters.  Hence your logic prf_dither_mode = mask_valid is not None is correct.

In my opinion, having a parameter of prf_dither_mode=True is a little more easier to understand than ' if mask valid is not None, then ..., for example'.  I will add error checking statements to make sure if prf_dither_mode is true, then there must be a mask_valid.

hb6688: @bhawkins, After thinking about it again, it may be a good idea to get rid of the redundancy and only have mask_valid to indicate prf dither mode.  I will address the other comments and come back to this.

if mask_valid is None:
    constant PRF
else:
    dithered PRF

share/nisar/defaults/focus.yaml
hb6688: @bhawkins , renamed it to normalized_min_rank_ratio

hb6688: This fraction serves two purposes: 1. Estimate the eigenvalue index that will be used to estimate minimum eigenvalue. 2. check to see if the Eigenvalue at computed eigenvalue index is meaningful.

python/packages/isce3/signal/compute_evd_cpi.py
hb6688: @rad-eng-59 In the top-level rfi_process_evd.py, this index is computed as follows and passed into the lower-level functions:

min_ev_valid_idx = int(np.floor(normalized_min_rank_ratio * cpi_len))

In my lower-level script, I do a check to avoid min_ev_valid_idx being set to greater than cpi_len

if min_ev_valid_idx >= cpi_len:
        raise ValueError(
            f"min_ev_valid_idx ({min_ev_valid_idx}) must be less than cpi_len ({cpi_len}). "
            "Since Python uses 0-based indexing, the maximum valid index is cpi_len - 1."
        )

python/packages/isce3/signal/rfi_detection_evd.py
hb6688: @rad-eng-59 Hi Hirad, I computed the min_ev_valid_idx as a function of cpi_len once at the top-level.  I do not want to do it at every lower-function.  The input to the top-level function is a ratio named: normalized_min_rank_ratio.  I do not think users will try to use the lower-level functions alone.

python/packages/isce3/signal/rfi_process_evd.py
hb6688: @rad-eng-59 , it should be (0, 1].  1 should be allowed. I will update the error checking

hb6688:
@bhawkins @rad-eng-59 , @Tyler-g-hudson , @hfattahi , I removed prf_dither_mode input argument, and made my best effort to ensure that the following logic is true with good code eadability.  I had to add an internal variable apply_gap_exclusion to manipulate sample covariance computation.

Top level:
if mask_valid is not None:
    apply_gap_exclusion = True
else:
    apply_gap_exclusion = False
 
Error checking
If apply_gap_exclusion is true and mask_valid is None:
    raiseError()

CPI-level:
if apply_gap_exclusion is False:
     ignore mask_valid_cpi, perform standard matrix multiplication
else:
     perform gap exclusion sample covariance estimation
     
I think my logic is sound, and if there is no obvious errors, I would like to keep it this way.  Please let me know if you see any logical errors in the code.  Thank you.

python/packages/isce3/signal/compute_evd_cpi.py
```diff
@@ -74,62 +74,62 @@ def eigen_decomp_sort(cov_matrix):
...
+    off_diag_overlap_ratio : float, optional, default = 0.25
+        Minimum overlap ratio used by gap exclusion covariance estimation
+    diag_valid_ratio : float, optional, default = 0.20
+        Minimum fraction of valid samples required to compute a diagonal term
+        in the sample covariance matrix entry R_ii.
+    min_ev_valid_idx: int, optional
+        Eigenvalue index used by threshold estimation to estimate the slow-time minimum
+        Eigenvalue slope. This parameter is also used to validate that the threshold block
+        has enough usable eigenvalues for robust sample covaraince estimation of a CPI.
+    rx_dynamic_range_db: int, optional, default = 50 dB
```
bhawkins, line 113:
```suggestion
    rx_dynamic_range_db: float, optional
```
I don't see why this needs to be an integer (ditto in the type annotation above)

bhawkins (commented):
I noticed some issues with the most recent changes. Rather than a litany of comments, I just created a PR with the desired fixes (https://github.com/hb6688/isce3_public/pull/1). That PR also moves the mask generation down a level, so we never have the whole thing in memory. It's a net reduction in lines of code, which I think is a good sign.

python/packages/isce3/signal/compute_evd_cpi.py
hb6688: @bhawkins , First of all, my logic in the code was not right in what I was trying to accomplish.  Somehow I thought about if in the case of fixed PRF, may be I still would like to use the mask.  Hence I created that variable apply_gap_exclusion but my code was not correct to begin with and as you pointed it would always go down the path of dithered PRF.  Your updates simplifies the logic, and I agree with it.  I already merged it, and I am just going to summarize it here for other to see:

In your updates, followings are accomplished:

1. In focus.py, if meta data indicates 'uniform_prf' mode, then subswaths_mask is None. Otherwise, populate the subswaths mask.  In the lower level compute_evd function, if mask_valid is None, treat it as fixed PRF and do straight forward matrix multiplication.  Otherwise, apply gap exclusion matrix multiplication.

2. Pushed subswaths mask reading down to lower-level functions.  Subswaths mask is read one Threshold block at a time with no more memory usage concern.

Thank you Brian.

hb6688: I will fix the docstring

bhawkins (approved):

python/packages/nisar/workflows/focus.py
```diff
@@ -1089,6 +1090,12 @@ def process_rfi(cfg: Struct, raw_data: np.ndarray,
         RSLC runconfig data
     raw_data : np.ndarray[np.complex64]
         Raw data layer.  May be modified in-place if mitigation is enabled.
+    swaths : np.ndarray [int], optional
+        Valid subswath samples, dims = (ns, nt, 2) where ns is the number of
+        sub-swaths, nt is the number of pulses, and the trailing dimension is
+        the [start, stop) indices of the sub-swath.  It's recommended to supply
+        this for modes with dithered PRI, where it will be used to normalize
+        the sample covariance matrix.
```
rad-eng-59, line 1098: It is also being used to determine whether a range block hit by TX gap has enough number of valid samples per two thresholds in cov matrix, one for diagonal and one for off-diagonal, if I'm not mistaken. So, its existence can also be useful for fixed PRF to skip bad CPI blocks in mitigation process. I leave the decision to @hb6688.

share/nisar/schemas/focus.yaml
```diff
@@ -421,6 +421,18 @@ slow_time_evd_options:
...
   num_cpi_per_threshold_block: int(min=2, required=False)
+  # minimum fraction of samples required for cross-variance terms
+  # of sample covariance matrix
+  off_diag_overlap_ratio: num(min=0.0, required=False)
+  # minimum fraction of samples required for co-variance terms
+  # of sample covariance matrix
+  diag_valid_ratio: num(min=0.0, required=False)
+  #This ratio will be used to determine the minimum number of valid Eigenvalues
+  ##required for a CPI. min_ev_valid_idx = int(np.round(min_rank_frac * cpi_len)) - 1
+  min_rank_frac: num(min=0.0, max=1.0, required=False)
```
rad-eng-59: Not a big deal, but just to be nitpicking,  the min rank of a square matrix can not be zero given there are at least some non-zero samples. Besides, the `min_ev_valid_idx` can  not be negative.
```suggestion
  min_rank_frac: num(exclusiveMinimum=0.0, max=1.0, required=False)
```

python/packages/isce3/signal/rfi_process_evd.py
```diff
@@ -114,6 +138,8 @@ def run_slow_time_evd(
     num_pulses_proc = cpi_len * num_cpi
     num_pulses_tb = cpi_len * num_cpi_tb
 
+    num_tb = num_pulses_proc // num_pulses_tb
```
rad-eng-59: Unused variable. 
```suggestion
```

python/packages/isce3/signal/rfi_process_evd.py
```diff
@@ -139,23 +171,49 @@ def run_slow_time_evd(
...
+    # Verify mask_valid: check to see if it is None or populated.
+    if (swaths is not None) and (swaths.shape[1] != raw_data.shape[0]):
+        raise ValueError("Require same number of rows in swaths and raw_data")
+
+    # Determine a valid Eigenvalue index to estimate minimum-Eigenvalue statistics,
+    # ensuring robustness against zero Eigenvalues caused by insufficient valid samples in a CPI.
+    min_ev_valid_idx = max(1, int(np.round(min_rank_frac * cpi_len)) - 1)
+
     # Run RFI Detection and Mitigation
     for idx_tb, tb_slow_time in enumerate(slice_gen(num_pulses_proc, num_pulses_tb)):
```
rad-eng-59: Unused var `idx_tb`
```suggestion
    for tb_slow_time in slice_gen(num_pulses_proc, num_pulses_tb):
```

python/packages/isce3/signal/rfi_process_evd.py
```diff
@@ -139,23 +171,49 @@ def run_slow_time_evd(
...
+        # Get valid data mask for all rows in current block.
+        if swaths is not None:
+            swaths_tb = swaths[:, tb_slow_time, :]
+            mask_valid = np.zeros((swaths_tb.shape[1], raw_data.shape[1]), bool)
+            for i in range(mask_valid.shape[0]):
+                for start, end in swaths_tb[:, i, :]:
+                    mask_valid[i, start:end] = True
+
         for idx_rng, tb_fast_time in enumerate(
```
rad-eng-59: Unused var `idx_rng`
```suggestion
        for tb_fast_time in 
```

python/packages/isce3/signal/rfi_process_evd.py
```diff
@@ -139,23 +171,49 @@ def run_slow_time_evd(
...
-            slice_gen(num_rng_samples, num_samples_rng_blk)
+            slice_gen(num_rng_samples, num_samples_rng_blk, combine_rem=True)
         ):
```
rad-eng-59, line 201:
```suggestion
        :
```

python/packages/isce3/signal/rfi_process_evd.py
```diff
@@ -139,23 +171,49 @@ def run_slow_time_evd(
             "Max number of deg. of freedom must be less than number of pulses in a CPI."
         )
 
+    # Verify mask_valid: check to see if it is None or populated.
+    if (swaths is not None) and (swaths.shape[1] != raw_data.shape[0]):
+        raise ValueError("Require same number of rows in swaths and raw_data")
+
+    # Determine a valid Eigenvalue index to estimate minimum-Eigenvalue statistics,
+    # ensuring robustness against zero Eigenvalues caused by insufficient valid samples in a CPI.
+    min_ev_valid_idx = max(1, int(np.round(min_rank_frac * cpi_len)) - 1)
```
rad-eng-59, line 179: Either issue error or warning. If warning, then perhaps limit/modify `max_deg_freedom` not to be equal or larger than `min_ev_valid_idx `. Note that these two input parameters `max_deg_freedom` and `min_rank_ratio` are NOT independent.  
```suggestion
    min_ev_valid_idx = max(1, int(np.round(min_rank_frac * cpi_len)) - 1)
    if max_deg_freedom >= min_ev_valid_idx:
        raise ValueError('Max RFI emitters shall be equla or less than min rank of the cov matrix! ') 
```

python/packages/isce3/signal/rfi_process_evd.py
```diff
@@ -139,23 +171,49 @@ def run_slow_time_evd(
...
+
     # Run RFI Detection and Mitigation
     for idx_tb, tb_slow_time in enumerate(slice_gen(num_pulses_proc, num_pulses_tb)):
+        # Get valid data mask for all rows in current block.
+        if swaths is not None:
+            swaths_tb = swaths[:, tb_slow_time, :]
+            mask_valid = np.zeros((swaths_tb.shape[1], raw_data.shape[1]), bool)
```
rad-eng-59:
```suggestion
            mask_valid = np.zeros((swaths_tb.shape[1], raw_data.shape[1]), dtype=bool)
```

python/packages/isce3/signal/rfi_detection_evd.py
```diff
@@ -62,17 +68,34 @@ def rfi_detect(
...
         defined by this parameter are compared. The one with the maximum STD is used for RFI
         Eigenvalue first difference computation.
-    threshold_params: ThresholdParams dataclass object
+    off_diag_overlap_ratio : float, default=0.25
+        Minimum overlap ratio used by gap exclusion covariance estimation
+    diag_valid_ratio : float, default=0.20
+        Minimum fraction of valid samples required to compute a diagonal term in the
+        sample covariance matrix entry R_ii.
+    rx_dynamic_range_db: float, default=50.0
+        Radar platform receiver dynamic range. This is applied as a threshold
```
rad-eng-59, line 91:
```suggestion
        Radar platform receiver dynamic range in dB. This is applied as a threshold
```

python/packages/isce3/signal/rfi_detection_evd.py
```diff
@@ -62,17 +68,34 @@ def rfi_detect(
...
+        to determine if the Eigenvalue under test is meaningfully signficant. If the
+        Eigenvalue under test is less than this threshold, it will be viewed as unusable.
+    mask_valid : np.ndarray bool, [num_pulses x num_rng_samples], default=None
```
rad-eng-59:
```suggestion
    mask_valid : np.ndarray bool or None, default=None.
         The shape of 2-D array must be [num_pulses x num_rng_samples] if provided. 
```

python/packages/isce3/signal/rfi_detection_evd.py
```diff
@@ -62,17 +68,34 @@ def rfi_detect(
...
+        Eigenvalue index used by threshold estimation to estimate the slow-time minimum
+        Eigenvalue slope. This parameter is also used to validate that the threshold block
+        has enough usable eigenvalues for robust sample covaraince estimation of a CPI.
+    num_max_trim: int, default=0
         Number of large-value outliers to be trimmed in slow-time minimum Eigenvalues.
-    num_min_trim: int, default = 0
+    num_min_trim: int, default=0
         Number of small-value outliers to be trimmed in slow-time minimum Eigenvalues
-    max_num_rfi_ev: int
+    max_num_rfi_ev: int, , default=2
```
rad-eng-59:
```suggestion
    max_num_rfi_ev: int, default=2
```

python/packages/isce3/signal/compute_evd_cpi.py
```diff
@@ -146,25 +142,241 @@ def compute_evd_tb(
...
+    num_pulses, num_rng_samples = data.shape
+
+    if mask_valid_cpi is None:
+        mask_valid_cpi = np.ones(data.shape, dtype=bool)
+
+    if mask_valid_cpi.shape != data.shape:
+        raise ValueError(f"CPI mask shape {mask_valid_cpi.shape} != CPI data shape {data.shape}")
+
+    if not (0.0 <= off_diag_overlap_ratio <= 1.0):
+        raise ValueError("off_diag_overlap_ratio must be between 0 and 1.")
```
rad-eng-59: Should not be within `(0, 1]` ? 
```suggestion
    if not (0.0 < off_diag_overlap_ratio <= 1.0):
        raise ValueError("off_diag_overlap_ratio must be between 0 and 1.")
```

python/packages/isce3/signal/compute_evd_cpi.py
```diff
@@ -146,25 +142,241 @@ def compute_evd_tb(
...
+
+    if not (0.0 <= diag_valid_ratio <= 1.0):
```
rad-eng-59: Should not be within `(0, 1]` ?
```suggestion
    if not (0.0 < diag_valid_ratio <= 1.0):
```

rad-eng-59 (approved):
I have some minor concerns that being addressed in the comments. Please check them out. But overall it looks good. Thank you both.

python/packages/isce3/signal/compute_evd_cpi.py
hb6688: @rad-eng-59 , I will correct docstring based on your comments, and put some thoughts into your mask question.

Thank you for your comments and review.

python/packages/isce3/signal/rfi_process_evd.py
hb6688: @rad-eng-59 , you are absolutely right about the unused variables. I use them for debugging purpose and there will be some more debugging.  I am going to do the following so I can do debugging in the future if needed.  Hopefully this is OK with you.

for _, tb_slow_time in enumerate(...):
    ...
    for _, tb_fast_time in enumerate(...):

python/packages/isce3/signal/rfi_process_evd.py
hb6688: @rad-eng-59 ,  min_ev_valid_idx is the index used to estimate minimum Eigenvalue.  In theory, max_deg_freedom can be greater than min_ev_valid_idx.  I will issue a warning.

hfattahi:
@hb6688 please rebase with develop. The Mac OS unit tests is already solved. If you rebase, the MacOS build and test also will pass.

python/packages/isce3/signal/rfi_process_evd.py
rad-eng-59: You can keep those unused variables as they are if you are keeping `enumerate`. Either way is fine. I just wanted to make sure you are aware of them. Thanks.

PR #240, opened 2026-03-19
Update SM SAS Version to v0.4.1 for R05.01.3 

xhuang-jpl:
This PR is to update the SM SAS version to v0.4.1 for the R05.01.3 where we bring back the historical SM products to replace the GCOV, which will significantly reduce the runtime, in addition, the common mask based on the surface flag is applied to the sigma0; and the updates for each algorithm.


hfattahi (approved):

Tyler-g-hudson (approved):
LGTM!

hfattahi:
Looks like your branch is behind develop. The Mac unit test is resolved. If you rebase with develop, that failure should be resolved.

xhuang-jpl:
I have rebased with the develop branch, and not sure why the MacOS unit test still fail. This will not block this PR's merge.

PR #241, opened 2026-03-19
Update the base Docker image to the 2026-03-01 patch release and upgrade pysolid to obtain bug fix

jshimada47:
Also update pysolid to v0.3.4 in the runtime and dev Conda spec files.

### DIFF
--- a/tools/imagesets/oracle8conda/dev/spec-file.txt
+++ b/tools/imagesets/oracle8conda/dev/spec-file.txt
@@ -238,7 +238,7 @@ https://conda.anaconda.org/conda-forge/linux-64/pyproj-3.7.1-py312h03c6e1f_1.con
 https://conda.anaconda.org/conda-forge/linux-64/pyre-1.12.5-py312h7f27a60_5.conda
 https://conda.anaconda.org/conda-forge/linux-64/pyrsistent-0.20.0-py312h66e93f0_1.conda
 https://conda.anaconda.org/conda-forge/linux-64/pyside6-6.9.1-py312hdb827e4_0.conda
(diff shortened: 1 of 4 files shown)


jshimada47:
@Tyler-g-hudson I used your script create_spec_file.py to find the correct URL for pysolid v0.3.4 on conda-forge after updating it to reference the newer base Docker image.  I ran the script for the 'runtime' and 'dev' spec-files.  Many of the other packages in the spec-files had newer versions available; however, I changed only the pysolid entry in the spec-files.  I also left the Miniforge version of 24.11.3-2 unchanged in create_spec_file.py.  (The current version is 26.1.0-0.)  I've been able to build a new Docker image with this configuration and will run some quick tests on isce3_scale this afternoon.

Tyler-g-hudson:
Thanks Joanne, please let me know the result - I'm a little concerned about only updating the one package since its dependencies may also have changed, which could result in some unexpected behavior.

jshimada47:
@Tyler-g-hudson @hfattahi So far one test case has completed RSLC/GSLC/GCOV successfully and one test case failed in RSLC about 30 seconds after execution started.  I reported the problem on [Brian's PR](https://github.com/isce-framework/isce3/pull/221).  I have two other test cases that have been running RSLC for > 15 minutes now.  I need to leave the office now but will start an InSAR test case later tonight.

jshimada47:
All the test cases have completed successfully, including the 40+5 case that failed at RSLC previously (and the 77-MHz case that I hadn't run earlier).  The ~160-MB InSAR log files are now 3-MB files.

Tyler-g-hudson (approved):
Since you've tested and everything works, LGTM!

PR #243, opened 2026-03-23
Fix solid earth tides computation fail when frequencyA is missing

xhuang-jpl:
This PR fixes a bug where solid earth tides computation fails when `frequencyA` is missing. The logic now falls back to `frequencyB` if available, and raises an error if neither exists.


python/packages/nisar/workflows/solid_earth_tides.py
```diff
@@ -266,9 +266,17 @@ def _extract_ref_epoch(ds_name: str):
         projection_dataset = h5_obj[f'{rdr_grid_path}/projection']
         epsg = projection_dataset.attrs['epsg_code']
 
-         # Wavelength in meters
-        wavelength = isce3.core.speed_of_light / \
-                h5_obj[f'{gunw_obj.GridsPath}/frequencyA/centerFrequency'][()]
+        # Wavelength in meters; fall back to frequencyB if frequencyA is not found
+        for freq in ('A', 'B'):
```
Tyler-g-hudson, line 270: Is there a reason why we capture first available of A or B instead of calculating a different datacube for each? They would have different values, wouldn't they?

Tyler-g-hudson (approved):
LGTM, just one question for clarification

python/packages/nisar/workflows/solid_earth_tides.py
xhuang-jpl: good question @Tyler-g-hudson, for NISAR InSAR processing, we are always processing frequencyA even if the product has both freqA and freqB. This fix is for users who need to process the freqB stand alone.

hfattahi (approved):
LGTM!

python/packages/nisar/workflows/solid_earth_tides.py
Tyler-g-hudson: Sure, I've understood that part - what I mean to ask is why we have selected to do that when the values might be different between two different data cubes, is it because we only expect people to use it for the primary frequency?

xhuang-jpl: We always expect people use the frequencyA as the primary at least for the NISAR.

PR #246, opened 2026-03-30
Add qFSP boundary mask to RSLC

bhawkins:
This PR implements the `inputDataExceptionMask` in RSLC products for the LSAR qFSP-related exception codes. It handles all six possible qFSP exception codes. For now, the strategy to is locate the peaks of the AC table weights in the EL angle domain and use the EL-angle LUT to generate the mask. This mask is in the raw data domain, so it's resampled to the RSLC grid using reskew LUTs. (These could be used to reskew other tables in some future PR, too). I added a vectorized `LUT2d::eval` overload to make this faster.

Eyeballing the plots for the case I tested on, I'm not sure the mask covers everything where there's obvious residual phase. So maybe instead of defining the overlap region as peak-to-peak, we should be more conservative and use a -6 dB threshold or something. I hope folks can do some InSAR tests to help figure out the best approach. 

Note that the mask is populated based on the contents of the input L0B `hasInputDataException` flag, so that must be populated to see any effect.

Another question is what to do about mixed-mode frames where one L0B has a qFSP slip and its neighbor doesn't. For now, the qFSP overlap region is marked invalid for the entirety of the output product.

### DIFF
--- a/python/extensions/pybind_isce3/core/LUT2d.cpp
+++ b/python/extensions/pybind_isce3/core/LUT2d.cpp
@@ -22,6 +22,7 @@ template<typename T>
 void addbinding(py::class_<LUT2d<T>> &pyLUT2d)
 {
 
+    using const_vec_t = const Eigen::Ref<const Eigen::VectorXd>;
     pyLUT2d
         .def(py::init<>())
         .def(py::init<const T&>())
@@ -138,7 +139,8 @@ void addbinding(py::class_<LUT2d<T>> &pyLUT2d)
             return self.data().map();
         })
         .def("eval", py::overload_cast<const double, const double>(&LUT2d<T>::eval, py::const_))
-        .def("eval", py::overload_cast<double,const Eigen::Ref<const Eigen::VectorXd>&>(&LUT2d<T>::eval, py::const_))
+        .def("eval", py::overload_cast<double,const const_vec_t&>(&LUT2d<T>::eval, py::const_))
+        .def("eval", py::overload_cast<const_vec_t&, const_vec_t&>(&LUT2d<T>::eval, py::const_))
         ;
 }
 
--- a/python/extensions/pybind_isce3/core/LUT2d.cpp
+++ b/python/extensions/pybind_isce3/core/LUT2d.cpp
@@ -139,7 +139,7 @@ void addbinding(py::class_<LUT2d<T>> &pyLUT2d)
             return self.data().map();
         })
         .def("eval", py::overload_cast<const double, const double>(&LUT2d<T>::eval, py::const_))
-        .def("eval", py::overload_cast<double,const const_vec_t&>(&LUT2d<T>::eval, py::const_))
+        .def("eval", py::overload_cast<double, const_vec_t&>(&LUT2d<T>::eval, py::const_))
         .def("eval", py::overload_cast<const_vec_t&, const_vec_t&>(&LUT2d<T>::eval, py::const_))
         ;
 }
--- a/python/packages/isce3/geometry/rdr2rdr.py
+++ b/python/packages/isce3/geometry/rdr2rdr.py
@@ -120,10 +120,10 @@ def make_reskew_lut(t_axis: Linspace,
         Dictionary specifying convergence parameters for geo2rdr solver.
         Keys among {"tol_height", "time_start", "time_start"}
         See isce3.geometry.geo2rdr_bracket
-    method: str
-        LUT2d interpolation method
-    b_error: bool
-        Enable LUT2d bounds error checking
+    method: str, optional
+        LUT2d interpolation method. Defaults to "bilinear".
+    b_error: bool, optional
+        Enable LUT2d bounds error checking. Defaults to True.
 
     Returns
     -------
--- a/python/packages/isce3/geometry/rdr2rdr.py
+++ b/python/packages/isce3/geometry/rdr2rdr.py
@@ -1,4 +1,5 @@
-from isce3.core import Ellipsoid, LookSide, LUT2d, Orbit
+import numpy as np
+from isce3.core import Ellipsoid, LookSide, LUT2d, Orbit, Linspace
 from isce3.geometry import DEMInterpolator, rdr2geo_bracket, geo2rdr_bracket
 from typing import Union, Optional
 
@@ -70,3 +71,79 @@ def rdr2rdr(t: float,
     tout, rout = geo2rdr_bracket(xyz, orbit_out, doppler_out, wavelength, side,
                                  **geo2rdr_params)
     return tout, rout
+
+
+def make_reskew_lut(t_axis: Linspace,
+                    r_axis: Linspace,
+                    orbit: Orbit,
+                    side: Union[str, LookSide],
+                    doppler_in: LUT2d,
+                    wavelength: float,
+                    dem: DEMInterpolator = DEMInterpolator(),
+                    doppler_out: Optional[LUT2d] = None,
+                    orbit_out: Optional[Orbit] = None,
+                    rdr2geo_params=dict(),
+                    geo2rdr_params=dict(),
+                    method: str = "bilinear",
+                    b_error: bool = True):
+    """
+    Generate a 2D look-up table for converting coordinates from one radar
+    geometry to another with a different Doppler (reskew) or orbit (motion
+    compensation).
+
+    Parameters
+    ----------
+    t_axis : isce3.core.Linspace
+	    Input azimuth time axis for the table, seconds past orbit epoch
+    r_axis : isce3.core.Linspace
+	    Input slant range axis for the table, m
+    orbit : isce3.core.Orbit
+	    Orbit defining radar motion on input path
+    side : {"left", "right", isce3.core.LookSide.Left, isce3.core.LookSide.Right}
+	    Flag describing which side the radar is looking.
+    doppler_in : isce3.core.LUT2d
+        Doppler look up table associated with `t_axis` and `r_axis`, Hz
+    wavelength : float
+        Wavelength associated with `doppler_in`, m
+    dem : isce3.geometry.DEMInterpolator, optional
+        Digital elevation model, m above ellipsoid.  Defaults to h=0.
+    doppler_out : isce3.core.LUT2d, optional
+        Doppler look up table vs output range and azimuth time, Hz
+        Defaults to `doppler_in`.
+    orbit_out : isce3.core.Orbit, optional
+        Orbit defining radar motion on output path.  Defaults to input `orbit`.
+    rdr2geo_params : dict, optional
+        Dictionary specifying convergence parameters for rdr2geo solver.
+        Keys among {"tol_aztime", "look_min", "look_max"}
+        See isce3.geometry.rdr2geo_bracket
+    geo2rdr_params: dict, optional
+        Dictionary specifying convergence parameters for geo2rdr solver.
+        Keys among {"tol_height", "time_start", "time_start"}
+        See isce3.geometry.geo2rdr_bracket
+    method: str
+        LUT2d interpolation method
+    b_error: bool
+        Enable LUT2d bounds error checking
+
+    Returns
+    -------
+    t_lut : isce3.core.LUT2d
+        Azimuth time in output geometry as a 2D function of (time, range) in the
+        input geometry, seconds past orbit epoch
+    r_lut : isce3.core.LUT2d
+        Range in output geometry as a 2D function of (time, range) in the input
+        geometry, m
+    """
+    m = len(t_axis)
+    n = len(r_axis)
+    t_out = np.zeros((m, n))
+    r_out = np.zeros((m, n))
+    for i, t_in in enumerate(t_axis):
+        for j, r_in in enumerate(r_axis):
+            t_out[i,j], r_out[i,j] = rdr2rdr(t_in, r_in, orbit, side,
+                doppler_in, wavelength, dem=dem, doppler_out=doppler_out,
+                orbit_out=orbit_out, rdr2geo_params=rdr2geo_params,
+                geo2rdr_params=geo2rdr_params)
+    t_lut = LUT2d(r_axis, t_axis, t_out, method=method, b_error=b_error)
+    r_lut = LUT2d(r_axis, t_axis, r_out, method=method, b_error=b_error)
+    return t_lut, r_lut
--- a/python/packages/nisar/cal/qfsp_slip.py
+++ b/python/packages/nisar/cal/qfsp_slip.py
@@ -2,6 +2,7 @@ from enum import Flag, unique
 import numpy as np
 from typing import Sequence
 
+from isce3.core import LUT2d
 from nisar.products.readers.instrument import InstrumentParser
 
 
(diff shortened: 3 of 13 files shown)


python/packages/isce3/geometry/rdr2rdr.py
```diff
@@ -70,3 +71,79 @@ def rdr2rdr(t: float,
...
+        Orbit defining radar motion on output path.  Defaults to input `orbit`.
+    rdr2geo_params : dict, optional
+        Dictionary specifying convergence parameters for rdr2geo solver.
+        Keys among {"tol_aztime", "look_min", "look_max"}
+        See isce3.geometry.rdr2geo_bracket
+    geo2rdr_params: dict, optional
+        Dictionary specifying convergence parameters for geo2rdr solver.
+        Keys among {"tol_height", "time_start", "time_start"}
+        See isce3.geometry.geo2rdr_bracket
+    method: str
```
bhawkins:
```suggestion
    method: str, optional
```

python/packages/isce3/geometry/rdr2rdr.py
```diff
@@ -70,3 +71,79 @@ def rdr2rdr(t: float,
...
+        LUT2d interpolation method
+    b_error: bool
```
bhawkins:
```suggestion
    b_error: bool, optional
```

python/packages/nisar/workflows/focus.py
```diff
@@ -1836,6 +1858,19 @@ def focus(runconfig, runconfig_path=""):
...
+        del opts["dtype"]
+        mask = slc.create_anomaly_mask(frequency, shape=og.shape, **opts)
+        if instparser is not None:
+            log.info(f"Writing inputDataExceptionMask for frequency{frequency}")
+            nisar.cal.qfsp_slip.write_anomaly_mask(anomaly_code, mask,
+                og.sensing_times, og.slant_ranges, tn_lut, rn_lut, el_lut,
+                instparser)
+        else:
+            log.warning("Internal calibration (INT_CAL) file was not provided "
+                "so unable to populate inputDataExceptionMask")
```
bhawkins, line 1872: Need to think about the interaction with EAP on/off.

bhawkins (commented):
Only got part of the way through this PR in our meeting today. Comments so far:

python/extensions/pybind_isce3/core/LUT2d.cpp
```diff
@@ -138,7 +139,8 @@ void addbinding(py::class_<LUT2d<T>> &pyLUT2d)
             return self.data().map();
         })
         .def("eval", py::overload_cast<const double, const double>(&LUT2d<T>::eval, py::const_))
-        .def("eval", py::overload_cast<double,const Eigen::Ref<const Eigen::VectorXd>&>(&LUT2d<T>::eval, py::const_))
+        .def("eval", py::overload_cast<double,const const_vec_t&>(&LUT2d<T>::eval, py::const_))
```
rad-eng-59: duplicated `const`?
```suggestion
        .def("eval", py::overload_cast<double, const_vec_t&>(&LUT2d<T>::eval, py::const_))
```

python/packages/isce3/geometry/rdr2rdr.py
```diff
@@ -70,3 +71,79 @@ def rdr2rdr(t: float,
...
+        Enable LUT2d bounds error checking
```
rad-eng-59:
```suggestion
        Enable LUT2d bounds error checking. Defaults to True.
```

python/packages/isce3/geometry/rdr2rdr.py
```diff
@@ -70,3 +71,79 @@ def rdr2rdr(t: float,
...
+    rdr2geo_params : dict, optional
+        Dictionary specifying convergence parameters for rdr2geo solver.
+        Keys among {"tol_aztime", "look_min", "look_max"}
+        See isce3.geometry.rdr2geo_bracket
+    geo2rdr_params: dict, optional
+        Dictionary specifying convergence parameters for geo2rdr solver.
+        Keys among {"tol_height", "time_start", "time_start"}
+        See isce3.geometry.geo2rdr_bracket
+    method: str
+        LUT2d interpolation method
```
rad-eng-59:
```suggestion
        LUT2d interpolation method. Defaults to "bilinear".
```

python/packages/nisar/cal/qfsp_slip.py
```diff
@@ -0,0 +1,172 @@
...
+from isce3.core import LUT2d
+from nisar.products.readers.instrument import InstrumentParser
+
+
+@unique
+class AnomalyCode(Flag):
+    """
+    NISAR data anomaly codes (bit flags)
+    """
+    NO_ANOMALY = 0
```
rad-eng-59, line 14: FYI, L0B processor is using the same integer values for the same purpose for the first 6 fields but with more concise names (see below). Given our L0B process does not process S-SAR,  it does not need to cover new fields. However, for the sake of consistency between names, we should perhaps request a change to L-SAR L0B to use the same dataclass. Here is the similar class in the current L0B process:
```python
class QfspOutlierId(Flag):
    """NISAR L-band qFSP outlier identifier w/ bitwise operation"""
    NA = 0
    H0 = auto()
    H1 = auto()
    H2 = auto()
    V0 = auto()
    V1 = auto()
    V2 = auto()
```
Btw, for bitwise enum.Flag, the `auto()` will automatically generates the power of 2 values in the ascending order starting from `2^0`.

python/packages/nisar/cal/qfsp_slip.py
```diff
@@ -0,0 +1,172 @@
...
+def abs2(z):
+    return z.real**2 + z.imag**2
+
+
+# Type hints: (start, end) of elevation (EL) angle interval
+ELAngleInterval = tuple[float, float]
+# There can be multiple intervals associated with each anomaly.
+Boundaries = dict[AnomalyCode, Sequence[ELAngleInterval]]
+
+def get_qfsp_mask_boundaries(anomaly_code: Union[AnomalyCode, int],
```
rad-eng-59, line 34: As discussed in the meeting, you can use Antenna pattern HDF5 product whom used to generate AC table in instrument HDF5. The respective antenna parser will have some very useful methods. For instance, to get EL angle of the peak for all beams and their common AZ boresight angle, `AntennaParser` has this method handy:
```python
el_peaks, az_boresight = AntennaParser(hdf5file).locate_beams_peak(pol)
```
FYI, regardless of the EL angle resolution of the patterns, the actual peak locations are locally and precisely estimated in this antenna method. AC table have a EL angle resolution of `50 mdeg`.

python/packages/nisar/cal/qfsp_slip.py
```diff
@@ -0,0 +1,172 @@
...
+    tn_lut, rn_lut : isce3.core.LUT2d
+        Reskew tables providing native Doppler time and range as functions of
+        zero-Doppler (time, range).
+    el_lut : isce3.core.LUT2d
+        Table providing antenna EL angle as a function of native Doppler
+        (time, range).
+    int_cal : InstrumentParser
+        NISAR LSAR INT_CAL file containing the angle-to-coefficient (AC) tables.
+    """
+    nt = len(t0_axis)
```
rad-eng-59:
```suggestion
    nt = t0_axis.size
```

python/packages/nisar/cal/qfsp_slip.py
```diff
@@ -0,0 +1,172 @@
...
+    nr = len(r0_axis)
```
rad-eng-59:
```suggestion
    nr = r0_axis.size
```

python/packages/nisar/cal/qfsp_slip.py
```diff
@@ -0,0 +1,172 @@
...
+    if dataset.shape[0] != nt:
+        raise ValueError("Mask shape[0] is incompatible with time axis length")
+    if dataset.shape[1] != nr:
+        raise ValueError("Mask shape[1] is incompatible with range axis length")
+
+    # Full image mask may be too big to fit in memory.  If it's an HDF5 dataset
+    # then use chunk size as azimuth block size.
+    block_size = dataset.shape[0]
```
rad-eng-59:
```suggestion
    block_size = nt
```

python/packages/nisar/cal/qfsp_slip.py
```diff
@@ -0,0 +1,172 @@
...
+    NISAR data anomaly codes (bit flags)
+    """
+    NO_ANOMALY = 0
+    SLIP_QFSP_H0 = 1 << 0
+    SLIP_QFSP_H1 = 1 << 1  # only one seen in LSAR as of 2026-03-27
+    SLIP_QFSP_H2 = 1 << 2
+    SLIP_QFSP_V0 = 1 << 3
+    SLIP_QFSP_V1 = 1 << 4
+    SLIP_QFSP_V2 = 1 << 5
+    SLIP_SSAR = 1 << 6
```
rad-eng-59, line 21: Not sure why we need to use `S-SAR`or even `RESERVED` given they are not properly being handled.

python/packages/nisar/cal/qfsp_slip.py
```diff
@@ -0,0 +1,172 @@
...
+
+    # Before doing anything else, check for no-anomaly since we can return
+    # early in that case.
+    if len(boundaries) == 0:
+        for block_start in range(0, nt, block_size):
+            block = slice(block_start, min(block_start + block_size, nt))
+            nb = block.stop  - block.start
+            buf = np.zeros((nb, dataset.shape[1]), dataset.dtype)
+            buf[...] = AnomalyCode.NO_ANOMALY.value
+            write_block(buf, np.s_[block, :])
```
rad-eng-59, line 151: This can be any of three possibilities rather than `NO_ANOMALY`.
I would use anomaly_code as it is being passed by L0B process which for L-SAR it will automatically be `0` when there is no qFSP issue while it preserves the value for any other issue down the line to be handled accordingly in the future. Otherwise, I don't see the point of defining any other anomalies than qFSP in the respective enum.
That being said:

```suggestion
    if len(boundaries) == 0:
        for block_start in range(0, nt, block_size):
            block = slice(block_start, min(block_start + block_size, nt))
            nb = block.stop  - block.start
            buf = np.zeros((nb, dataset.shape[1]), dataset.dtype)
            buf[...] = AnomalyCode(anomaly_code).value
            write_block(buf, np.s_[block, :])
```

python/packages/nisar/cal/qfsp_slip.py
```diff
@@ -0,0 +1,172 @@
...
+        Zero-Doppler range axis associated with RSLC image.
+    tn_lut, rn_lut : isce3.core.LUT2d
+        Reskew tables providing native Doppler time and range as functions of
+        zero-Doppler (time, range).
+    el_lut : isce3.core.LUT2d
+        Table providing antenna EL angle as a function of native Doppler
+        (time, range).
+    int_cal : InstrumentParser
+        NISAR LSAR INT_CAL file containing the angle-to-coefficient (AC) tables.
+    """
```
rad-eng-59, line 121:
```suggestion
    Notes
    -------
    This function generates invalid mask simply for 12-channel L-SAR product with qFSP sample slip anomaly.
    """
```

python/packages/nisar/cal/qfsp_slip.py
```diff
@@ -0,0 +1,172 @@
...
+
+# Type hints: (start, end) of elevation (EL) angle interval
+ELAngleInterval = tuple[float, float]
+# There can be multiple intervals associated with each anomaly.
+Boundaries = dict[AnomalyCode, Sequence[ELAngleInterval]]
+
+def get_qfsp_mask_boundaries(anomaly_code: Union[AnomalyCode, int],
+                             int_cal: InstrumentParser) -> Boundaries:
+    """
+    Determine EL angle intervals associated with NISAR anomaly codes.
```
rad-eng-59:
```suggestion
    Determine EL angle intervals covering qFSP transition regions of 12-channel (three qFSPs) L-band NISAR associated with qFSP sample slip anomaly codes.
```

python/packages/nisar/cal/qfsp_slip.py
```diff
@@ -0,0 +1,172 @@
...
+    peak_angles = dict()
+    for rxpol in ("H", "V"):
+        coeff = int_cal.get_angle2coef(rxpol)
+        angles = int_cal.el_angles_ac(rxpol)
+        peak_indices = np.argmax(abs2(coeff), axis=1)
+        # TODO Could interpolate to find peak.
+        peak_angles[rxpol] = np.array([angles[i, j] for (i, j) in
+            enumerate(peak_indices)])
+
+    # (start, end) EL angles between beam x and y peaks
```
rad-eng-59, line 64: Perhaps it is worth checking the size of the arrays prior to slicing assuming `12` channels.
```suggestion
    # (start, end) EL angles between beam x and y peaks
    if (peak_angles['H'].size != 12 or peak_angles['V'].size != 12):
        raise ValueError(f'Number of channels for H/V is {peak_angles['H'].size}/{peak_angles['V'].size} instead of 12 expected for L-SAR NISAR!')
```

python/packages/nisar/workflows/focus.py
```diff
@@ -1836,6 +1858,19 @@ def focus(runconfig, runconfig_path=""):
                                         extended_radar_grid.slant_ranges,
                                         beta0_lut, sigma0_lut, gamma0_lut)
 
+        # Set anomaly mask. Need to do some geometry.
+        opts = get_dataset_creation_options(cfg, og.shape)
+        del opts["dtype"]
+        mask = slc.create_anomaly_mask(frequency, shape=og.shape, **opts)
+        if instparser is not None:
+            log.info(f"Writing inputDataExceptionMask for frequency{frequency}")
```
rad-eng-59, line 1866: Perhaps one can check for `EAP` and if it is `OFF` issue a warning that the mask values can be invalid due to `EAP=OFF`. 
I suggest we simply generate a qFSP sample-slip mask when *EAP=True* which also guarantees that both antenna and instrument HDF5 are available. In case of *EAP=False*, a warning is issued w/ empty dataset.

python/packages/nisar/cal/qfsp_slip.py
```diff
@@ -0,0 +1,172 @@
...
+    # NOTE This converts the dtype to float64, but that's okay as long as there
+    # are fewer than 52 bits in the mask (mantissa of float64).
+    mask_lut = LUT2d(el_lut.x_axis, el_lut.y_axis, el_mask, method="nearest",
+                     b_error=False)
+
+    for block_start in range(0, nt, block_size):
+        block_end = min(block_start + block_size, nt)
+        nb = block_end - block_start
+        # Allocate each chunk to avoid HDF5 I/O as much as possible.
+        mask_chunk = np.zeros((nb, nr), dataset.dtype)
```
rad-eng-59:
```suggestion
        mask_chunk = np.full(fill_value=AnomalyCode.NO_ANOMLAY.value, shape=(nb, nr), dtype=dataset.dtype)
```

rad-eng-59 (approved):
LGTM. Thanks.
The code still doesn't warn or handle the case when `EAP=False`.
I wonder what is the decision on that. I did put my suggestion in the comment earlier.

hfattahi (approved):
Thank you @bhawkins for this PR. We reviewed the PR offline and I have tested on a pair of datasets with and without qFSP slips. The mask makes sense. One item is that I had to use a 10x finer grid for Doppler to get a mask which would finely change if topography variation exists. Shall we update the PSF for finer spacing of Doppler grid?

Here is the new mask with default settings:

Image

and here is the one with finer posting of the Doppler LUT:

Image

PR #247, opened 2026-04-02
Incorporate the qFSP mask into InSAR mask

xhuang-jpl:
This PR is to incorporate the qFSP mask into the InSAR mask, where many places are updated including:

1. Add the hasInputDataException to the identification group
2. Incorporate the qSFP mask into mask in the InSAR dataset
3. Update the `geocode_insar.py`
4. Update the `interpret_subswath_mask` function


python/packages/nisar/products/insar/utils.py
```diff
@@ -555,20 +569,44 @@ def generate_insar_subswath_mask(ref_rslc_obj,
...
                                             ref_subswaths,
                                             sec_subswaths)
-                subswath_mask.append(subswath_mask_id)
+
+                    # reference RSLC input exception mask id
+                    ref_input_exception_mask_id = ref_input_exception_mask[int(i),int(j)] << 16
+
+                    # secondary RSLC input  exception mask id
+                    sec_i = int(int(i) + azimuth_off[0,int(j)] + 0.5)
+                    sec_j = int(int(j) + range_off[0,int(j)] + 0.5)
```
xhuang-jpl: Use round() function here

python/packages/nisar/products/insar/utils.py
```diff
@@ -555,20 +569,44 @@ def generate_insar_subswath_mask(ref_rslc_obj,
...
+                    # mask id
+                    mask_id = subswath_mask_id | ref_input_exception_mask_id | sec_input_exception_mask_id
+
+                # append the mask id
+                mask.append(mask_id)
+
         # The azimuth index is not in the radar grid meaning no subswath mask
         else:
-            subswath_mask += [0] * len(rg_idx_arr)
+            mask += [0] * len(rg_idx_arr)
```
xhuang-jpl, line 609: may use the fillvalue  (255) here

hfattahi:
We reviewed this PR in a PR review session. Overall it looks good pending testing and very minor comments provided offline.

hfattahi (approved):
LGTM! Thank you @xhuang-jpl for this addition. The PR looks good and I have extensively tested with real data. I do understand that we are making a change to an existing dataset in the InSAR products. We had two choices to make either add couple of extra mask datasets or modify the current mask. After detailed evaluation we decided to modify the existing mask in a largely backward compatible manner to ensure user convenience. We understand that users may need to slightly modify their InSAR readers to read mask but with this modification one can use one unified parser to parse mask dataset from both new and old products. 

Here is an example code snippet for parsing RUNW mask from both old and new mask dataset. 

```
def parse_mask(mask_value):
    """
    Decodes a RUNW/RIFG mask (8 bits or 32 bits).
    """
    # regardless if the mask_value is 8 or 32 bits, we convert to 32 bits
    mask_value = mask_value & 0xFFFFFFFF
    
    # Bits 0–7: Subswath encoding
    subswath_byte = mask_value & 0xFF
    
    ref_subswath = (subswath_byte // 10) % 10
    sec_subswath = subswath_byte % 10
    
    # Bits 8–15: Secondary RSLC anomaly flags
    sec_anomaly = (mask_value >> 8) & 0xFF

    # Bits 16–23: Reference RSLC anomaly flags
    ref_anomaly = (mask_value >> 16) & 0xFF

    # Bits 24–31: Reserved for future use
    reserved = (mask_value >> 24) & 0xFF

    return {
        "ref_subswath": ref_subswath,
        "sec_subswath": sec_subswath,
        "sec_anomaly_raw": sec_anomaly, # Binary string for bitwise flags
        "ref_anomaly_raw": ref_anomaly, # Binary string for bitwise flags
        "is_valid_subswath": (ref_subswath > 0) & (sec_subswath > 0),
        "anomaly_mask": (ref_anomaly == 0) & (sec_anomaly == 0)
    }
```

The code snippet requires small modification for GUNW which has extra water mask. 
In a future PR we will add an ionospheric phase mask to 1 bit of the reserved bits.

python/packages/nisar/products/insar/utils.py
```diff
@@ -494,16 +494,18 @@ def generate_dem_rdr(radar_grid_obj,
...
+                        sec_rslc_h5_obj,
+                        range_offset_path,
+                        azimuth_offset_path,
+                        freq,
+                        azi_idx_arr,
+                        rg_idx_arr):
 
     """
-    Generate the InSAR subswath 2d array mask
+    Generate the InSAR  2d array mask
```
seongsujeong:
```suggestion
    Generate the InSAR 2d array mask
```
just a nit pick

seongsujeong (approved):
LGTM.

PR #251, opened 2026-04-10
PTA: Add chip to outputs and estimate signal-to-clutter ratio.

Tyler-g-hudson:
This PR adds the PTA chip as a PTA output, and incorporates a method of calculating signal-to-clutter (SCR) ratio in dB. This value would be useful for populating error bars in PTA trending.

### DIFF
--- a/python/packages/isce3/cal/point_target_info.py
+++ b/python/packages/isce3/cal/point_target_info.py
@@ -621,7 +621,7 @@ def analyze_point_target(
             peak_magnitude=results[0]["magnitude"],
         )
 
-        results[0]["signal_clutter_ratio"] = scr
+        results[0]["signal clutter ratio"] = scr
 
     return results
 
@@ -1134,11 +1134,20 @@ def tofloatvals(x):
 
     Modifies the dictionary in-place and returns None.
     """
+    def list2floats(my_list):
+        """
+        Convert a list of any dimensions into a list of floats. The list must either
+        contain only other lists or only values that can be converted into floats.
+        """
+        if all(isinstance(xi, list) for xi in my_list):
+            return [list2floats(xi) for xi in my_list]
+        return [float(xi) for xi in my_list]
+
     for k in x:
         if type(x[k]) == dict:
             tofloatvals(x[k])
         elif type(x[k]) == list:
-            x[k] = [float(xki) for xki in x[k]]
+            x[k] = list2floats(x[k])
         else:
             x[k] = float(x[k])
 
--- a/python/packages/isce3/cal/point_target_info.py
+++ b/python/packages/isce3/cal/point_target_info.py
@@ -986,7 +986,7 @@ def estimate_scr(
         )
         return np.nan
 
-    # Create an estimated chip that is the column and row of the peak magnitude plus
+    # Create an estimation chip that is the column and row of the peak magnitude plus
     # clutter_half_width pixels in either direction, converted to units of linear power.
     scr_est_chip = np.abs(
         chip[
--- a/python/packages/isce3/cal/point_target_info.py
+++ b/python/packages/isce3/cal/point_target_info.py
@@ -966,7 +966,7 @@ def estimate_scr(
 
     # If the peak magnitude was not passed in, estimate it.
     if peak_magnitude is None:
-        peak_magnitude = np.nanmax(chip)
+        peak_magnitude = np.nanmax(np.abs(chip))
 
     # Get the location of the peak magnitude.
     k = np.nanargmax(np.abs(chip))
--- a/python/packages/isce3/cal/point_target_info.py
+++ b/python/packages/isce3/cal/point_target_info.py
@@ -610,6 +610,10 @@ def analyze_point_target(
         pixel_spacing=pixel_spacing,
     )
 
+    # XXX: The current method of estimating SCR does not work for geolocated products
+    # which may have skewed sidelobes. Doing this would require resampling these
+    # products to acquire a chip usable by the SCR estimator, or else finding some new
+    # means of estimating and masking the locations of the side lobes on the chip.
     if geo_heading is None:
 
         scr = estimate_scr(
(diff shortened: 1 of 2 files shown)


Tyler-g-hudson:
This is a new calculation for me, so I'd like for folks to read carefully over the code and let me know if I've done anything wrong.

@bhawkins There was a point where you said that the peak power and clutter power should be multiplied by the resolution area - these terms canceled out when calculating the ratio so I removed them to simplify the code. Let me know if I have done this in error.

Tyler-g-hudson:
@nemo794 This is the PR I was discussing with you earlier today.

python/packages/isce3/cal/point_target_info.py
```diff
@@ -914,6 +931,92 @@ def get_slice_indices(n: int) -> np.ndarray:
...
+        raise ValueError("clutter_half_width must be a positive number greater than 2.")
+    if clutter_half_width > chip.shape[0] / 2 or clutter_half_width > chip.shape[1] / 2:
+        raise ValueError(
+            f"clutter_half_width of {clutter_half_width} must be less than half of "
+            f"chip dimensions {chip.shape[0]} and {chip.shape[1]}."
+        )
+
+    # If the peak magnitude was not passed in, estimate it.
+    if peak_magnitude is None:
+        peak_magnitude = np.nanmax(chip)
```
JensReimann: This should be 
    np.nanmax(np.abs(chip))
right? 
"chip" is a complex arrray and one needs to calculate the magnitude first or np.nanmax would look for the index with maximum real or imaginary component.

Tyler-g-hudson: Good catch! Fixed!

JensReimann:
Skipping the multiplication with resolution area should be okay. It cancels as you correctly stated. The ratio in 1014 must be unitless.

python/packages/isce3/cal/point_target_info.py
```diff
@@ -914,6 +931,92 @@ def get_slice_indices(n: int) -> np.ndarray:
     return return_dict, None
 
 
+def estimate_scr(
+    chip,
+    clutter_half_width: int = 12,
```
rad-eng-59, line 936: Wouldn't it be better to define it as fraction of the half input chip size?
Or alternatively, set the default to `None` and determine its value based on input chip rather than  setting it to absolute value which may not work for smaller chirp size than 32.

rad-eng-59 (commented):
The methodology can be improved with more flexibility by doing the following, IMO.

Baseband RSLC within the 2-D chirp in both direction and get  the  Range/AZ impulse response as we normally do for PTA analysis. Then, multiply the two response to create 2-D response. Do a proper peak normalization by using peak of 2-D impulse response assuming `SNR>>10 dB`.
Finally, subtract that deterministic non-random signal from chip box and calculate variance (not sum of square of samples!) over 2-D  residual within chip given random clutter/background noise.
This will give you clutter plus additive noise not just clutter.
Note that the assumption here is that CR impulse response is non-random that affects the DC in both range and AZ over the 2-D chip while the clutter+noise is random with zero mean but some non-zero variance that you are after.

No need to define `clutter_half_chip` or check for boundaries. The hidden assumption here is that background clutter for any chip size  whether it is centered around CR or not, are pretty homogeneous.

Edit: Alternatively, perform EVD with rank equals to half of chip length (or less but larger than 4) and find min Eigen value to represent clutter+noise power.

python/packages/isce3/cal/point_target_info.py
Tyler-g-hudson: This is something I hoped to discuss in PR review - my understanding of the push-and-pull affecting this parameter is that it should be wide enough to capture a reasonable cross-section of the clutter while not being so large that it captures terrain-related variation in magnitude, per an offline discussion with @bhawkins.

With the above having been said, it's not obvious to me that varying SCR estimation chip size by the broader PTA chip is the right method unless the PTA chip is pretty small relative to the optimal SCR estimation chip, which makes this analysis harder to do in general.

Maybe the better option is to acquire this chip from the SLC instead of the PTA chip, using the known offsets from PTA to select a clutter estimation chip directly from the SLC so we don't have to worry about edge cases like when the PTA chip is small.

JensReimann:
> The methodology can be improved with more flexibility by doing the following, IMO.
> 
> Baseband RSLC within the 2-D chirp in both direction and get the Range/AZ impulse response as we normally do for PTA analysis. Then, multiply the two response to create 2-D response. Do a proper peak normalization by using peak of 2-D impulse response assuming `SNR>>10 dB`. Finally, subtract that deterministic non-random signal from chip box and calculate variance (not sum of square of samples!) over 2-D residual within chip given random clutter/background noise. This will give you clutter plus additive noise not just clutter. Note that the assumption here is that CR impulse response is non-random that affects the DC in both range and AZ over the 2-D chip while the clutter+noise is random with zero mean but some non-zero variance that you are after.
> 
> No need to define `clutter_half_chip` or check for boundaries. The hidden assumption here is that background clutter for any chip size whether it is centered around CR or not, are pretty homogeneous.
> 
> Edit: Alternatively, perform EVD with rank equals to half of chip length (or less but larger than 4) and find min Eigen value to represent clutter+noise power.

I'm not sure about the approach you described above. The main assumption of the current approach is that the clutter is homogeneous in the region around the target. This assumption is made whenever statistics (e.g. the mean value) are used to calculate the clutter contribution.
Canceling the target response only improves the situtation as it increases the number of samples used for the statistic. However, the estimate may deteriorate if not all of the target energy is canceled, i.e., due to ionospheric effects like scintillation.

One alternative is a non-statistical approach in which the clutter at the pixels with the maximum magnitude is estimated, and then the signal-to-clutter ratio is calculated using those pixels. However, I am concerned that the clutter estimation will be contaminated by the remaining target energy, especially since it is in a region of the IRF with high SCR, where a significant amount of energy must be canceled.

PR #252, opened 2026-04-11
Add guide and configurable outlier removal to low-pass filtering

oberonia78:
This PR updates the ionosphere low-pass filtering workflow to make gap filling more robust in noisy scenes and to make the smoothing/outlier behavior configurable.

The main issue in the previous implementation was that interpolation could propagate noisy patterns from valid pixels into invalid regions before the final low-pass filtering step. This PR introduces a smoother guide image for filling, optional robust outlier rejection before filling, and configurable parameters for guide construction and outlier detection.

This PR does not affect on the areas over not-masked areas. It improves only areas where pixels are masked and their neighbor pixels. 

**Main changes**
Update low_pass_filter to improve stability in noisy regions by introducing:

NaN-aware guide smoothing (nan_aware_gaussian)
Optional robust outlier removal before filling
Configurable parameters for guide and outlier detection

**Changes**

- Add parameters:
   - guide_filter_method
  - guide_median_size
  - outlier_threshold
  - outlier_min_scale

- Build guide image using:
  - median + Gaussian (default)
  - Gaussian only
  - or no smoothing

- Apply MAD-based outlier detection (optional)
- Skip outlier computation when disabled
- Preserve original valid pixels after filling

Previous: 
image

Current:
image

ionosphere difference (old - new):
image


python/packages/isce3/atmosphere/ionosphere_filter.py
```diff
@@ -188,30 +326,75 @@ def low_pass_filter(
...
+    """
+    valid = np.isfinite(image).astype(float)
+    image0 = np.nan_to_num(image, nan=0.0)
+
+    smooth_num = gaussian_filter(image0, sigma=sigma)
+    smooth_den = gaussian_filter(valid, sigma=sigma)
+
+    out = np.full_like(image, np.nan, dtype=float)
+    good = smooth_den > 1e-6
+    out[good] = smooth_num[good] / smooth_den[good]
```
hfattahi, line 399: This reminds me of a similar attempt we had a decade ago to avoid the more complicated iterative filtering;) At that time I got mixed results with very good results and very bad results. So I gave up at that time for the reasons I explain below. But I hope this algorithm is improved and will lead to consistent improvements in the filtered data.
 
I'm a bit concerned about this approximation. I call it approximation because of the following:
Let's say our noisy data is F and we have a Mask M with 0, 1 values. Then the masked data is :
Fm = F.M 
Where Fm is the masked data. When we filter the masked data with a filter kernel G we have :

G*Fm = G * (F.M)
Where * is convolution. 

However the problem is the filtering artifacts at the edges of the masked regions. We would like to have filtered data without those edge artifacts. One could rewrite the equation above:

G*(F.M) ~ G*F . G*M

And we are interested in the original filtered data without the ringing introduced by the filtered mask. So we could normalize or deconvolve (even though people may distinguish between the two, but to me it is decnvolution!):

G*F ~ G*(F.M)/G*M

So as you see it is an approximation and can only be exact if F locally (within the smoothing kernel) is constant. Variation of F within the smoothing window starts violating the equation and the more high frequency in the window, diverges the equation from exact.

hfattahi (commented):
Thanks @oberonia78 . We will review this PR in more details shortly, but here is my main comment so far.

python/packages/isce3/atmosphere/ionosphere_filter.py
```diff
@@ -28,6 +29,11 @@ def __init__(self,
                  sig_y,
                  iteration=1,
                  filling_method='nearest',
+                 guide_filter_method="median_gaussian",
+                 guide_median_size=3,
+                 outlier_threshold=3.5,
+                 outlier_min_scale=0.0,
+                 mad_scale_factor=1.4826,
```
xhuang-jpl, line 36: where is the mad_scale_factor 1.4826 from? why we need the scale factor?

python/packages/isce3/atmosphere/ionosphere_filter.py
hfattahi: We discussed this PR offline. While the theoretical details above is correct, but the intention here is not to replace this approach with independent gap filling algorithm in place. The existing workflow interpolates unfiltered data to fill gaps based on masks and the current PR changes that by first filtering the ionospheric phase over valid pixels only and then fill the gaps using filtered data. The guided filter above is used only in the initial filtering of the valid pixels. So this is fine with me.

hfattahi (approved):
LGTM! The results from this PR has over all improved the filtering of ionospheric phase. The added functionality is configurable and this addition allows to be tested at scale.

python/packages/isce3/atmosphere/ionosphere_filter.py
```diff
@@ -212,6 +357,49 @@ def low_pass_filter(
...
+    Returns
+    -------
+    out : numpy.ndarray
+        Smoothed array of the same shape as `image`.
+    """
+    valid = np.isfinite(image).astype(float)
+    image0 = np.nan_to_num(image, nan=0.0)
+
+    smooth_num = gaussian_filter(image0, sigma=sigma)
+    smooth_den = gaussian_filter(valid, sigma=sigma)
```
xhuang-jpl, line 402: gaussian filter does not need a filer size?

python/packages/isce3/atmosphere/ionosphere_filter.py
```diff
@@ -212,6 +357,49 @@ def low_pass_filter(
...
+
+    out = np.full_like(image, np.nan, dtype=float)
+    good = smooth_den > 1e-6
+    out[good] = smooth_num[good] / smooth_den[good]
```
xhuang-jpl, line 409: what does this  **out[good] = smooth_num[good] / smooth_den[good]** do?

xhuang-jpl (commented):
Thank @oberonia78 , it looks good to me. Few more questions, and then I will approve it.

python/packages/isce3/atmosphere/ionosphere_filter.py
oberonia78: `smooth_num` is the Gaussian-smoothed sum of nearby pixel values, but missing pixels contribute 0
`smooth_den` is the Gaussian-smoothed sum of nearby validity weights, effectively telling you how much real data contributed.
Dividing renomalizes the result so Nan gaps do not artificially pull the average down. We are smoothing only good pixels.

python/packages/isce3/atmosphere/ionosphere_filter.py
oberonia78: This scipy function does not require the filter size.

python/packages/isce3/atmosphere/ionosphere_filter.py
oberonia78: This `mad_scale_factor` is a  scale factor that converts MAD (median absolute deviation) into a sigma-like robust spread estimate for outlier detection.
For a standard normal variable Z ~ N(0, 1),  median(|Z|) = Phi^{-1}(0.75) ≈ 0.67449,
so std ≈ MAD / 0.67449 ≈ 1.4826 * MAD.

oberonia78: I'll add explanation on `default`

python/packages/isce3/atmosphere/ionosphere_filter.py
xhuang-jpl: JFYI: There is a radius parameter, I am ok if you use None for this PR. 

https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.gaussian_filter.html

python/packages/isce3/atmosphere/ionosphere_filter.py
xhuang-jpl: you simply replace the NaN with 0 for the `smooth_num`, but those zeros still have the weights for the filter.  If there are zeros, the `smooth_num` might be bit under estimated. What do you think?

oberonia78: This is why we divide the soomth_num by smooth_den. 

for example, 
```

test_array = [20, 20, np.nan,  20, 20]
test_array2 = [20, 20, 0,  20, 20]

test_filtered = nan_aware_gaussian(test_array)
test_filtered2 = nan_aware_gaussian(test_array2)

print(test_filtered)
[20. 20. 20. 20. 20.]
print(test_filtered2)]
[17. 15. 14. 15. 17.]
```
As you see, this filter is not affected by NaN value which is not underestimated.

xhuang-jpl (approved):
Thank you for addressing my concerns. LGTM!

PR #253, opened 2026-04-11
Fix a bug in projecting water to cryosphere radargrid

oberonia78:
This PR fixes three issues related to the use of water masks in ionospheric processing:

1. When applying the water mask for ionosphere estimation on the main and side bands, the process failed because Frequency A was hard-coded. This PR updates the logic to dynamically select the appropriate projection based on whether the Frequency B radar grid is used.

2. The processing also failed when using water masks in EPSG:3413 or EPSG:3031, because the code incorrectly assumed an EPSG:4326 projection. This PR fixes the issue by properly handling the input projection and performing the necessary coordinate transformation.

3. `insar.py` deletes the intermediate files but some of them is required for water projection. This PR moves the steps deleting the intermediate files.


python/packages/isce3/unwrap/preprocess.py
```diff
@@ -222,66 +222,227 @@ def _get_gdal_raster_shape_type(raster_path):
...
+        raster_srs.AutoIdentifyEPSG()
+    except Exception:
+        pass
+
+    raster_epsg = raster_srs.GetAuthorityCode(None)
+    if raster_epsg is None:
+        raise RuntimeError("Could not determine raster EPSG from projection.")
+    raster_epsg = int(raster_epsg)
+
+    xmin, ymin, xmax, ymax = map(float, bbox)
```
hfattahi, line 275: Is this a type?
```suggestion
    xmin, ymin, xmax, ymax = map(bbox)
```

python/packages/isce3/unwrap/preprocess.py
```diff
@@ -222,66 +222,227 @@ def _get_gdal_raster_shape_type(raster_path):
...
+            "_read_gdal_with_bbox currently supports only north-up rasters."
+        )
+
+    raster_srs = osr.SpatialReference()
+    raster_srs.ImportFromWkt(proj)
+
+    try:
+        raster_srs.AutoIdentifyEPSG()
+    except Exception:
+        pass
```
hfattahi, line 268: what is this helping with?

python/packages/isce3/unwrap/preprocess.py
```diff
@@ -222,66 +222,227 @@ def _get_gdal_raster_shape_type(raster_path):
...
+
+    xmin, ymin, xmax, ymax = map(float, bbox)
+
+    # target SRS = bbox_epsg
+    dst_srs = osr.SpatialReference()
+    dst_srs.ImportFromEPSG(int(bbox_epsg))
+    try:
+        dst_srs.SetAxisMappingStrategy(osr.OAMS_TRADITIONAL_GIS_ORDER)
+        raster_srs.SetAxisMappingStrategy(osr.OAMS_TRADITIONAL_GIS_ORDER)
+    except Exception:
```
oberonia78, line 283: raise error?

python/packages/isce3/unwrap/preprocess.py
```diff
@@ -222,66 +222,227 @@ def _get_gdal_raster_shape_type(raster_path):
...
+        pass
```
hfattahi, line 284: Looks like the order of axis is important for us. If that is the case then you should not pass but raise an exception.

python/packages/isce3/unwrap/preprocess.py
```diff
@@ -222,66 +222,227 @@ def _get_gdal_raster_shape_type(raster_path):
...
+        corners = [
+            tx_to_raster.TransformPoint(xmin, ymin)[:2],
+            tx_to_raster.TransformPoint(xmin, ymax)[:2],
+            tx_to_raster.TransformPoint(xmax, ymin)[:2],
+            tx_to_raster.TransformPoint(xmax, ymax)[:2],
+        ]
+        xs = [p[0] for p in corners]
+        ys = [p[1] for p in corners]
+        xmin_src, xmax_src = min(xs), max(xs)
+        ymin_src, ymax_src = min(ys), max(ys)
```
hfattahi, line 299: Does this work for crossing dateline? Please see the stage_dem.py where we do similar transformations and I think we were more careful crossing dateline.

python/packages/isce3/unwrap/preprocess.py
```diff
@@ -222,66 +222,227 @@ def _get_gdal_raster_shape_type(raster_path):
...
+        p01 = tx_to_bbox.TransformPoint(x0, y0 + dy)[:2]
+
+        est_dx = abs(p10[0] - p00[0])
+        est_dy = abs(p01[1] - p00[1])
+
+        # fallback in case estimate becomes zero or numerically unstable
+        if not np.isfinite(est_dx) or est_dx <= 0:
+            est_dx = max((xmax - xmin) / 1000.0, 1e-6)
+        if not np.isfinite(est_dy) or est_dy <= 0:
+            est_dy = max((ymax - ymin) / 1000.0, 1e-6)
```
hfattahi, line 350: This does not make sense. What is going here?

python/packages/isce3/unwrap/preprocess.py
```diff
@@ -222,66 +222,227 @@ def _get_gdal_raster_shape_type(raster_path):
...
+    # For your case (bbox_epsg=4326), output grid becomes lon/lat.
+    if bbox_epsg == raster_epsg:
+        out_xres = abs(dx)
+        out_yres = abs(dy)
+    else:
+        tx_to_bbox = osr.CoordinateTransformation(raster_srs, dst_srs)
+
+        # transform two neighboring source points to estimate target resolution
+        p00 = tx_to_bbox.TransformPoint(x0, y0)[:2]
+        p10 = tx_to_bbox.TransformPoint(x0 + dx, y0)[:2]
```
oberonia78, line 340: Revise from here to line 350

python/packages/isce3/unwrap/preprocess.py
```diff
@@ -222,66 +222,227 @@ def _get_gdal_raster_shape_type(raster_path):
...
+        p01 = tx_to_bbox.TransformPoint(x0, y0 + dy)[:2]
+
+        est_dx = abs(p10[0] - p00[0])
+        est_dy = abs(p01[1] - p00[1])
```
hfattahi, line 344: Please double check and clean up.

hfattahi (commented):
Thanks @oberonia78 . Please double check how come project_map_to_radar works between 4326 and UTM and not between polar stereo and 4326! Few more comments below as we discussed in PR review.

python/packages/nisar/workflows/ionosphere.py
```diff
@@ -1205,6 +1220,20 @@ def run(cfg: dict, runw_hdf5: str):
         filling_method=filling_method,
         outputdir=os.path.join(iono_path, iono_method))
 
+    # compute the full water masks
+    water_mask_b_blk = None
+    water_mask_a_blk = None
+    if "water" in mask_type and filter_bool:
+        water_mask_path = cfg[
+            "dynamic_ancillary_file_group"]["water_mask_file"]
```
xhuang-jpl, line 1228: what if the water_mask_path is None? please do the check

python/packages/isce3/unwrap/preprocess.py
```diff
@@ -222,66 +222,227 @@ def _get_gdal_raster_shape_type(raster_path):
...
+    if warped_band is None:
+        raise RuntimeError("Failed to access warped raster band.")
+
+    arr = warped_band.ReadAsArray()
+    if arr is None:
+        raise RuntimeError("Failed to read warped raster window.")
+
+    warped_gt = warped_ds.GetGeoTransform()
+    if warped_gt is None:
+        raise RuntimeError("Warped raster geotransform is missing.")
```
xhuang-jpl, line 385: In addition to raising the RuntimeError, please use journal to log those error.

PR #256, opened 2026-04-20
Add an optional azimuth offset filter after the outlier detection for the rubbersheeting

xhuang-jpl:
#### Description                
                                                                                      
  Adds an optional azimuth offset filter to the rubbersheet workflow to suppress offset  
  jumps caused by ionospheric disturbances.
                                                                                         
  #### Changes      
                                                                                         
  - **`python/packages/nisar/workflows/rubbersheet.py`**: Implemented azimuth offset     
  filter that applies mean or median filtering in both azimuth and range directions after
   outlier detection (lines 320-331)                                                     
                  
  - **`share/nisar/schemas/insar.yaml`**: Added `azimuth_offset_filter_options` schema   
  defining `offsets_filter` (none/median/mean) and `kernel_size` parameters
                                                                                         
  - **`share/nisar/defaults/insar.yaml`**: Added `azimuth_offset_filter` configuration   
  with default settings (`offsets_filter: none`, `kernel_size: 31`)
                                                                                         
  #### Impact       

  New optional feature for improved offset quality. Default behavior unchanged           
  (`offsets_filter: none`). Users can enable by setting `offsets_filter` to 'median' or
  'mean' in the runconfig.


share/nisar/schemas/insar.yaml
```diff
@@ -550,6 +550,10 @@ rubbersheet_options:
     # outlier holes (if present)
     interpolation_method: enum('no_interpolation', 'nearest', 'linear', 'cubic', 'idw', required=False)
 
+    # Aggressive azimuth offset filter applied right after outlier detection to suppress
+    # offset jumps caused by defocus from ionospheric disturbances.
```
hfattahi, line 554:
```suggestion
    # Additional azimuth offset low-pass filtering applied after outlier detection in the rubbersheeting    #workflow to suppress offset jumps mainly caused by ionospheric disturbances.
```

python/packages/nisar/workflows/rubbersheet.py
```diff
@@ -316,6 +316,21 @@ def run_rubbersheet_with_interpolation(cfg: dict, output_hdf5: str = None):
                     offset_az_culled, offset_rg_culled = identify_outliers(
                         str(dense_offsets_dir),
                         rubbersheet_params, valid_mask)
+
+                    # Apply aggressive azimuth offset filter to suppress the offsets jupmps
```
hfattahi, line 320: see my other comment on schema

python/packages/nisar/workflows/rubbersheet.py
```diff
@@ -894,6 +926,133 @@ def _interpolate_offsets(offset, interp_method):
     return offset_interp
 
 
+def apply_filter(array, window_size, filter_type='mean', axis='azimuth'):
```
hfattahi: Looks like the default should be both based on offline discussions:
```suggestion
def apply_filter(array, window_size, filter_type='mean', axis='both'):
```

python/packages/nisar/workflows/rubbersheet.py
```diff
@@ -894,6 +926,133 @@ def _interpolate_offsets(offset, interp_method):
...
+        2D array to filter (shape: azimuth x range)
+    window_size: int or tuple
+        Size of the filtering window.
+        - If int: window size for the specified axis (or both if axis='both')
+        - If tuple: (window_size_azimuth, window_size_range) when axis='both'
+        Must be odd and >= 1.
+    filter_type: str
+        Type of filter to apply: 'mean' or 'median' (default: 'mean')
+    axis: str
+        Direction to filter: 'azimuth', 'range', or 'both' (default: 'azimuth')
```
hfattahi:
```suggestion
        Direction to filter: 'azimuth', 'range', or 'both' (default: 'both')
```

python/packages/nisar/workflows/rubbersheet.py
```diff
@@ -894,6 +926,125 @@ def _interpolate_offsets(offset, interp_method):
...
+                    mode='constant', constant_values=np.nan)
+
+    if axis == 'both':
+        # Create 2D sliding window view using stride tricks
+        # Shape will be (nrows, ncols, window_size_azimuth, window_size_range)
+        shape = (nrows, ncols, window_size_az, window_size_rg)
+        strides = (padded.strides[0], padded.strides[1], padded.strides[0], padded.strides[1])
+
+        # Create a view with 2D sliding windows (no copy, just metadata)
+        windows = np.lib.stride_tricks.as_strided(padded, shape=shape, strides=strides)
```
seongsujeong: FYI: https://github.com/scipy/scipy/issues/20012

seongsujeong (commented):
I did a bit of search, and looks like `sliding_window_view` is preferred than `as_stride` in terms of safety. I've left more information in the comment.

python/packages/nisar/workflows/rubbersheet.py
```diff
@@ -894,6 +926,133 @@ def _interpolate_offsets(offset, interp_method):
...
+        windows_flat = windows.reshape(nrows, ncols, -1)
+
+        # Apply filter function along the flattened window axis (axis=2)
+        with warnings.catch_warnings():
+            warnings.filterwarnings('ignore', r'All-NaN slice encountered')
+            warnings.filterwarnings('ignore', r'Mean of empty slice')
+            if filter_type == 'mean':
+                filtered = np.nanmean(windows_flat, axis=2)
+            else:  # median
+                filtered = np.nanmedian(windows_flat, axis=2)
```
hfattahi: Similar to @seongsujeong I also did some independent investigation and gemini tells me that  np.lib.stride_tricks.as_strided in general is not safe and np.lib.stride_tricks.sliding_window_view is safer. Eventhough I think you have used it safely. So this may not be a big concern but I would like you to take a closer look. 
Also the other issue is np.nanmedian which can not work with virtual data but rather makes a copy of the data and depending of the size of the input, the number of copies may explode! Apparently for a 11x11 widnow size the data gets copied 121 times by np.nanmean. I think this might be still fine for azimuth offsets which are usually small. But I thought you should be aware and make sure the program won't fall short of Memory when running on our regular machines. What is the size of our azimuth offsets usually?

hfattahi (commented):
Thanks @xhuang-jpl . Overall looks good. Here are few comments, mainly minor. Perhaps the major one is the numpy functions you use for making a virtual window and the safety concerns with how memory is handled with those functions. See details below.

python/packages/nisar/workflows/rubbersheet.py
xhuang-jpl: Thank you @hfattahi and @seongsujeong. I have updated the method, instead of using the stide window which might explode the memory, I used the ndimage.generic_filter, which is memory safe and the runtime is also fine. Our original offsets size is about 1000x1000, with 31x31 window size, the runtime is about 10 secs for the mean and 40 secs for the median. As you can see below. 

image

xhuang-jpl:
@hfattahi and @seongsujeong if you are ok with the change, please take one more look. Thank you

hfattahi:
Thank you @xhuang-jpl . The changes look good to me. I know that you ran many case studies. I would like to see the results for the last pair that we discussed offline before I approve this PR.

hfattahi (approved):
LGTM! Thanks @xhuang-jpl .

python/packages/nisar/workflows/rubbersheet.py
seongsujeong, line 320: A reminder to revise the comment based on the suggestion of @hfattahi .

python/packages/nisar/workflows/rubbersheet.py
```diff
@@ -739,6 +756,21 @@ def _offset_blending(off_product_dir, rubbersheet_params, layer_keys, mask = Non
         if nan_count_az > 0:
             offset_az_culled, _ = identify_outliers(str(off_product_dir / layer_key),
                                                     rubbersheet_params, mask)
+
+            # Apply aggressive azimuth offset filter to suppress the offsets jupmps
```
seongsujeong, line 760: Suggest to revise the comment based on the suggestion from @hfattahi

seongsujeong (approved):
LGTM. Please revise the comments based on the suggestion from @hfattahi. 
Approving the PR to expedite the process.

python/packages/nisar/workflows/rubbersheet.py
xhuang-jpl: Thank you @seongsujeong. Those are code comments, it should be fine. However, for the insar.yaml in the schema and default are kept consistent.

PR #257, opened 2026-04-24
Alternative frequency domain RFI filter (tone-rank)

bhawkins:
## Summary                                                                                                                                                     

This PR implements the "tone-rank" algorithm for narrowband RFI detection and mitigation in raw SAR data, with configurable replacement strategies and diagnostic output. The main difference between the existing FDNF algorithm is the complete lack of averaging, either in slow-time or range-frequency. This probably reduces detection performance of persistent tones but allows us to mitigate intermittent noise with fewer artifacts.
                                                                                                                                                                 
  ## Key Features 

  **RFI Detection and Mitigation**                                                                                                                               
  - Spectral-domain detection using lifted exponential statistical model estimated from rank statistic.
  - STFT-based processing with COLA windowing                                                                                                                    
  - Temporal interpolation for mitigation with configurable fallback strategies                                                                                  
                                                                                                                                                                 
  **User Configuration**                                                                                                                                         
  - `interpolate` (bool): Enable/disable temporal interpolation                                                                                                  
  - `fill_value` ("noise" | "zero"): Fallback replacement strategy                                                                                               
  - Configurable detection parameters (block size, false positive rate, quantile)                                                                                
                                                                                                                                                                 
  **Diagnostic Output**                                                                                                                                          
  - HDF5 metadata with per-block statistics (ISR, mean signal power, hit rates)                                                                                  
  - Time/range/frequency axes for analysis                                                                                                                       
  - Written to `metadata/RFI/` group in RSLC product                                                                                                             
                                                                                                                                                                 
  **Integration**                                                                                                                                                
  - Integrated into focus.py workflow                                                                                                                            
  - Supports both detection-only and mitigation modes
  - Compatible with dithered and non-dithered acquisition modes                                                                                                  
                                                                                                                                                                 
  ## Files Changed                                                                                                                                               
                                                                                                                                                                 
  - `python/packages/isce3/signal/rfi_tone_rank.py`: Core algorithm implementation                                                                               
  - `python/packages/nisar/workflows/focus.py`: Workflow integration
  - `share/nisar/schemas/focus.yaml`: Schema validation                                                                                                          
  - `share/nisar/defaults/focus.yaml`: Default configuration                                                                                                     

<img width="956" height="456" alt="tone_rank_compare" src="https://github.com/user-attachments/assets/5ef081c6-325b-449e-9b69-50c14cbffe52" />

### DIFF
--- a/python/packages/isce3/signal/rfi_tone_rank.py
+++ b/python/packages/isce3/signal/rfi_tone_rank.py
@@ -32,6 +32,8 @@ def get_spectral_mask(
     # small.
     k = round(reference_quantile * n)
     kth_power = np.partition(power_spectra, k)[k]
+    if kth_power == 0.0:
+        return np.zeros(spectra.shape, dtype=bool), 0.0
     # Estimate the parameter of a lifted exponential distribution.
     λ = exp_from_quantile(reference_quantile, kth_power, bandwidth)
     # Determine threshold.  Since we've estimated a statistical model, we can
--- a/python/packages/isce3/signal/rfi_tone_rank.py
+++ b/python/packages/isce3/signal/rfi_tone_rank.py
@@ -1,3 +1,4 @@
+import h5py
 import logging
 import numpy as np
 from scipy.fft import fft, ifft, fftfreq, fftshift
@@ -245,23 +246,74 @@ def remove_loud_tones(
     fill_value="noise",
 ):
     """
-    t : np.ndarray [float64]
-        Pulse times (seconds since orbit/grid epoch).
+    Detect and optionally mitigate narrowband RFI using spectral rank method.
+
+    Processes raw data in overlapping blocks using a Short-Time Fourier Transform
+    (STFT) approach. RFI is detected in the spectral domain using a lifted
+    exponential statistical model. Detected RFI samples are replaced using
+    temporal interpolation or fill values.
+
+    Parameters
+    ----------
+    z : np.ndarray[complex64]
+        Raw data, shape (num_pulses, num_range_bins).
+    t : np.ndarray[float64]
+        Pulse times in seconds since orbit/grid epoch, length num_pulses.
     r : isce3.core.Linspace
-        Range to each sample (meters).
-    swaths : np.ndarray [int]
-        Valid subswath samples, dims = (ns, nt, 2) where ns is the number of
-        sub-swaths, nt is the number of pulses, and the trailing dimension is
-        the [start, stop) indices of the sub-swath.
-    doppler : isce3.core.LUT2d [double]
-        Raw data Doppler look up table.  Must be valid over entire grid.
+        Slant range to each sample in meters.
+    swaths : np.ndarray[int]
+        Valid subswath samples, shape (num_subswaths, num_pulses, 2).
+        Last dimension contains [start, stop) indices of each subswath.
+    doppler : isce3.core.LUT2d
+        Raw data Doppler centroid look-up table in Hz. Must be valid over
+        entire grid.
+    block_dims : tuple[int, int], optional
+        Processing block size (azimuth, range). Default is (512, 1024).
+    reference_quantile : float, optional
+        Quantile used to estimate the rate parameter of the lifted exponential
+        distribution, in [1-bandwidth, 1). Default is 0.5 (median).
+    nominal_false_positive_rate : float, optional
+        Target false positive rate for RFI detection, in (0, 1].
+    bandwidth : float, optional
+        Ratio of chirp bandwidth to sample rate, in (0, 1].
+    detect_only : bool, optional
+        If True, only detect RFI without mitigation. If False, replace
+        detected RFI samples.
+    zout : np.ndarray[complex64], optional
+        Output buffer for mitigated data. Must have same shape as z.
+        If None, z is modified in-place.
     interpolate : bool, optional
         If True, attempt linear/nearest-neighbor interpolation for RFI samples.
-        If False, replace directly with fill_value. Default is True.
+        If False, replace directly with fill_value.
     fill_value : str, optional
         Fallback value when interpolation fails or is disabled.
         "noise": use random Gaussian noise (default)
         "zero": use zero
+
+    Returns
+    -------
+    means : np.ndarray[float]
+        Estimated mean signal power per block from lifted exponential model,
+        shape (num_az_blocks, num_range_blocks). Equal to 1/λ where λ is the
+        rate parameter.
+    isr : np.ndarray[float]
+        Interference-to-signal ratio per block,
+        shape (num_az_blocks, num_range_blocks).
+    f : np.ndarray[float]
+        Normalized frequency axis for hits array, length block_dims[1].
+        Units are cycles per sample.
+    hits : np.ndarray[uint32]
+        Count of detected RFI samples at each frequency bin,
+        shape (num_az_blocks, num_range_blocks, block_dims[1]).
+
+    Notes
+    -----
+    The algorithm processes data in overlapping blocks using COLA (Constant
+    Overlap-Add) windowing in range. Each block is transformed to the spectral
+    domain where RFI tones appear as anomalously loud samples. Detection uses
+    the lifted exponential model (see get_spectral_mask). When mitigation is
+    enabled, detected samples are replaced using temporal interpolation from
+    adjacent clean pulses, with fallback to random noise or zeros.
     """
     # Check inputs
     if not (z.ndim == len(block_dims) == 2):
@@ -298,6 +350,7 @@ def remove_loud_tones(
     f = fftshift(fftfreq(block_dims[1]))
     meta_shape = (num_az_blocks, num_range_blocks)
     isr = np.zeros(meta_shape)
+    means = np.zeros(meta_shape)
     hits = np.zeros(meta_shape + (block_dims[1],), dtype=np.uint32)
 
     block_ranges = np.zeros(num_range_blocks)
@@ -349,6 +402,7 @@ def remove_loud_tones(
                 spectra[:, j, :] = fill_missing(spectra[:,j,:], fd, block_times,
                     mask_replace, mask_valid_blk, noise, interpolate, fill_value)
             hits[iblock, j, :] = fftshift(np.sum(mask_replace, axis=0))
+            means[iblock, j] = 1 / λ
         # skip inverse FFTs and assignment if not required.
         if not detect_only:
             # range inverse STFT
@@ -359,4 +413,4 @@ def remove_loud_tones(
                 nw = len(window)
                 zout[rows, cols] += z_block[:nb, j, :nw]
 
-    return isr, f, hits
+    return means, isr, f, hits
--- a/python/packages/isce3/signal/rfi_tone_rank.py
+++ b/python/packages/isce3/signal/rfi_tone_rank.py
@@ -390,6 +390,13 @@ def remove_loud_tones(
         # TODO could weight by window
         block_ranges[j] = r[(cols.start + cols.stop) // 2]
 
+    # Generate a pool of noise with σ=1 once, and we'll scale it for each
+    # block as needed.  We'll generate a random offset for each block to
+    # increase entropy without having to regenerate noise for each block.
+    max_offset = num_az_blocks * num_range_blocks
+    num_noise = max_offset + np.prod(block_dims)
+    std_noise = circular_gaussian_noise(num_noise)
+
     for iblock, rows in enumerate(az_blocks):
         block_results = []
         # calculate azimuth time of block
@@ -417,9 +424,7 @@ def remove_loud_tones(
             )
             if not detect_only:
                 σ = np.sqrt(0.5 / λ) if λ > 0.0 else 0.0
-                num_noise = min(np.sum(mask_replace),
-                    z_block.shape[1] * 991 // 97)
-                noise = circular_gaussian_noise(num_noise, σ)
+                noise = σ * std_noise[np.random.randint(0, max_offset):]
                 fd = doppler.eval(block_times[iblock], block_ranges[j])
                 cols, window = slices_windows[j]
                 nw = len(window)
--- a/python/packages/isce3/signal/rfi_tone_rank.py
+++ b/python/packages/isce3/signal/rfi_tone_rank.py
@@ -3,6 +3,7 @@ import logging
 import numpy as np
 from scipy.fft import fft, ifft, fftfreq, fftshift
 from . import cola_windows
+from isce3.noise.noise_power_est_func import cpi_slice_gen
 
 log = logging.getLogger("isce3.signal.rfi")
 
@@ -350,7 +351,8 @@ def remove_loud_tones(
 
     slices_windows = list(cola_windows(z.shape[1], block_dims[1]))
     num_range_blocks = len(slices_windows)
-    num_az_blocks = 1 + (z.shape[0] - 1) // block_dims[0]
+    az_blocks = list(cpi_slice_gen(z.shape[0], block_dims[0]))
+    num_az_blocks = len(az_blocks)
     zbshape = (block_dims[0], num_range_blocks, block_dims[1])
     z_block = np.zeros(zbshape, z.dtype)
 
@@ -366,27 +368,20 @@ def remove_loud_tones(
         # TODO could weight by window
         block_ranges[j] = r[(cols.start + cols.stop) // 2]
 
-    for iblock, block_start in enumerate(range(0, z.shape[0], block_dims[0])):
+    for iblock, rows in enumerate(az_blocks):
         block_results = []
-        # last block is smaller
-        nb = min(block_dims[0], z.shape[0] - block_start)
-        block_end = block_start + nb
-        rows = slice(block_start, block_end)
         # calculate azimuth time of block
-        block_times[iblock] = t[block_start + nb // 2]
+        block_times[iblock] = t[rows].mean()
         pulse_times = t[rows]
         # populate valid data mask
         mask_valid[...] = False
-        for i, i_pulse in enumerate(range(block_start, block_end)):
+        for i, i_pulse in enumerate(range(rows.start, rows.stop)):
             for start, end in swaths[:, i_pulse, :]:
                 mask_valid[i, start:end] = True
         # apply window to each range block
         for j, (cols, window) in enumerate(slices_windows):
             nw = len(window)
-            z_block[:nb, j, :nw] = window[None, :] * z[rows, cols]
-        # crop for last azimuth block
-        if nb < block_dims[0]:
-            z_block = z_block[:nb, ...]
+            z_block[:, j, :nw] = window[None, :] * z[rows, cols]
         # Range STFT.  Use consistent FFT size even for edges where window
         # may be shorter so that frequency metadata are consistent.
         spectra = fft(z_block, n=block_dims[1], axis=2)
@@ -406,8 +401,8 @@ def remove_loud_tones(
                 fd = doppler.eval(block_times[iblock], block_ranges[j])
                 cols, window = slices_windows[j]
                 nw = len(window)
-                mask_valid_blk = np.zeros((nb, block_dims[1]), bool)
-                mask_valid_blk[:, :nw] = mask_valid[:nb, cols]
+                mask_valid_blk = np.zeros(block_dims, bool)
+                mask_valid_blk[:, :nw] = mask_valid[:, cols]
                 spectra[:, j, :] = fill_missing(spectra[:,j,:], fd, pulse_times,
                     mask_replace, mask_valid_blk, noise, interpolate, fill_value)
             hits[iblock, j, :] = fftshift(np.mean(mask_replace, axis=0))
@@ -420,7 +415,7 @@ def remove_loud_tones(
             zout[rows, ...] = 0
             for j, (cols, window) in enumerate(slices_windows):
                 nw = len(window)
-                zout[rows, cols] += z_block[:nb, j, :nw]
+                zout[rows, cols] += z_block[:, j, :nw]
 
     return block_times, block_ranges, f, means, isr, hits
 
--- a/python/packages/isce3/signal/rfi_tone_rank.py
+++ b/python/packages/isce3/signal/rfi_tone_rank.py
@@ -27,7 +27,7 @@ def exp_from_quantile(p, vp, bw=1.0):
     Returns
     -------
     λ : float | np.ndarray
-        The rate parameter of the lifted expononential distribution.
+        The rate parameter of the lifted exponential distribution.
 
     Notes
     -----
--- a/python/packages/isce3/signal/rfi_tone_rank.py
+++ b/python/packages/isce3/signal/rfi_tone_rank.py
@@ -33,7 +33,7 @@ def get_spectral_mask(
     k = round(reference_quantile * n)
     kth_power = np.partition(power_spectra, k)[k]
     if kth_power == 0.0:
-        return np.zeros(spectra.shape, dtype=bool), 0.0
+        return np.zeros(spectra.shape, dtype=bool), 0.0, 0.0
     # Estimate the parameter of a lifted exponential distribution.
     λ = exp_from_quantile(reference_quantile, kth_power, bandwidth)
     # Determine threshold.  Since we've estimated a statistical model, we can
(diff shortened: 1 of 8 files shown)


bhawkins:
Here are some example plots of the diagnostic metadata.

tone_rank_metadata

share/nisar/schemas/focus.yaml
```diff
@@ -405,10 +405,11 @@ rfi_options:
   #   - 'ST-EVD': slow-time eigenvalue decomposition algorithm
   #   - 'FDNF': frequency domain notch filtering algorithm
   # Defaults to 'ST-EVD'.
```
Tyler-g-hudson, line 409:
```suggestion
# RFI detection algorithm. Supported values include:
#   - 'ST-EVD': slow-time eigenvalue decomposition algorithm
#   - 'FDNF': frequency domain notch filtering algorithm
#   - 'tone-rank': ...
# Defaults to 'ST-EVD'.
```
This behavior could use documentation in this schema as well as the defaults runconfig - currently I don't think users would expect it to be available.

python/packages/isce3/signal/rfi_tone_rank.py
```diff
@@ -0,0 +1,499 @@
...
+    statistical model.
+    """
+    n = np.prod(spectra.shape)
+    power_spectra = abs2(spectra)
+    power_spectra.shape = (n,)
+    # partition seems to be a little faster than quantile, I guess because the
+    # latter does a full sort and interpolates?  For large n the difference is
+    # small.
+    k = round(reference_quantile * n)
+    kth_power = np.partition(power_spectra, k)[k]
```
vbrancat, line 115: `kth_power` is the order statistic at index `k = round(reference_quantile * n)` (equivalent to `np.sort(power_spectra)[k])`, which does not match `np.quantile(..., method="linear")` in general. For example, for even values of `n` at `p=0.5` this is a single middle sample, not the average of the two middle samples.

python/packages/isce3/signal/rfi_tone_rank.py
```diff
@@ -0,0 +1,499 @@
...
+    # Estimate the parameter of a lifted exponential distribution.
+    λ = exp_from_quantile(reference_quantile, kth_power, bandwidth)
+    # Determine threshold.  Since we've estimated a statistical model, we can
+    # achieve a specified nominal false positive rate by thresholding the
+    # nominal CDF at one minus that value.
+    q = 1.0 - nominal_false_positive_rate
+    too_loud = exp_from_quantile(q, λ, bandwidth)
+    # Use threshold to generate mask.
+    mask = power_spectra >= too_loud
+    isr = np.sum(power_spectra[mask]) / np.sum(power_spectra[~mask])
```
vbrancat: If `mask` is all `True`, `np.sum(power_spectra[~mask])` is `0` and the ISR ratio blows up or becomes `inf`/`nan`. Same if nearly all power is in the masked set. May be, you want to define the ISR for these edge cases and/or document the adopted convention

Consider defining ISR for that edge case (e.g. inf, nan, or 0 with a comment) and/or documenting the convention.

python/packages/isce3/signal/rfi_tone_rank.py
```diff
@@ -0,0 +1,499 @@
...
+    ----------
+    group : h5py.Group
+        HDF5 group to write datasets into.
+    t : np.ndarray[float]
+        Azimuth times at center of each azimuth block in seconds since epoch,
+        length num_az_blocks. Written as "nativeDopplerTime" dataset.
+    r : np.ndarray[float]
+        Slant ranges at center of each range block in meters,
+        length num_range_blocks. Written as "slantRange" dataset.
+    freq : np.ndarray[float]
```
vbrancat, line 467: This is a bit confusing because in `remove_loud_tones`, `f` was normalized (in cycles per sample).

python/packages/isce3/signal/rfi_tone_rank.py
```diff
@@ -0,0 +1,499 @@
...
+        Value corresponding to the given percentile.  For example, the median
+        value.
+    bw : float, optional
+        The portion of the distribution governed by an exponential distribution.
+        See notes below.  Values in interval (0, 1].
+
+    Returns
+    -------
+    λ : float | np.ndarray
+        The rate parameter of the lifted expononential distribution.
```
vbrancat:
```suggestion
        The rate parameter of the lifted exponential distribution.
```

tests/python/packages/isce3/signal/rfi_tone_rank.py
```diff
@@ -0,0 +1,55 @@
...
+        nominal_false_positive_rate=fpr, bandwidth=1
+    )
+
+    # Check basic bookkeeping.
+    assert len(freq) == block_dims[1]
+    assert len(freq) == block_hits.shape[2]
+
+    # Make sure we mostly suppressed the correct tone.
+    i, j = 0, 10
+    many_hits = block_hits[i, j, :] > 10
```
vbrancat: I might have got it wrong but in the implementation I reviewed earlier, `hits`/`block_hits` looked like to be in "per-bin fractions" in [0, 1]. If this is the case, shouldn't this be < 1?

tests/python/packages/isce3/signal/rfi_tone_rank.py
```diff
@@ -0,0 +1,55 @@
+import numpy as np
+import numpy.testing as npt
+
+from isce3.core import LUT2d
+from isce3.signal.rfi_tone_rank import remove_loud_tones, abs2
+
+
+def test_tone_rank():
+    # repeatable tests, https://www.youtube.com/watch?v=a6iW-8xPw3k
+    np.random.seed(12345)
```
vbrancat, line 13: Just a personal opinion, but a more robust unit test would be to test the statistical core of the algorithm without noise. Something like:
- pick λ and bw
- solve for a quantile p and value v analytically (invert the lifted-exponential CDF you document)
-  assert `exp_from_quantile(p, v, bw)` recovers λ within some tolerance. 

Not sure if it is worth adding edge cases like `bw` boundary or invalid `p` values

vbrancat (commented):
This looks really cool! Just a couple of comments/inconsistencies spotted during the code review.

python/packages/isce3/signal/rfi_tone_rank.py
```diff
@@ -0,0 +1,499 @@
...
+    ----------
+    z : np.ndarray
+        Complex spectral data to fill, shape (m, n)
+    fd : float
+        Doppler centroid frequency in Hz
+    t : np.ndarray
+        Pulse times in seconds since epoch, length m
+    mask_replace : np.ndarray
+        Boolean mask marking samples to replace, shape (m, n)
+    mask_valid : np.ndarray
```
hfattahi: mask_valid is in time domain while the data is in range frequency domain. We can not use the mask as is.

python/packages/isce3/signal/rfi_tone_rank.py
```diff
@@ -0,0 +1,499 @@
...
+        # copy for modification
+        cols_need_replacement = mask_replace[i, :].copy()
+
+        if interpolate:
+            # Previous and next pulse, with reflection boundary condition.
+            iprev, inext = i - 1, i + 1
+            if i == 0:
+                iprev = i + 1
+            if i == m - 1:
+                inext = i - 1
```
bhawkins, line 204: This doesn't work if `m==1`

hfattahi, line 204: This won't work for m=1, i.e. a block of 1 pulse

python/packages/isce3/signal/rfi_tone_rank.py
```diff
@@ -0,0 +1,499 @@
...
+                num_noise = min(np.sum(mask_replace),
+                    z_block.shape[1] * 991 // 97)
+                noise = circular_gaussian_noise(num_noise, σ)
+                fd = doppler.eval(block_times[iblock], block_ranges[j])
+                cols, window = slices_windows[j]
+                nw = len(window)
+                mask_valid_blk = np.zeros((nb, block_dims[1]), bool)
+                mask_valid_blk[:, :nw] = mask_valid[:nb, cols]
+                spectra[:, j, :] = fill_missing(spectra[:,j,:], fd, pulse_times,
+                    mask_replace, mask_valid_blk, noise, interpolate, fill_value)
```
bhawkins: `mask_valid` is in the time-domain, but we're using it in the frequency domain. That doesn't make sense and requires modifying the approach. Maybe we should just not use `mask_valid` at all, since the affected samples were already nulled in the time domain. Alternatively, some kind of reduction is required like `invalid_rows = np.any(mask_valid, axis=1)`.

bhawkins (commented):
Two bugs found in today's online review.

python/packages/isce3/signal/rfi_tone_rank.py
```diff
@@ -0,0 +1,499 @@
...
+    if len(t) != m:
+        raise ValueError(f"expected len(t)=={m} got {len(t)}")
+    if mask_replace.shape != (m, n):
+        raise ValueError(f"{mask_replace.shape=} does not match {z.shape=}")
+    if mask_valid.shape != (m, n):
+        raise ValueError(f"{mask_valid.shape=} does not match {z.shape=}")
+
+    # Deramp Doppler.
+    deramp = np.exp(-1j * 2 * np.pi * fd * t).astype(z.dtype)
+    zout = deramp[:, None] * z
```
bhawkins, line 189: Since we're in the range-frequency domain we could separately deramp each bin using its own frequency. Not sure if it's worth the trouble, but the behavior should be clarified in the docstring/comments.

tests/python/packages/isce3/signal/rfi_tone_rank.py
bhawkins: Yeah, I changed the API from counts to fraction and forgot to update the unit test. I guess it didn't fail because numpy.testing doesn't complain about empty test vectors. Thanks for catching that!

python/packages/isce3/signal/rfi_tone_rank.py
bhawkins: Resolved by making the mask one boolean per row, and the user can control how much to care about gaps in the data via a new parameter in the config file.

python/packages/isce3/signal/rfi_tone_rank.py
bhawkins: Resolved by changing blocking logic to use constant-size blocks (re-using pulses at the end, if necessary). Also patched this function to work okay if m==1 for safety, but now that isn't relied upon.

python/packages/isce3/signal/rfi_tone_rank.py
bhawkins: Noted in docstring.

python/packages/isce3/signal/rfi_tone_rank.py
bhawkins: Yep! Understood. They converge as n→∞ and we're in the large n case, so I went with the faster algorithm.

python/packages/isce3/signal/rfi_tone_rank.py
```diff
@@ -0,0 +1,506 @@
...
+    # Estimate the parameter of a lifted exponential distribution.
+    λ = exp_from_quantile(reference_quantile, kth_power, bandwidth)
+    # Determine threshold.  Since we've estimated a statistical model, we can
+    # achieve a specified nominal false positive rate by thresholding the
+    # nominal CDF at one minus that value.
+    q = 1.0 - nominal_false_positive_rate
+    too_loud = exp_from_quantile(q, λ, bandwidth)
+    # Use threshold to generate mask.
+    mask = power_spectra >= too_loud
+    isr = np.sum(power_spectra[mask]) / np.sum(power_spectra[~mask])
```
bhawkins: The mask=True bins have both signal and interference. That means the interference is overestimated (since it's summing signal + noise) while the signal is underestimated (since the mask=True bins aren't counted towards signal). Ideally there'd be some arithmetic corrections here, but for this purpose maybe it's okay.

python/packages/isce3/signal/rfi_tone_rank.py
```diff
@@ -0,0 +1,506 @@
...
+
+def exp_from_quantile(p, vp, bw=1.0):
+    """
+    Determine the rate parameter of the lifted exponential distribution from a
+    quantile and its corresponding value.
+
+    Parameters
+    ----------
+    p : float | np.ndarray
+        Percentile in [1-bw, 1).  For example, 0.5 for the median.
```
bhawkins:
```suggestion
    p : float | np.ndarray
        Quantile in [1-bw, 1).  For example, 0.5 for the median.
```

python/packages/isce3/signal/rfi_tone_rank.py
```diff
@@ -0,0 +1,506 @@
...
+    vp : float | np.ndarray
+        Value corresponding to the given percentile.  For example, the median
+        value.
+    bw : float, optional
+        The portion of the distribution governed by an exponential distribution.
+        See notes below.  Values in interval (0, 1].
```
bhawkins:
```suggestion
    bw : float, optional
        The portion of the distribution governed by an exponential distribution.
        In the context of the tone-rank algorithm, it's the chirp bandwidth normalized
        by the sample rate.
        See notes below.  Values in interval (0, 1].
```

python/packages/isce3/signal/rfi_tone_rank.py
```diff
@@ -0,0 +1,506 @@
...
+    az_blocks = list(cpi_slice_gen(z.shape[0], block_dims[0]))
+    num_az_blocks = len(az_blocks)
+    zbshape = (block_dims[0], num_range_blocks, block_dims[1])
+    z_block = np.zeros(zbshape, z.dtype)
+
+    f = fftshift(fftfreq(block_dims[1]))
+    meta_shape = (num_az_blocks, num_range_blocks)
+    isr = np.zeros(meta_shape)
+    means = np.zeros(meta_shape)
+    hits = np.zeros(meta_shape + (block_dims[1],), dtype=np.float32)
```
bhawkins: Maybe these should all be float32? Not sure why two need to be double precision.

python/packages/isce3/signal/rfi_tone_rank.py
```diff
@@ -0,0 +1,506 @@
...
+        for i, i_pulse in enumerate(range(rows.start, rows.stop)):
+            for start, end in swaths[:, i_pulse, :]:
+                mask_valid[i, start:end] = True
+        # apply window to each range block
+        for j, (cols, window) in enumerate(slices_windows):
+            nw = len(window)
+            z_block[:, j, :nw] = window[None, :] * z[rows, cols]
+        # Range STFT.  Use consistent FFT size even for edges where window
+        # may be shorter so that frequency metadata are consistent.
+        spectra = fft(z_block, n=block_dims[1], axis=2)
```
bhawkins, line 416: Is this parallel by default?

python/packages/isce3/signal/rfi_tone_rank.py
```diff
@@ -0,0 +1,506 @@
...
+                spectra[:, j, :],
+                reference_quantile,
+                nominal_false_positive_rate,
+                bandwidth,
+            )
+            if not detect_only:
+                σ = np.sqrt(0.5 / λ) if λ > 0.0 else 0.0
+                num_noise = min(np.sum(mask_replace),
+                    z_block.shape[1] * 991 // 97)
+                noise = circular_gaussian_noise(num_noise, σ)
```
bhawkins: This is too cute. Just generate a bunch of standard normal noise outside the loop, generate a random offset inside the loop, and scale it by $\sigma$.

python/packages/isce3/signal/rfi_tone_rank.py
```diff
@@ -0,0 +1,506 @@
...
+                fd = doppler.eval(block_times[iblock], block_ranges[j])
+                cols, window = slices_windows[j]
+                nw = len(window)
```
bhawkins:
```suggestion
```

tests/python/packages/isce3/signal/rfi_tone_rank.py
```diff
@@ -0,0 +1,58 @@
...
+    signal.shape = (m, n)
+
+    f = 97 / 313  # ratio of different primes
+    tone = np.exp(1j * 2 * np.pi * f * np.arange(n))
+    phases = np.random.uniform(0, 2 * np.pi, size=m)
+    interference = tone[None, :] * np.exp(1j * phases[:, None])
+
+    isr = np.sum(abs2(interference)) / np.sum(abs2(signal))
+    isr_desired = 10.0
+    z = signal + interference * isr_desired / isr
```
bhawkins:
```suggestion
    z = signal + interference * np.sqrt(isr_desired / isr)
```
Need a square root here?

tests/python/packages/isce3/signal/rfi_tone_rank.py
bhawkins: Good point. Added Monte Carlo sim to unit tests.

python/packages/isce3/signal/rfi_tone_rank.py
bhawkins: Implemented an arithmetic correction that avoids this problem.

bhawkins:
Thanks everyone for the extremely useful feedback. I think I've addressed all the main points and will re-run a full-frame test case tomorrow.

hfattahi:
> Thanks everyone for the extremely useful feedback. I think I've addressed all the main points and will re-run a full-frame test case tomorrow.



> Thanks everyone for the extremely useful feedback. I think I've addressed all the main points and will re-run a full-frame test case tomorrow.

Thanks @bhawkins for addressing team's comments. Looks good to me. Let me know how the test goes with the final run and I will approve. 
@hb6688 @rad-eng-59 @Tyler-g-hudson take a final look please. We will not have a review meeting unless anyone has a concern that needs to be discussed further.

hb6688:
@bhawkins , I think this is a very good PR which enhances our ability to detect and mitigate RFI.  It is very flexible, efficient frequency domain algorithm, and has an adaptive element in its threshold derivation which is crucial for NISAR's large swath width.  We should merge this PR and resolve the remaining issue in later releases.  I would like to just point out a few issues we discussed yesterday for future discussions.

1. The derivation of ISR should be updated and improved.  The current computation is over-counting the signal frequency bins.  In my opinion the ISR should be similar to what is shown in a spectrum analyzer with common unit, e.g. dB/Hz to derive a dBc value.   I plan to provide a simpler maximum ISR or INR ratio metadata for ST-EVD instead of total ISR or INR.  We can always discuss this in the future.

2. Metadata outputs need additional coordination since RFI likelihood for your algorithm is currently not yet finalized.

hb6688 (approved):

bhawkins:
I ran two full-frame cases this morning and both looked reasonable.

bhawkins:
> The derivation of ISR should be updated and improved.

@hb6688 ICYMI I made several small patches last night, including this one with the arithmetic correction we discussed in yesterday's meeting: https://github.com/isce-framework/isce3/pull/257/changes/118e1033b0bd9e3895198a21e6f07732cd16a35f. Not sure if that's what you meant or something else. Happy to hack on it more in a future PR to help improve and/or harmonize with the other RFI algorithms.

hfattahi (approved):
LGTM!. Thanks for this great addition @bhawkins . Nice work.

Ignore my comments which are already addressed.

python/packages/isce3/signal/rfi_tone_rank.py
```diff
@@ -0,0 +1,520 @@
...
+    # achieve a specified nominal false positive rate by thresholding the
+    # nominal CDF at one minus that value.
+    q = 1.0 - nominal_false_positive_rate
+    too_loud = exp_from_quantile(q, λ, bandwidth)
+    # Use threshold to generate mask.
+    mask = power_spectra >= too_loud
+    # Compute interference-to-signal ratio now that we've labeled the data.
+    # Since masked values are presumably the sum of signal and interference,
+    # we'll include an arithmetic correction using the stats model.
+    mean_power = 1 / λ
```
rad-eng-59:
```suggestion
    mean_power = 1 / λ**2
```

python/packages/isce3/signal/rfi_tone_rank.py
```diff
@@ -0,0 +1,520 @@
...
+    masked_sig_power = np.sum(mask) * mean_power
+    rfi_power = np.sum(power_spectra[mask]) - masked_sig_power
```
rad-eng-59, line 132:
```suggestion
    rfi_power = max(np.sum(power_spectra[mask]) - masked_sig_power, 0)
```

python/packages/isce3/signal/rfi_tone_rank.py
```diff
@@ -0,0 +1,520 @@
...
+    bandwidth=5 / 6,
+):
+    """
+    Detect RFI in spectral domain data using lifted exponential model.
+
+    Uses a statistical model to identify spectral samples that are anomalously
+    loud compared to the expected distribution of the signal. The method assumes
+    that the power spectrum follows a lifted exponential distribution and uses
+    a quantile of the data to estimate the distribution parameters.
+
```
rad-eng-59, line 74: Not sure but perhaps it is worth mentioning that the spectral density of desired signal (echo+noise) is assumed to be relatively constant over entire TX band, that is PSD is pretty flat over within passband. 
Imagine, there is ramp of a couple of dB across range bandwidth (not talking  about small ripples) due to instrument in particular for relatively wide-band mode. You could still use this method as it is by, for instance, sub-banding in range frequency and treat each range sub-band as `spectra` to this function with its own `lambda`. With latter, you can assume flattness over PSD and also avoid problem with point-like targets/CR captured partially where that can introduce unwanted bias in the median even in the  absence of RFI. Basically, your parameter distribution varies in both time and frequency. The distribution still exponential. Alternatively, a simplified Gamma distribution perhaps works for both random and urban targets in the presence of not-flat spectra. Anyway, I have not tried and experienced this by myself so I might be mistaken.

python/packages/isce3/signal/rfi_tone_rank.py
```diff
@@ -0,0 +1,520 @@
...
+            else:
+                raise ValueError(f"Invalid fill_value: {fill_value}")
+
+    # Put Doppler back on.
+    zout *= deramp[:, None].conj()
+    return zout
+
+
+def circular_gaussian_noise(n, σ=1, dtype=np.complex64):
+    return σ * (np.random.normal(size=n) + 1j * np.random.normal(size=n))
```
rad-eng-59, line 247:
```suggestion
def circular_gaussian_noise(n, σ=1, dtype=np.complex64, seed=100):
    # σ = / sqrt(2) , I'm assuming we don't need this scalar even when we get STD=sqrt(2) w/ sigma=1!
    rng = np.random.default_rng(seed=seed)
    noise = rng.normal(size=n, scale=σ) + 1j * rng.normal(size=n, scale=σ)
    return noise.astype(dtype)
```

python/packages/isce3/signal/rfi_tone_rank.py
```diff
@@ -0,0 +1,520 @@
...
+    if valid_rows.shape != (m,):
+        raise ValueError(f"{valid_rows.shape=} does not match {z.shape[0]=}")
+    if interpolate and m < 2:
+        log.warning(f"Disabling interpolation for block of {m} rows "
+            "because there are no neighbors to use for interpolation.")
+        interpolate = False
+
+    # Deramp Doppler.
+    deramp = np.exp(-1j * 2 * np.pi * fd * t).astype(z.dtype)
+    zout = deramp[:, None] * z
```
rad-eng-59, line 189: Just for readability. Feel free to ignore
```suggestion
    zout = deramp[:, np.newaxis] * z
```

python/packages/isce3/signal/rfi_tone_rank.py
```diff
@@ -0,0 +1,520 @@
...
+    z : np.ndarray
+        Complex spectral data to fill, shape (m, n)
+    fd : float
+        Doppler centroid frequency in Hz.  It is not scaled for each frequency
+        bin in `z`.  If wideband processing is needed, just baseband `z` before
+        calling this function and set `fd=0`.
+    t : np.ndarray
+        Pulse times in seconds since epoch, length m
+    mask_replace : np.ndarray
+        Boolean mask marking samples to replace, shape (m, n)
```
rad-eng-59, line 156: Just for clarity. Feel free to ignore this:
```suggestion
        Boolean mask marking range frequency bins (columns) per pulse (row) of `z` to replace, shape (m, n)
```

python/packages/isce3/signal/rfi_tone_rank.py
```diff
@@ -0,0 +1,520 @@
...
+        length num_range_blocks.
+    f : np.ndarray[float]
+        Normalized frequency axis for hits array, length block_dims[1].
+        Units are cycles per sample.
+    means : np.ndarray[float]
+        Estimated mean signal power per block from lifted exponential model,
+        shape (num_az_blocks, num_range_blocks). Equal to 1/λ where λ is the
+        rate parameter.
+    isr : np.ndarray[float]
+        Interference-to-signal ratio per block,
```
rad-eng-59, line 332:
```suggestion
        Interference-to-signal power ratio (linear) per block,
```

python/packages/isce3/signal/rfi_tone_rank.py
```diff
@@ -0,0 +1,520 @@
...
+        for i, i_pulse in enumerate(range(rows.start, rows.stop)):
+            for start, end in swaths[:, i_pulse, :]:
+                mask_valid[i, start:end] = True
+        # apply window to each range block
+        for j, (cols, window) in enumerate(slices_windows):
+            nw = len(window)
+            z_block[:, j, :nw] = window[None, :] * z[rows, cols]
+        # Range STFT.  Use consistent FFT size even for edges where window
+        # may be shorter so that frequency metadata are consistent.
+        spectra = fft(z_block, n=block_dims[1], axis=2)
```
rad-eng-59, line 416: I think it uses single core by default. Also, not sure if it is actually efficient for small/short `1K` FFT.
```suggestion
        spectra = fft(z_block, n=block_dims[1], axis=2, workers=-1)
```

python/packages/isce3/signal/rfi_tone_rank.py
```diff
@@ -0,0 +1,520 @@
...
+                valid_rows = ((~mask_valid[:, cols]).mean(axis=1)
+                    <= max_gap_fraction)
+                spectra[:, j, :] = fill_missing(spectra[:,j,:], fd, pulse_times,
+                    mask_replace, valid_rows, noise, interpolate, fill_value)
+            hits[iblock, j, :] = fftshift(np.mean(mask_replace, axis=0))
+            means[iblock, j] = 1 / λ if λ > 0.0 else 0.0
+        # skip inverse FFTs and assignment if not required.
+        if not detect_only:
+            # range inverse STFT
+            z_block[...] = ifft(spectra, axis=2)
```
rad-eng-59, line 439:
```suggestion
            z_block[...] = ifft(spectra, axis=2, workers=-1)
```

python/packages/isce3/signal/rfi_tone_rank.py
rad-eng-59, line 416: Not following the comment. The `z_block` has a fixed shape and its last dimension always has `block_dims[1]` elements. So, FFT by default uses the same size as that dimension. will the following break?

```suggestion
        spectra = fft(z_block, axis=2 , workers=-1)
```

python/packages/isce3/signal/rfi_tone_rank.py
```diff
@@ -0,0 +1,520 @@
...
+                mask_valid[i, start:end] = True
+        # apply window to each range block
+        for j, (cols, window) in enumerate(slices_windows):
+            nw = len(window)
+            z_block[:, j, :nw] = window[None, :] * z[rows, cols]
+        # Range STFT.  Use consistent FFT size even for edges where window
+        # may be shorter so that frequency metadata are consistent.
+        spectra = fft(z_block, n=block_dims[1], axis=2)
+        # filter RFI
+        for j in range(z_block.shape[1]):
```
rad-eng-59, line 418: Perhaps easier to follow:
```suggestion
        for j in range(num_range_blocks):
```

python/packages/isce3/signal/rfi_tone_rank.py
```diff
@@ -0,0 +1,520 @@
...
+    for j, (cols, _) in enumerate(slices_windows):
+        # TODO could weight by window
+        block_ranges[j] = r[(cols.start + cols.stop) // 2]
+
+    # Generate a pool of noise with σ=1 once, and we'll scale it for each
+    # block as needed.  We'll generate a random offset for each block to
+    # increase entropy without having to regenerate noise for each block.
+    max_offset = num_az_blocks * num_range_blocks
+    num_noise = max_offset + np.prod(block_dims)
+    std_noise = circular_gaussian_noise(num_noise)
```
rad-eng-59, line 398: Not sure how one should define power for the a complex with circular Gaussian distribution. 
If I use default, the STD of the output will be `sqrt(2)` not `1.0` w/o the built-in scalar (scale/sqrt(2)). 
I guess what matters is consistency of scalar from Exponent to Rayleigh to Gaussian distributions.

rad-eng-59 (commented):
Given I'm too slow to follow the details and avoid wasting everyone's time,  I did simply checked out  the `rfi_tone_rank` module. Feel free to ignore the trivial comments.
LGTM. Thank you for the PR.

PR #262, opened 2026-05-01
Modify QP Logic in Parsing CalType and ChirpCorrelator

rad-eng-59:
This quick PR slightly modifies the logic used in extracting chirp calibration path types and chirp correlators from L0B high-rate telemetry for simply linear Quad-Pol (QP). 
This fixes the issue with radiometric calibration (*EAP=ON*) and noise estimation (*NEB*) in RSLC L1 product for QP case where the first TX pol is **H**. 
Besides, it fixes the issue with missing sniffer pulses in noise estimator workflow from raw L0B products for QP.
This PR has been successfully tested by generating RSLC w/ EAP=ON and estimating noise power from L0B over two different Fixed-PRF QP modes supporting both H and V as the first TX polarization.

Here is the example of 3-second QP RSLC browse images toggling between EAP=OFF/ON:

- *A & HH/HV/VH/VV*
EAP_OFF2ON

- *B & HH*  
EAP_OFF2ON_B_HH

### DIFF
--- a/python/packages/nisar/noise/noise_estimation_from_raw.py
+++ b/python/packages/nisar/noise/noise_estimation_from_raw.py
@@ -19,8 +19,7 @@ from isce3.core import DateTime, TimeDelta
 from nisar.products.readers.Raw import (
     chirpcorrelator_caltype_from_raw,
     is_raw_quad_pol,
-    first_tx_pol_for_quad,
-    opposite_linear_pol
+    first_tx_pol_for_quad
 )
 
 # Global Noise-related Constants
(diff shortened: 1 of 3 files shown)


rad-eng-59:
The `if-else` condition for handling `tx=H/V` in `Raw` is actually appropriate as it is. So no code change.
However, I have added more comments and description to clarify the QP logic and to state hidden assumptions in regards to sniffer pulses for L-band NISAR products. Hope it makes more sense now.

bhawkins (approved):
LGTM.

I see now that the in the TX-H case that it pulls out the TX-V cal data before decimating by two, so I agree we need that branch in Raw.py.

hfattahi (approved):
LGTM. Thanks @rad-eng-59 for the fix. Not easy to track that down! Please wait for [this PR](https://github.com/isce-framework/isce3/pull/264) to finish which fixes CI, make sure to pull the changes from the CI fixing PR which allows the CI failures here to resolve and then merge.

PR #263, opened 2026-05-01
Fix RSLC mask in mixed mode cases

bhawkins:
This PR addresses the problem observed in mixed-mode cases where a small sliver is marked invalid at the boundaries between observations. Here's an illustration of the valid (start, end) indices for the following granule

NISAR_L1_PR_RSLC_009_005_A_019_4005_DHDH_A_20251228T125147_20251228T125222_X05013_N_F_J_001

fig_bad_mask

What happens is there's a small time interval between the last pulse of one observation and the first pulse of the next observation. I had considered this case and implemented the logic [here](https://github.com/isce-framework/isce3/blob/bfe0c1a99acc498ad661a46956b749460b5e7c26/python/packages/isce3/focus/valid_regions.py#L407) to deal with it. However, I guess I hadn't put a lot of thought into the tolerance for merging the polygons. The algorithm is looping over scanlines intersecting the polygons, so it uses a range threshold. The basic relationship relating a range gap Δr to an azimuth gap Δs is $\Delta r = \Delta s / \tan\theta_{sq}$. The necessary merge threshold to cover a one-pulse or so gap works out to up to 250 range samples depending on your degree of conservatism (in the granule above the range gap was 75 samples). That's longer than our shortest pulse, so it's not a reasonable way to solve the problem.

Instead, I added a step after getting the raw data bounding boxes that forces them to share an edge when they're separated by less than some user-provided threshold. I set the default to 2 ms. The example above had a gap of 1.116 ms, which is honestly a little bigger than I expected.

I also changed the `Raw.getSubSwathBboxes` method to report an upper time bound of one PRI past the last pulse, which is consistent with what it does in range and with the interior bboxes due to DWP updates. However, those changes are not really necessary given the fix in higher-level logic.

With these changes there's no longer any pinching of the valid data mask around the 2-second mid-frame observation.

### DIFF
--- a/python/packages/nisar/workflows/focus.py
+++ b/python/packages/nisar/workflows/focus.py
@@ -1619,8 +1619,9 @@ def get_focused_sub_swaths(rawlist, out_chan, grid, orbit, doppler, dem, azres,
         T = raw.getChirpParameters(freq, txpol)[3]
         chirp_durations.extend(len(bbox_lists) * [T])
 
-    # Force azimuth continuity since Raw.getSubSwathBboxes only guesses about
-    # last PRI.
+    # Force azimuth continuity since Raw.getSubSwathBboxes doesn't know final
+    # PRI so there's a 1-pulse gap between observations.  Note that there
+    # should be no gap between 10-second DWP updates.
     for i in range(len(raw_bbox_lists) - 1):
         # Each subswath should have the same start/end time, just different
         # ranges.
@@ -1630,7 +1631,7 @@ def get_focused_sub_swaths(rawlist, out_chan, grid, orbit, doppler, dem, azres,
         if dt <= max_observation_gap:
             log.info(f"Merging observations separated by {dt * 1e6:.2f} us "
                 f"at {orbit.reference_epoch + TimeDelta(t_cur)}")
-            if dt <= 0.0:
+            if dt < 0.0:
                 # The time difference should always be positive since there's at
                 # least one PRI between the end of one observation and the start
                 # of the next one.  However, as of 2026-05-04, L0B time stamps
--- a/python/packages/nisar/workflows/focus.py
+++ b/python/packages/nisar/workflows/focus.py
@@ -1645,7 +1645,7 @@ def get_focused_sub_swaths(rawlist, out_chan, grid, orbit, doppler, dem, azres,
                 log.warning("Time decremented between observations.  "
                     "Assuming seamless transition.")
             for bbox in raw_bbox_lists[i]:
-                bbox.last.time = t_next
+                bbox.last.time = max(t_next, t_cur)
         else:
             log.warning(f"Gap between observations {dt:7f} s exceeds threshold "
                 f"for seamless observations ({max_observation_gap} s).")
--- a/python/packages/nisar/workflows/focus.py
+++ b/python/packages/nisar/workflows/focus.py
@@ -1546,7 +1546,7 @@ def get_output_range_spacings(rawlist: list[Raw], common_mode: PolChannelSet):
 def get_focused_sub_swaths(rawlist, out_chan, grid, orbit, doppler, dem, azres,
                            rdr2geo_params=dict(), geo2rdr_params=dict(),
                            ignore_failure=False, polygon_segment_length=50.0,
-                           num_ignore=25):
+                           num_ignore=25, max_observation_gap=800e-6):
     """
     Determine fully-focused regions of the image in a format suitable for
     populating the validSamplesSubSwathX RSLC datasets.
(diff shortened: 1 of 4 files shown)


python/packages/nisar/workflows/focus.py
```diff
@@ -1611,6 +1616,25 @@ def get_focused_sub_swaths(rawlist, out_chan, grid, orbit, doppler, dem, azres,
...
+    # last PRI.
+    for i in range(len(raw_bbox_lists) - 1):
+        # Each subswath should have the same start/end time, just different
+        # ranges.
+        t_cur = raw_bbox_lists[i][0].last.time
+        t_next = raw_bbox_lists[i + 1][0].first.time
+        # abs() since difference could be negative if Raw.getSubSwathBboxes
+        # guess for the final PRI is larger than the actual final PRI.
+        dt = abs(t_next - t_cur)
+        if 0 < dt <= max_observation_gap:
```
bhawkins: Maybe it's cleaner not to take `abs()` here and treat any negative `dt` as continuous. That way it doesn't matter how bad the guess is for the final PRI, at least in one direction.

Either that or remove the kludge to `getSubSwathBboxes` since that's the only way to get negative `dt`. If we do that then we have to change the condition on the big gap log message to `else`.

share/nisar/schemas/focus.yaml
```diff
@@ -309,6 +309,13 @@ mask_options:
   # one.
   num_ignore: num(min=0, required=False)
 
+  # Max allowed time (in seconds) between the last pulse of one
+  # observation and the first pulse of the following observation
+  # for the two to be considered seamless.  Larger raw data gaps
+  # may result in a synthetic aperture being marked invalid in the
+  # RSLC.
+  max_observation_gap: num(min=0, required=False)
```
bhawkins, line 317:
```suggestion
  max_observation_gap: num(exclusiveMinimum=0, required=False)
```
We should check if Yamale allows this.

bhawkins (commented):
Comments from today's online review.

Not in this PR, but have a look at `rasterize_subswath_polygons` to make sure there can't be a wraparound due to subtraction of uint32 values.

share/nisar/schemas/focus.yaml
bhawkins: I don't see an easy way to do this. The docs for the number validator ([here](https://github.com/23andMe/Yamale#number---numminfloat-maxfloat)) only list `min` and `max`, and I don't see any other keywords in the Yamale source code, either. It does allow defining custom validators, but that's a fair amount of work and the code would have to be copied into PCM somehow since they use the schemas, too.

bhawkins:
> Not in this PR, but have a look at rasterize_subswath_polygons to make sure there can't be a wraparound due to subtraction of uint32 values.

I had a look at the code and verified the values are sorted a few lines above ([here](https://github.com/isce-framework/isce3/blob/bfe0c1a99acc498ad661a46956b749460b5e7c26/python/packages/isce3/focus/valid_regions.py#L401)) so it's guaranteed the subtraction is safe.

bhawkins:
Hacking on this I realized I had a sign error for `final_pri` in the Raw bbox method, so the gap between files is actually half what I'd reported before, which is in the expected range of NISAR PRIs:
```
INFO:focus:Merging observations separated by 558.00 us at 2025-12-28T12:51:59.102761296
INFO:focus:Merging observations separated by 540.00 us at 2025-12-28T12:52:01.245760564
```
Anyway, following Hirad's suggestion I just got rid of that code altogether.

I also realized there is a circumstance where one file could seem to start before the previous one ends. That's because the L0B time stamps are derived from the LRCLK counter using the SCLK-SCET file generated for each downlink pass. So it's possible two observations could be downlinked on separate passes and use different SCLK-SCET conversions that are different enough to make time go backwards. Given our system requirements that should never happen, but I put in a special log message in case it does.

python/packages/nisar/workflows/focus.py
```diff
@@ -1611,6 +1619,37 @@ def get_focused_sub_swaths(rawlist, out_chan, grid, orbit, doppler, dem, azres,
...
+                # are derived from LRCLK counts using a model that's updated
+                # every downlink pass.  If the observations were downlinked on
+                # separate passes, it's conceivable that time could go backwards
+                # (though this would violate requirements).  If that happens it
+                # seems safe to assume that's a seamless transition, so just log
+                # it and proceed.
+                log.warning("Time decremented between observations.  "
+                    "Assuming seamless transition.")
+            for bbox in raw_bbox_lists[i]:
+                bbox.last.time = t_next
```
rad-eng-59: Just to be on the safe side and cover both positive and negative case, perhaps the following?
```suggestion
                bbox.last.time = max(t_next, t_cur)
```

rad-eng-59 (approved):
LGTM. Thank you for the update.

hfattahi (approved):
Thanks for the fix and the updates after PR review. LGTM.

PR #268, opened 2026-05-20
Fix recently-introduced mixed-mode bug

bhawkins:
I should know better by now, but when adding the tone-rank RFI suppression feature I messed up the bookkeeping of the subband frequency ID. This causes problems in mixed-mode processing, particularly when mixing 77 MHz with any dual-frequency mode. In that case we downsample the 77 MHz data twice, so there's one iteration where the output RSLC data will be labeled "B" but we're using L0B data from "A".

This PR changes the code to use the correct label for querying the source L0B.

### DIFF
--- a/python/packages/nisar/workflows/focus.py
+++ b/python/packages/nisar/workflows/focus.py
@@ -2092,7 +2092,7 @@ def focus(runconfig, runconfig_path=""):
                 # Only write rich HDF5 for tone-rank
                 (rfi_results_h5.require_group(f"raw{raw_times[0]:05.0f}")
                     if using_tone_rank else None),
-                raw.getCenterFrequency(frequency),
+                raw.getCenterFrequency(channel_in.freq_id),
                 fs,
             )
             rfi_results[(frequency, pol)].append(


hfattahi (approved):
LGTM! Thanks @bhawkins .

python/packages/nisar/workflows/focus.py
rad-eng-59, line 2090: Should not one rescale the Doppler values per center frequency of input raw data `channel_in.freq_id` where RFI is mitigated prior to filtering to a common frequency band `frequency`? perhaps the Doppler's sensitivity to relative bandwidth less than `5%` is negligible in entire workflow.

rad-eng-59 (approved):
LGTM. Thanks.

python/packages/nisar/workflows/focus.py
bhawkins: Yes, I had the same thought about the Doppler centroid (DC) here. The presummer has the same issue. By the time we get to azcomp the data have been filtered, so the usage there is obviously correct.

Thinking about these azimuth interpolation steps a little more, since we're going to filter the range spectrum in a later step, maybe it's actually better to use the corresponding DC for centering the azimuth spectrum in earlier steps. That is, if we had put the range-filtering first, we'd obviously want to use the DC for that band. But they're linear operations that should commute, so the order shouldn't actually matter.

Following that train of thought, why are we using the unmodified center frequency here? Alas, it's only being used to label the frequency axis of the range spectra, so that's the correct behavior.

Like you say, it's a small effect anyhow.

PR #269, opened 2026-05-20
Scale RSLC mixed-mode filter by bandwidth ratio

bhawkins:
The range reference function is normalized by its white noise gain, which helps keep the calibration constant when we change the window parameter. However, in the mixed-mode case that normalization also increases the pass-band gain in proportion to the bandwidth reduction. The latter effect is not desirable since it causes a radiometric offset.

To test this change, I found a case near the south pole where we mix 77 MHz (half-swath) with 40+5 MHz (full-swath) data. Looking at the 77 & 5 MHz mix is most obvious, since the change is almost 12 dB in that case. With the change, we get a nice continuous backscatter image.

mixed_mode_normalization

Note 1: I just used the browse QA images for the above figure, so they have independent color maps. You can see from the histogram and profile that it's actually the 77 MHz getting darker to match the 5 MHz part.

Note 2: This test case also requires #268

### DIFF
--- a/python/packages/nisar/workflows/focus.py
+++ b/python/packages/nisar/workflows/focus.py
@@ -1347,6 +1347,10 @@ def prep_rangecomp(cfg, raw, raw_grid, channel_in, channel_out, cal=None,
     log.info("Normalizing chirp to unit white noise gain.")
     chirp *= 1.0 / np.linalg.norm(chirp)
 
+    if channel_in.band != channel_out.band:
+        log.info("Re-scaling by mixed-mode filter bandwidth ratio")
+        chirp *= np.sqrt(channel_out.band.width / channel_in.band.width)
+
     # Careful to use effective TBP after mixed-mode filtering.
     time_bw_product = channel_out.band.width**2 / abs(K)
 


hfattahi (approved):
LGTM! Thanks @bhawkins .

gshiroma (commented):
LGTM! Thank you for fixing this, @bhawkins !

gshiroma (approved):

python/packages/nisar/workflows/focus.py
```diff
@@ -1347,6 +1347,10 @@ def prep_rangecomp(cfg, raw, raw_grid, channel_in, channel_out, cal=None,
     log.info("Normalizing chirp to unit white noise gain.")
     chirp *= 1.0 / np.linalg.norm(chirp)
 
+    if channel_in.band != channel_out.band:
+        log.info("Re-scaling by mixed-mode filter bandwidth ratio")
+        chirp *= np.sqrt(channel_out.band.width / channel_in.band.width)
```
rad-eng-59, line 1352: This will also scale noise power, `NEB`, that already being estimated within output bandwidth after range comp filtering. I guess that should be okay given SNR has not changed by this scaling. I wonder what we used to report for NEB in mixed mode case.

rad-eng-59 (commented):
It seems like my comments/reviews are already too late again.

PR #274, opened 2026-05-24
Fix bug with repeated calls to form_pattern

bhawkins:
I was trying to optimize the antenna pattern weights and thought I was going crazy because I'd get different patterns every time I called `AntennaPattern.form_pattern`. It turns out #231 introduced a bug where the receive pattern weights mutate the stored LNA/caltone ratio on every call.

In focus.py we only call `form_pattern` twice on each instance, once for the science data and then again for the noise data. So in production the image is fine but the noise equivalent backscatter is calculating the wrong RX pattern whenever the weights aren't unity.

### DIFF
--- a/python/packages/nisar/antenna/pattern.py
+++ b/python/packages/nisar/antenna/pattern.py
@@ -399,7 +399,7 @@ class AntennaPattern:
             channel_adj_fact_rx = (
                 self.rx_imb[self.freq_band, txrx_pol].lna_caltone_ratio)
             if self.channel_adj_fact_rx[rxp] is not None:
-                channel_adj_fact_rx *= np.asarray(
+                channel_adj_fact_rx = channel_adj_fact_rx * np.asarray(
                     self.channel_adj_fact_rx[rxp])
 
             # Split up provided timespan into groups with the same range timing


python/packages/nisar/antenna/pattern.py
```diff
@@ -399,7 +399,7 @@ def form_pattern(self, tseq, slant_range: Linspace,
             channel_adj_fact_rx = (
                 self.rx_imb[self.freq_band, txrx_pol].lna_caltone_ratio)
             if self.channel_adj_fact_rx[rxp] is not None:
-                channel_adj_fact_rx *= np.asarray(
+                channel_adj_fact_rx = channel_adj_fact_rx * np.asarray(
                     self.channel_adj_fact_rx[rxp])
```
rad-eng-59, line 403: I don't understand how the variable `channel_adj_fact_rx` is being mutated here when it is being initialized prior to being updated by external factor in every call.  So, no matter how many times we call this function, it will be reset to its initial value coming from `LNA/CALTONE` pre-computed by the respective ctor. 
However, I have been under the impression that there is an implicit copy of the array `self.rx_imb[self.freq_band, txrx_pol].lna_caltone_ratio` assigned to `channel_adj_fact_rx`, otherwise it would rewrite the pre-computed `LNA/CALTONE` and we need to explicitly make a copy of that array here prior to updating with external factors.
Apparently, I was wrong and we need to make an explicit copy of the array. That is

```python
channel_adj_fact_rx = self.rx_imb[self.freq_band, txrx_pol].lna_caltone_ratio.copy()
if self.channel_adj_fact_rx[rxp] is not None:
                channel_adj_fact_rx *= np.asarray(
                    self.channel_adj_fact_rx[rxp])
```
This way, the `LNA/CALTONE` ratio is not overwritten. Again there is no mutation on the line you have modified.

rad-eng-59: Just tested a QP case by simply modifying the line `400` with **self.rx_imb[self.freq_band, txrx_pol].lna_caltone_ratio.copy())** and the respective NEB values look reasonable (except  around TX gap due to the TX chirp leakage in particular for band=B).
My bad for naively assuming the array of a dataclass being implicitly copied.

rad-eng-59 (approved):
LGTM. Thank you for the fix.
I thought we could fix the problem earlier in case somebody add another line in the future and make the same mistake I did. 
Btw, that makes me realize that elements of a numpy-array type member  of so-called immutable frozen dataclass is still mutable! Very annoying.

hfattahi (approved):
LGTM! Thanks @bhawkins

PR #275, opened 2026-05-26
Improve DEM bounding box selection when product coordinates are in a different projection than the DEM

gshiroma:
This PR improves the DEM bounding box selection when the product bounding box projection differs from the DEM projection, the previous implementation reprojected only the four corners of the bounding box to determine the DEM extent. 

This is inaccurate when reprojecting between coordinate systems where straight edges become curves (e.g. UTM to geographic), causing the computed bounding box to miss the bulge introduced by geodetic curvature. The result was a DEM subset that was too small, leading to interpolation failures or incorrect elevation values near the edges.

Instead of projecting the four corners only, all four edges of the input bounding box are now densely sampled (`n_edge_samples` points per edge, default `11`) before reprojection, ensuring the true geographic extent is captured regardless of projection curvature.

The antimeridian handling is simplified. Rather than treating the western and eastern edges separately, a single check over all sampled points is performed. If the DEM is in geographic coordinates and the reprojected X range exceeds 180 degrees, longitudes are remapped to [0, 360] before computing the extent.


cxx/isce3/geometry/loadDem.cpp
```diff
@@ -132,130 +132,92 @@ DEMInterpolator DEMRasterToInterpolator(
...
-
-            else {
+        const int N = std::max(n_edge_samples, 2);
+
+        // Densely sample all four edges of the input bounding box
+        // to capture curvature introduced by reprojection
+        // (e.g. UTM -> geographic).
+        std::vector<double> all_x, all_y;
+        all_x.reserve(4 * N);
+        all_y.reserve(4 * N);
```
gshiroma: Check if this can be reduced to one line

cxx/isce3/geometry/loadDem.cpp
```diff
@@ -132,130 +132,92 @@ DEMInterpolator DEMRasterToInterpolator(
...
+        min_y = *std::min_element(all_y.begin(), all_y.end());
+        max_y = *std::max_element(all_y.begin(), all_y.end());
+
+        min_x = *std::min_element(all_x.begin(), all_x.end());
+        max_x = *std::max_element(all_x.begin(), all_x.end());
+
+        // If the DEM is in geographic coordinates and the X range
+        // exceeds 180 degrees, an antimeridian crossing is likely.
+        // Retry in [0, 360] domain.
+        if (dem_raster.getEPSG() == 4326 && max_x - min_x > 180.0) {
```
gshiroma: ``
```suggestion
        if (dem_raster.getEPSG() == 4326 && (max_x - min_x) > 180.0) {
```

bhawkins (approved):
LGTM

gshiroma (commented):
Comments from PR review meeting (cc'ed @bhawkins and @seongsujeong )

cxx/isce3/geometry/loadDem.cpp
```diff
@@ -264,11 +224,15 @@ isce3::error::ErrorCode loadDemFromProj(
     min_y -= margin_y;
     max_y += margin_y;
 
-
-    float margin_x = dem_margin_x_in_pixels * dem_raster.dx();
+    float margin_x = dem_margin_x_in_pixels * std::abs(dem_raster.dx());
     min_x -= margin_x;
     max_x += margin_x;
 
+    if (dem_raster.getEPSG() == 4326) {
```
seongsujeong, line 235:
```suggestion
    # make sure that the latitude falls between [-90.0, 90.0] after applying `margin_y`
    if (dem_raster.getEPSG() == 4326) {
```
Adding a comment for clarification

cxx/isce3/geometry/loadDem.cpp
```diff
@@ -264,11 +224,15 @@ isce3::error::ErrorCode loadDemFromProj(
...
+        min_y = std::clamp(min_y, -90.0, 90.0);
```
seongsujeong, line 236: A nit pick: it seems `margin_y` is declared as float but min/max y is as double. It should not cause a big issue, but I would suggest to keep the consistency in the data type.

seongsujeong (approved):
LGTM. Leaving a couple of minor suggestion and comments.

PR #280, opened 2026-05-28
Update static layers product specs XML

gshiroma:
This PR updates the static layers product specs XML


nemo794:
Thanks @gshiroma . Hmm, based on this PR description: https://github-fn.jpl.nasa.gov/NISAR-ADT/NISAR_PIX/pull/343 , it looks like ISCE3's version of the XML is correct, but that PIX's version of the XML is incorrect.

If so, then we should close this PR in ISCE3, and instead merge that PR in PIX.

When you open the sample STATIC HDF5 outputs, can you confirm repo is matches the HDF5?

gshiroma:
No longer needed

PR #76, opened 2025-07-24
Pycuampcor V2

rtburns-jpl:
Merging upstream changes from Lijun for PyCuAmpcor, and changes from @vbrancat to fix the runconfigs and parsers to support his newly-added parameters

See corresponding PR in isce2 (https://github.com/isce-framework/isce2/pull/922):

> PyCuAmpcor Version 2 introduces a new workflow inspired by [GrIMP/SpeckleSource](https://github.com/fastice/speckleSource), in addition to the original ROIPAC/ampcor workflow. Below is an overview of the new workflow and additional improvements:
> 
> Key Features of the New GrIMP/SpeckleSource Workflow:
> 
>     * Starts with an anti-aliasing oversampled secondary chip over the entire search range.
> 
>     * Adds extra margins to facilitate correlation surface extraction.
> 
>     * Offers optional double-precision computations (also extended to the ROIPAC/ampcor workflow for consistency).
> 
> 
> The GrIMP/SpeckleSource workflow may be slower than ROIPAC/ampcor due to increased computational demands. However, it may improve accuracy in scenarios with low signal-to-noise ratio (SNR). More details on the workflows and how to choose can be found at contrib/PyCuAmpcor/README.md.
> 
> Other Improvements
> 
>     * Introduced an option to fix the starting pixel location, which is ideal for comparing results from different window sizes.
> 
>     * Added an example script for plotting the offset field, making it easier for users to visualize results.
> 
>     * Included correlation peak values in the output, alongside SNR and covariance, providing additional accuracy measures.
> 
>     * Resolved a memory deallocation issue that occurred when the run procedure was called repeatedly within a single script.

PR #282, opened 2026-06-01
Fix the Ampcor issues from Lijun's branch and increase the limit of the search radius 

xhuang-jpl:
This PR is to fix the Ampcor issue 

1.  **secondary window end pixel out of range in the across direction, at window (0, 686), pixel 27632** when the 20+5 data over Chile is processed. The runconfig of the dense_offset is 

```
window_range: 64
window_azimuth: 128
half_search_range: 32
half_search_azimuth: 64
skip_range: 40
skip_azimuth: 75
correlation_surface_oversampling_factor: 16
deramping_method: complex
```
2. **The unexpected results for different search windows.**


xhuang-jpl:
@lijun99 Thanks again for your PR, would you mind taking a look the merge? It seems that I cannot add you to be the reviewer.

lijun99:
I tested the merged version and it works fine. Thanks @xhuang-jpl !

xhuang-jpl:
> I tested the merged version and it works fine. Thanks @xhuang-jpl !

Thanks @lijun99 , I also tested, and works fine on both CPU and GPU.

hfattahi (approved):
The bug fixes look reasonable to me. thanks @lijun99 and @xhuang-jpl .

hfattahi:
@xhuang-jpl please update the PR description to clarify that you have also relaxed the schema and allow larger half search window.

cxx/isce3/cuda/matchtemplate/pycuampcor/cuAmpcorParameter.cu
```diff
@@ -256,60 +270,37 @@ void cuAmpcorParameter::setChunkStartPixels()
...
-            }
-            if(referenceStartPixelAcross[i] <0)
-            {
-                fprintf(stderr, "Reference Window start pixel out of range in Across, window (%d,%d), pixel %d\n", row, col, referenceStartPixelAcross[i]);
-                exit(EXIT_FAILURE);
-            }
+            if(referenceStartPixelDown[i] < 0)
+                fprintf(stderr, "Warning: Reference Window start pixel out of range in Down, window (%d,%d), pixel %d\n", row, col, referenceStartPixelDown[i]);
+            if(referenceStartPixelAcross[i] < 0)
+                fprintf(stderr, "Warning: Reference Window start pixel out of range in Across, window (%d,%d), pixel %d\n", row, col, referenceStartPixelAcross[i]);
```
seongsujeong, line 283: It looks like the previous behavior was to raise an exception when this check failed. The PR changes that behavior to emit a warning message instead. Could you explain the rationale behind this change?

xeno0: Good point @seongsujeong , it is because they are not the issue anymore, and just warn the user that the window might out of range. @lijun99  has done the fix via filling 0. Please check his reply:

`The isce3 version uses a strict range check, which could fail when we choose a specific starting pixel and some pixels from search image may fall out of range (at the edges). I have patched the code from the standalone version, which fills (0,0) when using pixels out of range. Now checkPixelInRange is optional.`

lijun99: [In Image loading/copying](https://github.com/isce-framework/isce3/pull/282/changes/0c9d6a6034a81b74c9284ed337ba899a7a8413eb#diff-5b59ace56593f685b5f9742c260a5eca71d21c766f5773e84cbaed12b3fcfc4a), zeros are now filled for out of range pixels (happening at edges if margins are not set properly or if we want to use a fixed start pixel location and vary chip sizes or search ranges).  Of course, if it happens, the results along the edge are not reliable and therefore warnings may be issued. But it should not prevent the program from running. 

Before the revision, accessing out of range pixels leads to memory access errors (seg fault, at least for CUDA), and therefore checking the range and raising an exception was the only plausible solution. 

Note also that the above bug fixes have been included in the [PyCuAmpcor V2 PR](https://github.com/isce-framework/isce3/pull/76). Somehow, I forgot to patch V1. So, current version of ISCE2 has similar issues.

cxx/isce3/cuda/matchtemplate/pycuampcor/cuArraysCopy.cu
```diff
@@ -114,8 +118,12 @@ __global__ void cuArraysCopyToBatchAbsWithOffset_kernel(const float2 *imageIn, c
     int outy = threadIdx.y + blockDim.y*blockIdx.y;
     if(idxImage>=nImages || outx >= outNX || outy >= outNY) return;
     int idxOut = idxImage*outNX*outNY + outx*outNY + outy;
-    int idxIn = (offsetX[idxImage]+outx)*inNY + offsetY[idxImage] + outy;
-    imageOut[idxOut] = make_float2(complexAbs(imageIn[idxIn]), 0.0);
+    int inx = offsetX[idxImage] + outx;
```
lijun99, line 121: Here is an example that range is checked during copying and zeros are filled when out of range happens.

seongsujeong (approved):
After the discussion with @xhuang-jpl offline, I am going to approve the PR because it is well tested with various test dataset.

PR #286, opened 2026-06-08
Fix BCAL normalization of first block

bhawkins:
In the elevation antenna pattern code, the transmit path is characterized by the ratio of high-power amplifier calibration loop (HCAL) and bypass calibration loop (BCAL) measurements (normalized by the BCAL of the first channel). The HCAL and BCAL measurements are interleaved in time, so some logic is needed to associate the correct pairs of HCAL and BCAL measurements.

The existing code assumes that the first pulse is a BCAL measurement, which is typically true for the first observation in a data take. However, subsequent observations may not start with a BCAL. In those cases, a block of HCAL weights does not get properly normalized and the resulting antenna pattern is wrong. This PR handles the case by using the first BCAL to normalized the first set of HCAL pulses, even if that BCAL happens later in time.

### DIFF
--- a/python/packages/nisar/antenna/beamformer.py
+++ b/python/packages/nisar/antenna/beamformer.py
@@ -815,6 +815,11 @@ def compute_transmit_pattern_weights(tx_trm_info, norm=False):
         if np.isclose(abs(tx_weights[bcal_lines_idx]).min(), 0):
             raise RuntimeError('Zero-value BCAL data is encountered!')
 
+        # handle case where first line isn't BCAL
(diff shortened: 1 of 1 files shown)


rad-eng-59 (approved):
LGTM, Thanks.
FYI, the first `200 ms` (`~ 500` pulses) of the first observation of a datatake have bad HCAL values which causes similar issue or even worse if not excluded from processing.

Tyler-g-hudson (approved):
LGTM, thanks!

hfattahi (approved):

PR #288, opened 2026-06-09
Support multiple preprocessing mask types for unwrap

oberonia78:
This PR updates `mask_type` to support multiple mask options for unwrap preprocessing. The parameter now accepts a list of selected mask types, such as:

mask_type: ['water', 'coherence']

This allows users to combine multiple filter masks, including water, median_filter, coherence, and subswath_mask, instead of selecting only a single mask type.


xhuang-jpl (approved):
LGTM. Thank you @oberonia78

hfattahi (approved):

PR #292, opened 2026-06-15
(InSAR) Replace en dash (U+2013) with ASCII hyphen in HDF5 descriptions

nemo794:
Replaced Unicode en dash character (U+2013 "–") with standard ASCII hyphen (U+002d "-") in 19 locations across 4 Python files. The en dash was causing display issues in some programs, appearing as "â€"" instead of "–".

This issue is not readily visible to the human eye in all programs; for example, the en dash renders nicely in HDFView.

However, the issue does appear in e.g. H5Web extension for VSCode:

Screenshot 2026-06-15 at 2 15 09 PM

This fix ensures HDF5 attribute descriptions display correctly across all programs and avoids confusion with special Unicode characters in source code.

**NOTE: If this PR is approved, this change will need to be reflected in the PIX Product Specs as well. Currently, the product specs also use a Unicode en dash.**

cc: @hfattahi

### DIFF
--- a/python/packages/isce3/signal/rfi_freq_null.py
+++ b/python/packages/isce3/signal/rfi_freq_null.py
@@ -140,7 +140,7 @@ def run_freq_notch(
     __________
     ..[1]  F. Meyer, J. Nicoll, and A. Doulgeris, “Correction and Characterization 
     of Radio Frequency Interference Signatures in L-Band Synthetic Aperture Radar 
-    Data”, IEEE Trans. Geosci. Remote Sens., vol. 51, no. 10, pp. 4961–4972, Oct. 2013.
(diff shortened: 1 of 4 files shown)


xhuang-jpl (approved):
LGTM, thank you @nemo794 for the checking, I will open a PR to fix the InSAR part.

oberonia78 (approved):
LGTM. Thanks @nemo794

nemo794:
Thank you @xhuang-jpl and @oberonia78 !

nemo794:
Note: For documentation purposes, this change has been reflected in the corresponding XML specs, here: https://github-fn.jpl.nasa.gov/NISAR-ADT/NISAR_PIX/pull/358

PR #298, opened 2026-06-17
Add margins to the STATIC product Doppler LUT grid boundaries to ensure full radargrid coverage.

gshiroma:
This PR adds margins to the Doppler LUT grid boundaries to ensure full radar grid coverage. The azimuth and range margins can be configured in the runconfig, with default values set to 20% of the radar grid extent in the corresponding direction.


python/packages/nisar/static/doppler.py
```diff
@@ -42,12 +44,14 @@ def make_native_doppler_lut(
         Digital elevation model specifying the height of topography, in meters above
         some reference ellipsoid, covering a region that spans the footprint of
         `radar_grid` on the ground.
-    az_spacing : float
-        The azimuth time spacing, in seconds, of the output LUT coordinate grid.
-        Must be > 0.
-    rg_spacing : float
-        The slant range spacing, in meters, of the output LUT coordinate grid.
-        Must be > 0.
```
Tyler-g-hudson, line 50: It looks like these spacing values are still in the function parameter list, so they should still be documented here.

python/packages/nisar/static/doppler.py
```diff
@@ -78,6 +82,19 @@ def make_native_doppler_lut(
...
+    if rg_margin is not None and rg_margin < 0.0:
+        raise ValueError(f"{rg_margin=}, must be >= 0 or None")
+
+    # Expand radar grid by 20% in each direction to compute the
+    # native Doppler LUT
+    if az_margin is None:
+        az_margin = int(0.2 * radar_grid.length / radar_grid.prf)
+    if rg_margin is None:
+        rg_margin = int(0.2 * radar_grid.width *
+                        radar_grid.range_pixel_spacing)
```
Tyler-g-hudson, line 102: This default feels like it could cause undesired behavior in automated systems, e.g. if a grid ends up being especially small or large for some reason. I think we should just select a numerical default that fits the typical case and use that.

python/packages/nisar/static/doppler.py
gshiroma: Good catch, @Tyler-g-hudson , this was fixed in https://github.com/isce-framework/isce3/pull/298/commits/9c4312f47af9edd2cadc091e413994ab41d73090. Thanks!

python/packages/nisar/static/doppler.py
gshiroma: Thank you, @Tyler-g-hudson . These numbers do not need to be exact because the Doppler LUT is computed on a relatively coarse grid. The only requirement is that the LUT fully covers the radar grid (with some margin). In `focus.py` the extents of the Doppler LUT is the union of all input L0B products, which is huge.  I also preferred not to hard-code margins that may need to change across different platforms. For example, for an airborne sensor, a fixed az. time margin would represent a completely different relative extent than it would for a spaceborne platform. But I'm also open to suggestions if you have something else in mind.

Tyler-g-hudson (approved):
Your logic makes sense. LGTM!

gshiroma:
Thank you, @Tyler-g-hudson !

seongsujeong (approved):
LGTM

gshiroma:
Thank you, @seongsujeong !

Issue #313, opened 2026-06-18
(Spec non-compliance) dtype Mismatch: `diagnosticModeFlag` (XML uint8 vs HDF5 uint32)


nemo794:
Data type mismatch between XML specification and HDF5 implementation for dataset `/science/LSAR/identification/diagnosticModeFlag`:
  - XML spec: uint8
  - HDF5 file: uint32

  **Impact:** Affects all 8 NISAR L1/L2 product types

  **Products Checked**
- NISAR_L1_PR_RSLC_001_149_D_080_2005_QPDH_A_20251003T130952_20251003T131016_X05016_N_P_J_001
- NISAR_L1_PR_RIFG_013_118_A_171_014_2000_SH_20260222T102514_20260222T102520_20260306T102515_20260306T102521_X05016_P_P_J_001
- NISAR_L2_PR_GUNW_001_062_A_019_006_2000_SH_20250927T114508_20250927T114542_20251126T114516_20251126T114545_X05016_P_F_J_001
  - NISAR_L2_PR_GCOV_001_149_D_080_2005_QPDH_A_20251003T130952_20251003T131016_X05016_N_P_J_001
  - NISAR_L2_PR_GOFF_005_106_A_140_006_7700_SH_20251117T140848_20251117T140925_20251129T140849_20251129T140925_X05016_P_P_J_001
  - NISAR_L1_PR_ROFF_007_106_A_140_008_7700_SH_20251211T140849_20251211T140926_20251223T140850_20251223T140926_X05016_N_P_J_001
  - NISAR_L1_PR_RUNW_007_106_A_140_008_7700_SH_20251211T140849_20251211T140926_20251223T140850_20251223T140926_X05016_N_P_J_001
  - NISAR_L2_PR_GSLC_001_149_D_080_2005_QPDH_A_20251003T130952_20251003T131016_X05016_N_P_J_001

cc: @hfattahi @bhawkins @gshiroma @xhuang-jpl @seongsujeong

gshiroma:
The `diagnosticModeFlag` dataset is copied from the RSLC product into the GCOV and GSLC products. Once the data type of this dataset is corrected in the RSLC product, the GCOV and GSLC products will be fixed as well.

xhuang-jpl:
same with the InSAR products

PR #315, opened 2026-06-18
Update GCOV & GSLC writer to ensure the diagnosticModeFlag is saved with dtype UInt8

gshiroma:
Partially addresses #313.
This PR updates the GCOV and GSLC writers to ensure that the `diagnosticModeFlag` is saved as `UInt8` to match the product specifications, regardless of the RSLC diagnosticModeFlag dtype.


gshiroma:
It's better to fix this in the RSLC product. Once it is fixed there, the GCOV and GSLC products will be fixed as well.

PR #322, opened 2026-06-19
Window correction for absolute calibration factor

### LINKED ISSUE #230: Incorrect Calculation of peak RCS
The calculation of the peak RCS in the file  python/packages/isce3/cal/radar_cross_section.py::measure_target_rcs() is missing a scaling factor.

This issue has been resolved at 

https://github.com/JensReimann/isce3/tree/peak_RCS_fix

but I'm not permitted to open a pull request for it. Please manually pull the changes.

A more detailed description of the problem can be found in the commit. Comments explaining the rationale for the change have been included in the source code.

bhawkins:
This PR fixes #230 by adding support for adjusting absolute calibration (abscal) measurements to account for the radiometric effects of apodization windows applied during SAR focusing.

When measuring corner reflector radar cross-section (RCS) for absolute calibration, the standard "box" method estimates power by measuring peak power and 3dB widths. However, this approach doesn't account for power loss due to windowing (apodization) in range and azimuth processing. This PR implements a correction factor that adjusts measured RCS values to be more comparable with methods that integrate sidelobes (e.g., ESA's approach).

The change is about 0.5 dB for NISAR. The correction can be enabled by selecting "adjusted_box" in the runconfig or CLI (cc @nemo794 since QA does its own schema validation).

### DIFF
--- a/python/packages/nisar/workflows/estimate_abscal_factor.py
+++ b/python/packages/nisar/workflows/estimate_abscal_factor.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 
 import argparse
 import json
+import logging
 import os
 import traceback
 import warnings
@@ -18,6 +19,8 @@ from scipy.optimize import root_scalar
 import isce3
 from isce3.core import abs2
 import nisar
+
+log = logging.getLogger("nisar.workflows.estimate_abscal_factor")
 
 
 def make_irf(weights, normalize=True):
@@ -47,9 +50,8 @@ def get_window_correction(weights):
     return area_win / width_win / (area_box / width_box)
 
 def get_abscal_correction(rslc):
-    with h5py.File(fn, mode="r") as h5:
-        weights_az = h5["/science/LSAR/RSLC/metadata/processingInformation/parameters/azimuthChirpWeighting"][:].astype("f8")
-        weights_rg = h5["/science/LSAR/RSLC/metadata/processingInformation/parameters/rangeChirpWeighting"][:].astype("f8")
+    weights_az = rslc.azimuthChirpWeighting
+    weights_rg, _, _ = rslc.rangeChirpWeighting
     corr_az = get_window_correction(weights_az)
     corr_rg = get_window_correction(weights_rg)
     return corr_rg * corr_az
@@ -152,6 +154,11 @@ def estimate_abscal_factor(
           assumes that the target response can be approximated by a 2-D rectangular
           function. The total power is estimated by multiplying the peak power by the
           3dB response widths in along-track and cross-track directions.
+
+        'adjusted_box':
+          Similar to 'box' but adjusted to account for the effects of
+          apodization windows in range and azimuth.  This should give results
+          more comparable with methods that integrate sidelobes (e.g., ESA).
 
         'integrated':
           Measures power using the integrated power method. The total power is measured
@@ -282,7 +289,12 @@ def estimate_abscal_factor(
     # Get platform attitude data.
     attitude = rslc.getAttitude()
 
-    window_correction = get_abscal_correction(rslc)
+    window_correction = 1.0
+    meas_power_method = power_method
+    if power_method == "adjusted_box":
+        meas_power_method = "box"
+        window_correction = get_abscal_correction(rslc)
+        log.info(f"Will adjust 'box' RCS estimates by {window_correction = }")
 
     # Estimate the absolute calibration error (the ratio of the measured RCS to the
     # predicted RCS) for a single corner reflector.
@@ -308,13 +320,11 @@ def estimate_abscal_factor(
             upsample_factor=upsample_factor,
             peak_find_domain=peak_find_domain,
             nfit=nfit,
-            power_method=power_method,
+            power_method=meas_power_method,
             pthresh=pthresh,
         )
 
-        measured_rcs *= window_correction
-
-        return measured_rcs / predicted_rcs
+        return measured_rcs / predicted_rcs * window_correction
 
     # Estimate the absolute radiometric calibration error of each corner reflector, and
     # format the results into an object that can be easily JSON-ified.
@@ -498,7 +508,7 @@ def parse_cmdline_args() -> dict[str, Any]:
     parser.add_argument(
         "--power-method",
         type=str,
-        choices=["box", "integrated"],
+        choices=["box", "adjusted_box", "integrated"],
         default="box",
         help=(
             "The method for estimating the target signal power (rectangular box method"
--- a/python/packages/nisar/workflows/estimate_abscal_factor.py
+++ b/python/packages/nisar/workflows/estimate_abscal_factor.py
@@ -24,6 +24,21 @@ log = logging.getLogger("nisar.workflows.estimate_abscal_factor")
 
 
 def make_irf(weights, normalize=True):
+    """
+    Create an impulse response function (IRF) from frequency domain weights.
+
+    Parameters
+    ----------
+    weights : array_like
+        Frequency domain weights (e.g., window or filter coefficients).
+    normalize : bool, optional
+        If True, normalize the weights to sum to 1. Defaults to True.
+
+    Returns
+    -------
+    callable
+        A function that evaluates the impulse response at a given time offset t.
+    """
     w = np.fft.fftshift(weights) / len(weights)
     if normalize:
         w *= 1.0 / np.sum(w)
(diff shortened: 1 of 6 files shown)


nemo794:
Thanks for tagging me @bhawkins ! It'd be good for the team to gain more familiarity with `nisarqa`. This is a straightforward update; AbsCal only impacts the RSLC QA workflow. When ready, please issue a PR in `nisarqa`. (Hint: search for "power_method" in the QA codebase to find all of the code locations to consider and/or modify.)

python/packages/nisar/workflows/estimate_abscal_factor.py
```diff
@@ -269,11 +389,11 @@ def estimate_abscal_error(
             upsample_factor=upsample_factor,
             peak_find_domain=peak_find_domain,
             nfit=nfit,
-            power_method=power_method,
+            power_method=meas_power_method,
             pthresh=pthresh,
         )
 
-        return measured_rcs / predicted_rcs
+        return measured_rcs / predicted_rcs * window_correction
```
bhawkins, line 396: Demonstrate that it's correct to multiply instead of divide here.

share/nisar/defaults/focus.yaml
```diff
@@ -963,12 +963,18 @@ runconfig:
...
+                # (e.g., ESA).
+
                 # 'integrated':
                 # Measures power using the integrated power method. The total
                 # power is measured by summing the power of bins whose power
                 # exceeds a predefined minimum power threshold.
-                # Default: box
-                power_method: box
+                # Default: adjusted_box
+                power_method: adjusted_box
```
bhawkins, line 977: Put that in the PSF

bhawkins (commented):
Comments from review 2026-09-11

python/packages/nisar/workflows/estimate_abscal_factor.py
gshiroma: @bhawkins, I believe `predicted_rcs` should be scaled by `window_correction`, so the ratio should be `measured_rcs / predicted_rcs / window_correction`, or, more clearly, `measured_rcs / (predicted_rcs * window_correction)`, no? Please double-check this.

gshiroma (commented):
The PR looks good. I just have one comment before approving it:

python/packages/nisar/workflows/estimate_abscal_factor.py
```diff
@@ -245,6 +358,13 @@ def estimate_abscal_factor(
     # Get platform attitude data.
     attitude = rslc.getAttitude()
 
+    window_correction = 1.0
```
Tyler-g-hudson, line 361: Why use a magic number here? Shouldn't this be a parameter?

Tyler-g-hudson (approved):
One small question, looks good otherwise pending @gshiroma's comments

python/packages/nisar/workflows/estimate_abscal_factor.py
bhawkins: I wouldn't really call this a magic number, it's just the multiplicative identity. So unless you enable the feature and assign the `window_correction` to a different value (a few lines later inside the `if` block), then the RCS doesn't change (x = 1 * y == y). It's just a way to avoid repeating an `if` statement every place the correction might be used.

python/packages/nisar/workflows/estimate_abscal_factor.py
bhawkins: No, the predicted RCS just comes from geometry. It's the measured RCS that needs adjustment based on the window. That's because the "box" method reports the measured RCS as peak * width_range * width_azimuth, but what we really want is the total power in the entire image due to the corner reflector. The discrepancy depends on the window. Here are some whiteboard notes to hopefully make things clearer:
abscal_box_correction
Maybe I should move the `* window_correction` up to two lines to make it clearer it applies to `measured_rcs`.

As a sanity check, I downloaded an RSLC over the Oklahoma CR array and ran the abscal tool with and without the correction factor as follows:
```bash
#!/bin/bash
csv="NISAR_ANC_CORNER_REFLECTORS_008.csv"
rslc="NISAR_L1_PR_RSLC_030_142_D_071_4005_DHDH_A_20260916T012538_20260916T012613_P05023_N_F_J_001.h5"
python3 -m nisar.workflows.estimate_abscal_factor \
    --crs $csv \
    -f A \
    -p HH \
    --rslc $rslc \
    --power-method box \
    -o box.json
python3 -m nisar.workflows.estimate_abscal_factor \
    --crs $csv \
    -f A \
    -p HH \
    --rslc $rslc \
    --power-method adjusted_box \
    -o adjusted_box.json
```

Note that we've already included the 0.5 dB bias in the EXT_CAL file currently in operations. Assuming we did that correctly and that the instrument is relatively stable, we should expect the abscal factor to get 0.5 dB closer to zero with the window adjustment. Here's a small script to analyze the results of the above:
```python
import numpy as np
import json
keys = ("box", "adjusted_box")
filenames = {k: k + ".json" for k in keys}
js = {k: json.load(open(filenames[k])) for k in keys}
acf = {k: np.array([x["absolute_calibration_factor"] for x in js[k]]) for k in keys}
def pow2db(x):
    return 10 * np.log10(x)
geomean_acf_db = {k: float(np.mean(pow2db(acf[k]))) for k in keys}
for k in keys:
    print(f"geometric mean abscal factor for {k} method (dB) = {geomean_acf_db[k]:.1f}")
```
which has the expected result:
```
$ python3 acf.py
geometric mean abscal factor for box method (dB) = 0.7
geometric mean abscal factor for adjusted_box method (dB) = 0.2
```

python/packages/nisar/workflows/estimate_abscal_factor.py
```diff
@@ -3,19 +3,127 @@
...
+    area : float
+        Integrated area under the impulse response power curve.
+    """
+    # Find half-power width using bracketing root-finding algorithm.
+    hw = root_scalar(lambda t: abs2(irf(t)) - 0.5, x0=0.0, x1=1.0).root
+    # full width = 2 * half width
+    width = float(2 * hw)
+
+    # Find area with numerical integration (quadrature).
+    # Integrate from [0, inf) and double the result.
```
gshiroma:
```suggestion
    # Integrate from 0 to `t_max` and double the result. `t_max` defaults to `inf`.
```
It looks like the integration is from `0` to `t_max`, which defaults to `inf`, but can be set to a finite value as well.

python/packages/nisar/workflows/estimate_abscal_factor.py
```diff
@@ -3,19 +3,127 @@
...
+
+    Returns
+    -------
+    float
+        Window correction factor (dimensionless, typically < 1).
+    """
+    # Note that DTFT is periodic, so limit integration to one period.
+    t_max = len(weights) / 2
+    width_win, area_win = get_irf_width_area(make_irf(weights), t_max=t_max)
+    width_box, area_box = get_irf_width_area(make_irf(weights > 0.0), t_max=t_max)
```
gshiroma:
```suggestion
    width_box, area_box = get_irf_width_area(make_irf(np.absolute(weights) > 0.0), t_max=t_max)
```
If `weights` can be complex (or negative), you probably want to take their absolute values.

python/packages/nisar/workflows/estimate_abscal_factor.py
```diff
@@ -3,19 +3,127 @@
...
+    weights : array_like
+        Frequency domain weights representing the window function.
+
+    Returns
+    -------
+    float
+        Window correction factor (dimensionless, typically < 1).
+    """
+    # Note that DTFT is periodic, so limit integration to one period.
+    t_max = len(weights) / 2
```
gshiroma: I thought `t_max` should be given in seconds?

python/packages/nisar/workflows/estimate_abscal_factor.py
```diff
@@ -3,19 +3,127 @@
...
+        complex value.
+    t_max : float, optional
+        Maximum integration limit for computing the area. Defaults to infinity.
+
+    Returns
+    -------
+    width : float
+        Full width at half maximum (FWHM) of the impulse response power.
+    area : float
+        Integrated area under the impulse response power curve.
```
gshiroma: Maybe add that the integration is done from `-t_max` to `t_max`? Or from `0` to `t_max` and doubled?

python/packages/nisar/workflows/estimate_abscal_factor.py
gshiroma: Thank you, @bhawkins, for providing all this information.

Regarding scaling `measured_rsc` or `predicted_rsc`, I think you could go either way, depending on which quantity you use as the reference (window or box). I’m fine with whichever approach you prefer, as long as everything is applied consistently.

I think I understand what I was getting wrong. You define the window correction in each direction as

```(area_win / width_win) / (area_box / width_box)```

which is actually the inverse of what I had in mind. I assumed `area_win` was the area under the function limited by the 3dB boundary, which is not the case. That would be something like `peak_win * width_win`. Since window and box functions are normalized in the spectrum, their peaks in time (at t=0) are both one. So, the "area" I had in mind was actually represented by `width_win` and `width_box`. With the peaks, maybe it would be something like:

```[area_win / (peak_win * width_win)] / [area_box / (peak_box * width_box)]```

And yes, I think this factor can then be applied multiplicatively to an RCS calculated using the windowed measured response to convert it to the equivalent box response. So I think I understand it now, and everything looks correct to me. Thanks again.

gshiroma (approved):
Thank you, @bhawkins , for addressing my comments. My comments above are minor. The PR looks good to me as is.

PR #325, opened 2026-06-21
Support for Second-band Channel Adjustments in INT-CAL HDF5

rad-eng-59:
This PR adds support for the TX/RX channel adjustment factors of the second frequency band (aka `B`) in instrument INT-CAL HDF5.
The EAP of RSLC workflow utilizes this feature simply for split-spectrum case where the second frequency band requires different adjustment than the first one (aka `A`) on either of TX/RX side for both polarization H/V due to *relatively large separation* between two frequency bands. This is necessary to account properly for frequency-dependency of the instrument in particular on band-limited RX side!

This PR adds the following features:
- Add a new dataset to instrument INT-CAL HDF5 for the TX/RX complex fudge factors of second frequency band, called `amplitudeB`.
- Update the instrument parser to read newly added dataset if available. If not, it will use the values for the primary band (backward compatible from the perspective of both ISCE3 python API and product version)
- Bump the INT-CAL product version to `v3.0` and provided sample product for testing. 
- EAP module `pattern.py` now uses this new feature per frequency band if available for only the second band of split spectrum case (SSP).


tests/python/packages/nisar/products/readers/instrument_parser.py
```diff
@@ -151,3 +151,82 @@ def test_instrument_parser_v2p0():
...
+    adj_fact_rx['V'] = np.full(
+        shape=(12,),
+        fill_value=_powdb_phasedeg_to_complex(-0.2, -20),
+        dtype='c16'
+    )
+
+    # construct the parser
+    with InstrumentParser(os.path.join(iscetest.data, sub_dir,
+                                       instrument_file)) as ins:
+        for pol in ('H', 'V'):
```
bhawkins, line 218: I think it'd be a good idea to add the `is_second_band=False` cases here to make sure it pulls out the ones for frequency A.

bhawkins (approved):
The logic is a little hard to follow, but I think it makes sense. We have the following cases to worry about:
* Sahara mode (RCID 153) where we have 5 MHz freq B but it's at the lower side of the band, so `is_second_band` is correctly set to `False`.
* Quasi-dual cases (RCID 135) where we have 5 MHz freq B but again they're at the lower side of the band, so `is_second_band` is correctly set to `False`.
* Split spectrum cases with missing data get handled correctly because A & B arrive in the same file (VCID), so the checks by polarization are fine.
* Other split spectrum cases work as expected.

This requires that modify the INT_CAL file, so it'd have to be redelivered for SDS to take advantage of this new feature. We may want to coordinate that update with whatever strategy we use for the Type 7 table issue.

Tyler-g-hudson (approved):
LGTM per offline discussion

rad-eng-59:
@bhawkins  @Tyler-g-hudson 
FYI, Already augmented  the INT-CAL test suite per off-line discussion.

PR #327, opened 2026-06-24
Handle case where no samples in a rangline are valid

bhawkins:
This patch allows RSLC processing to continue when there are pulses in the L0B with no samples marked as valid. Without these changes one will get an error message like
```
  File "/opt/conda/lib/python3.12/site-packages/isce3/focus/valid_regions.py", line 554, in get_gap_slices
    yield slice(None, pulse_swaths[0, 0])
                      ~~~~~~~~~~~~^^^^^^
IndexError: too many indices for array: array is 1-dimensional, but 2 were indexed
```
While this limitation has been around for a while, it gains new relevance with the recent L0B processor update that marks the first 500 pulses of a datatake as invalid during the HPA warm-up period.

The heart of this patch is the four lines added to `isce3.focus.fill_gaps`. The rest of the changes are just to generate some concise log messages to help debugging cases with missing data down the line.

### DIFF
--- a/python/packages/isce3/focus/valid_regions.py
+++ b/python/packages/isce3/focus/valid_regions.py
@@ -597,7 +597,7 @@ def false_runs(arr: np.ndarray) -> list[slice]:
     padded = np.concatenate(([True], arr, [True]))
 
     # Find where transitions occur
-    diff = np.diff(padded.view(np.uint8))
+    diff = np.diff(padded.view(np.int8))
 
     # A False run starts where True→False (diff == -1), ends where False→True (diff == 1)
     starts = np.where(diff == -1)[0]  # index in original array
--- a/python/packages/isce3/focus/valid_regions.py
+++ b/python/packages/isce3/focus/valid_regions.py
@@ -549,6 +549,10 @@ def fill_gaps(data, swaths, value=np.complex64(0)):
         pulse_swaths = np.asarray(
             [swath for swath in pulse_swaths if swath[1] > swath[0]]
         )
+        # If there's no valid data then the whole rangeline is a "gap".
+        if len(pulse_swaths == 0):
+            yield slice(None)
+            return
         num_swaths = pulse_swaths.shape[0]
         # Gap leading up to first swath.
         yield slice(None, pulse_swaths[0, 0])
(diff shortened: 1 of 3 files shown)


python/packages/isce3/focus/valid_regions.py
```diff
@@ -549,6 +549,10 @@ def get_gap_slices(pulse_swaths: np.ndarray) -> Iterator[slice]:
         pulse_swaths = np.asarray(
             [swath for swath in pulse_swaths if swath[1] > swath[0]]
         )
+        # If there's no valid data then the whole rangeline is a "gap".
+        if len(pulse_swaths) == 0:
+            yield slice(None)
+            return
```
hfattahi, line 555: Even though in reality and during warmup the entire pulse is not invalid but first 2/3rd and then 1/3rd is missing, looks like we have decided to mark all invalid in L0B. Anyways that is an upstream discussion which we already had. So that is ok to give up on the warm up part in RSLC. 
Out of curiosity what happens if the entire L0B has a mask that marks all pulses invalid? Does focus detect and complain, or with this minor change we are letting that go through?

bhawkins: I guess that's true about the first 1/3rd of the swath. We'd want to test that EAP does the right thing in that case.

If all pulses are marked completely invalid then yes processing would continue and the result would be an image of all zeros. Similarly, a block of 5 seconds of missing data would no longer be a fatal error. That's why I added the extra logging so there's at least some visibility into those sorts of edge cases.

The L0B .met file contains information about missing rangelines, so PCM could implement a rule to avoid RLSC processing stuff with missing data.

hfattahi: I agree that it is the responsibility of job manager (PCM) to check if we want a threshold for not submitting job. I think it is actually a good thing to let data be processed even if there is 5 sec or whatever gap in the data. Again somebody else's job (e.g, QA) to accept the partial SLC or not!

hfattahi (approved):
LGTM. Thanks @bhawkins

Tyler-g-hudson (approved):
LGTM, thanks!

python/packages/nisar/workflows/focus.py
```diff
@@ -1750,6 +1751,16 @@ def get_caltone_algorithm(cfg, fc, fs, n, is_dithered):
 
     return algorithm, wavelets
 
+
+def log_bad_pulses(swaths, max_slices=10):
```
rad-eng-59, line 1755: I assume the purpose of the `max_slices` is to avoid crowding the log/screen with too many bad range lines.

rad-eng-59 (approved):
LGTM. Thanks.

PR #328, opened 2026-06-24
Add radar grid decimation option to compute the static layers layover/shadow mask

gshiroma:
This PR adds support for radar grid decimation when computing the static layers layover/shadow mask. This is configured through the new `az_decimation_to_compute_layover_shadow_mask` and `rg_decimation_to_compute_layover_shadow_mask` parameters in the STATIC runconfig.


share/nisar/defaults/static.yaml
```diff
@@ -249,32 +249,32 @@ runconfig:
           # Must be >= 2.
           # Defaults to 5.
           pts_per_side: 5
+          # [OPTIONAL] Azimuth decimation factor to compute the layover shadow mask.
+          az_decimation_to_compute_layover_shadow_mask: 1
+          # [OPTIONAL] Range decimation factor to compute the layover shadow mask.
+          rg_decimation_to_compute_layover_shadow_mask: 1
```
Tyler-g-hudson:
```suggestion
          # [OPTIONAL] Azimuth decimation factor to compute the layover shadow mask in radar grid coordinates.
          layover_shadow_mask_az_decimation: 1
          # [OPTIONAL] Range decimation factor to compute the layover shadow mask in radar grid coordinates.
          layover_shadow_mask_rg_decimation: 1
```
This name is a little verbose, I think these compressed ones work better.

Tyler-g-hudson (approved):
I have only a nitpick about naming - note that this would need to extend to the schema and the places where the runconfig is accessed in static.py

share/nisar/defaults/static.yaml
gshiroma: I like these variable names. Thank you, @Tyler-g-hudson !

gshiroma: Updated in [1e9f2e4](https://github.com/isce-framework/isce3/pull/328/commits/1e9f2e47c4096099cab309c662bb8fcc75060d6f)

hfattahi (approved):
LGTM

PR #330, opened 2026-06-26
Fix the troposphere phase screen datacube across the dateline

xhuang-jpl:
**Problem:** The AOI bounding box used plain `min`/`max` over radar-grid longitudes. For scenes crossing ±180°, wrapped longitudes (e.g. `+179°`/`−179°`) inflate the AOI to ~358° wide, so RAiDER loads a near-global weather model and the delay cube is wrong for dateline-crossing acquisitions.

**Fix:** Wrap `raider_tropo_delay` in `compute_tropo_delay_dateline`, which detects the crossing and only then splits the output grid into east/west sub-AOIs, runs each against its own weather-model crop, and merges the cubes back along `x` (reassembled in original column order). Non-crossing scenes take a single call with unchanged behavior.

**Runtime:** It takes about **64** seconds on the c5.4xlarge instance, and the plot looks like 

image

### DIFF
--- a/python/packages/nisar/workflows/troposphere.py
+++ b/python/packages/nisar/workflows/troposphere.py
@@ -199,7 +199,7 @@ def compute_troposphere_delay(cfg: dict, gunw_hdf5: str):
                 Parameters
                 ----------
                 lon_datacube: np.ndarray
-                    radar-grid longitudes, last axis = range/x
+                    radar-grid longitudes in degrees, last axis = range/x
 
                 Returns
                 -------
@@ -220,13 +220,13 @@ def compute_troposphere_delay(cfg: dict, gunw_hdf5: str):
                 Parameters
                 ----------
                 lat: np.ndarray
-                    radar-grid latitudes, last axis = range/x
+                    radar-grid latitudes in degrees, last axis = range/x
                 lon: np.ndarray
-                    radar-grid longitudes, last axis = range/x
+                    radar-grid longitudes in degrees, last axis = range/x
                 cols: np.ndarray
                     column indices belonging to this side
                 side: str
-                    one of 'west', 'east', or 'all'
+                    one of 'west_dateline', 'east_dateline', or 'all'
                 margin: float
                     bounding-box padding in degrees
 
@@ -238,10 +238,10 @@ def compute_troposphere_delay(cfg: dict, gunw_hdf5: str):
                 lon = np.take(_norm_lon(lon), cols, axis=-1)
                 s = float(np.nanmin(lat)) - margin
                 n = float(np.nanmax(lat)) + margin
-                if side == "west":                                # +lon hemisphere
+                if side == "west_dateline":                                # +lon hemisphere
                     pos = lon[np.isfinite(lon) & (lon >= 0)]
                     return [s, n, float(np.min(pos)) - margin, 180.0]
-                if side == "east":                                # -lon hemisphere
+                if side == "east_dateline":                                # -lon hemisphere
                     neg = lon[np.isfinite(lon) & (lon < 0)]
                     return [s, n, -180.0, float(np.max(neg)) + margin]
                 return [s, n, float(np.nanmin(lon)) - margin, float(np.nanmax(lon)) + margin]
@@ -250,8 +250,7 @@ def compute_troposphere_delay(cfg: dict, gunw_hdf5: str):
                                     xpts, ypts, los, height_levels, out_proj,
                                     weather_model_type=None,
                                     weather_model_output_dir=None,
-                                    hres_converter=None, margin=0.1,
-                                    tropo_delay_fn=None):
+                                    hres_converter=None, margin=0.1):
                 '''
                 Compute the RAiDER tropospheric delay with antimeridian handling.
 
@@ -268,9 +267,9 @@ def compute_troposphere_delay(cfg: dict, gunw_hdf5: str):
                     weather model file (RAiDER NetCDF, or HRES NetCDF if
                     weather_model_type == 'HRES')
                 lat_datacube: np.ndarray
-                    radar-grid latitudes, last axis = range/x
+                    radar-grid latitudes in degrees, last axis = range/x
                 lon_datacube: np.ndarray
-                    radar-grid longitudes, last axis = range/x
+                    radar-grid longitudes in degrees, last axis = range/x
                 xpts: np.ndarray
                     output-grid x coordinates (xcoord_radar_grid)
                 ypts: np.ndarray
@@ -290,14 +289,11 @@ def compute_troposphere_delay(cfg: dict, gunw_hdf5: str):
                     used to convert an HRES NetCDF to the RAiDER internal NetCDF per sub-AOI
                 margin: float, optional
                     AOI padding in degrees (default 0.1)
-                tropo_delay_fn: callable, optional
-                    override for RAiDER's tropo_delay (defaults to raider_tropo_delay)
 
                 Returns
                 -------
                     the (merged) xarray.Dataset tropospheric delay cube
                 '''
-                rtd = tropo_delay_fn or raider_tropo_delay
 
                 xpts = np.asarray(xpts, float)
                 ypts = np.asarray(ypts, float)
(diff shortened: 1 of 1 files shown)


python/packages/nisar/workflows/troposphere.py
```diff
@@ -157,6 +157,178 @@ def compute_troposphere_delay(cfg: dict, gunw_hdf5: str):
...
+                ----------
+                lon: array_like
+                    longitudes in degrees
+
+                Returns
+                -------
+                    True if the longitude span exceeds 180 deg (i.e. crosses +/-180)
+                '''
+                lon = _norm_lon(lon)
+                return bool(np.nanmax(lon) - np.nanmin(lon) > 180.0)
```
bhawkins, line 189: Note that this trick only works up to about ±89° latitude. (See [here](https://github.com/isce-framework/nisarqa/pull/161#discussion_r3229957436) for rationale.)

python/packages/nisar/workflows/troposphere.py
```diff
@@ -157,6 +157,178 @@ def compute_troposphere_delay(cfg: dict, gunw_hdf5: str):
...
+                ----------
+                lon_datacube: np.ndarray
+                    radar-grid longitudes, last axis = range/x
+
+                Returns
+                -------
+                    1-D array of representative longitude per x-column
+                '''
+                lon = _norm_lon(lon_datacube)
+                if np.nanmax(lon) - np.nanmin(lon) > 180.0:
```
bhawkins, line 209: Same comment: this breaks within about 1° of the poles.

python/packages/nisar/workflows/troposphere.py
```diff
@@ -157,6 +157,178 @@ def compute_troposphere_delay(cfg: dict, gunw_hdf5: str):
...
+
+                For a split side the box is extended to +/-180 so the east and west halves
+                meet at the seam; for side='all' it is the plain nan-safe min/max.
+
+                Parameters
+                ----------
+                lat: np.ndarray
+                    radar-grid latitudes, last axis = range/x
+                lon: np.ndarray
+                    radar-grid longitudes, last axis = range/x
```
bhawkins:
```suggestion
                lat: np.ndarray
                    radar-grid latitudes in degrees, last axis = range/x
                lon: np.ndarray
                    radar-grid longitudes in degrees, last axis = range/x
```

python/packages/nisar/workflows/troposphere.py
```diff
@@ -157,6 +157,178 @@ def compute_troposphere_delay(cfg: dict, gunw_hdf5: str):
...
+                    bounding-box padding in degrees
+
+                Returns
+                -------
+                    bounding box as [S, N, W, E] in degrees
+                '''
+                lat = np.take(np.asarray(lat, float), cols, axis=-1)
+                lon = np.take(_norm_lon(lon), cols, axis=-1)
+                s = float(np.nanmin(lat)) - margin
+                n = float(np.nanmax(lat)) + margin
```
bhawkins, line 240: This could problematic if the scene contains a pole.

python/packages/nisar/workflows/troposphere.py
```diff
@@ -157,6 +157,178 @@ def compute_troposphere_delay(cfg: dict, gunw_hdf5: str):
...
+                ----------
+                dt: datetime
+                    acquisition datetime, passed through to RAiDER
+                weather_model_file: str
+                    weather model file (RAiDER NetCDF, or HRES NetCDF if
+                    weather_model_type == 'HRES')
+                lat_datacube: np.ndarray
+                    radar-grid latitudes, last axis = range/x
+                lon_datacube: np.ndarray
+                    radar-grid longitudes, last axis = range/x
```
bhawkins: Units?

python/packages/nisar/workflows/troposphere.py
```diff
@@ -157,6 +157,178 @@ def compute_troposphere_delay(cfg: dict, gunw_hdf5: str):
...
+                Parameters
+                ----------
+                lat: np.ndarray
+                    radar-grid latitudes, last axis = range/x
+                lon: np.ndarray
+                    radar-grid longitudes, last axis = range/x
+                cols: np.ndarray
+                    column indices belonging to this side
+                side: str
+                    one of 'west', 'east', or 'all'
```
bhawkins:
```suggestion
                    Side relative to the dateline. one of 'west', 'east', or 'all'
```
I was very confused until I realized "west" means "west of the dateline" instead of "western hemisphere", which are completely opposite meanings. Need to spell this out for the code to make any sense.

python/packages/nisar/workflows/troposphere.py
```diff
@@ -157,6 +157,178 @@ def compute_troposphere_delay(cfg: dict, gunw_hdf5: str):
...
+                merged = None
+                for side, cols in (("west", np.where(rep <= 180.0)[0]),
+                                ("east", np.where(rep > 180.0)[0])):
+                    if cols.size == 0:
+                        continue
+                    bounds = _side_bounds(lat_datacube, lon_datacube, cols, side, margin)
+                    td = _run(bounds, xpts[cols]).assign_coords(_xidx=("x", cols))
+                    merged = td if merged is None else xr.concat([merged, td], dim="x")
+                return merged.sortby("_xidx").drop_vars("_xidx")
+
```
bhawkins, line 330:
```suggestion
                    td = _run(bounds, xpts[cols]).assign_coords(_xidx=("x", rep[cols]))
                    merged = td if merged is None else xr.concat([merged, td], dim="x")
                return merged.sortby("_xidx").drop_vars("_xidx")

```
I'm not familiar with xarray, but I'm guessing this where the dateline gap is coming from. If you use the wrapped longitudes `rep[cols]` instead of the column indices as `_xidx` does the gap go away?

bhawkins (commented):
Thanks for the patch. The approach makes sense, though it has some limitations.

python/packages/nisar/workflows/troposphere.py
```diff
@@ -157,6 +157,178 @@ def compute_troposphere_delay(cfg: dict, gunw_hdf5: str):
...
+                margin: float, optional
+                    AOI padding in degrees (default 0.1)
+                tropo_delay_fn: callable, optional
+                    override for RAiDER's tropo_delay (defaults to raider_tropo_delay)
+
+                Returns
+                -------
+                    the (merged) xarray.Dataset tropospheric delay cube
+                '''
+                rtd = tropo_delay_fn or raider_tropo_delay
```
oberonia78: I'm not sure why `tropo_delay_fn` is introduced. Previously, it seems that you used the `radar_tropo_delay` at default, and I don't see the usage of `tropo_delay_fn`.

python/packages/nisar/workflows/troposphere.py
xhuang-jpl: The gap comes from the interpolation by RAiDER. This is because when it is across the dateline, the RAiDER will crop the weather model files to be within [xxx, 180] or [-180,xxx], which means that across the dateline, it might not have sufficient data for the interpolation.

python/packages/nisar/workflows/troposphere.py
xhuang-jpl: That is good point, I have checked our track frame boundingbox, which  is 

`-180.0 -88.08080392809086 180.0 77.61184575403183`

it seems to be within -89 and 89.

python/packages/nisar/workflows/troposphere.py
xhuang-jpl: it is a parameter of the compute_tropo_delay function and its default is radar_tropo_delay, I can delete it.

python/packages/nisar/workflows/troposphere.py
xhuang-jpl: Yes, the west and east means the west  dateline, i.e. the eastern hemisphere. I will spell it out to make it clear

python/packages/nisar/workflows/troposphere.py
xhuang-jpl: degrees

bhawkins (approved):
LGTM, thanks. Hopefully we can address the remaining gap in a future PR.

hfattahi (approved):
We reviewed in a PR review and these changes look good to me. 
Please open an issue for the remaining artifact at the dateline boundary.

PR #332, opened 2026-07-07
Improve error message related to missing telemetry (NSDS-5235)

bhawkins:
SDS has reported several errors related to missing telemetry. The interpretation of the log message isn't so obvious, especially since the logic in an error handling block itself contains an error, so one gets two exception tracebacks. This PR fixes the problem in the low-level error handling and adds a more informative error message in the higher-level code.

The actual problem at hand is that data were collected in a dual-pol mode (HH and HV), but presumably something went awry with the downlink such that only HV made it into the L0B file. The processor is configured with elevation antenna pattern (EAP) correction enabled, and the loopback calibration telemetry is required to compute the patterns. However, there's no such thing as crosspol loopback cal; the transmit-H pattern must be formed using analysis of the HH loopback cal. Since that isn't available in the file, the EAP processing cannot proceed.

### DIFF
--- a/python/packages/nisar/antenna/pattern.py
+++ b/python/packages/nisar/antenna/pattern.py
@@ -3,6 +3,7 @@ from collections import defaultdict
 from isce3.core import Orbit, Attitude, Linspace
 from isce3.geometry import DEMInterpolator
 import logging
+from pathlib import Path
 from nisar.products.readers.antenna import AntennaParser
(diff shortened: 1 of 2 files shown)


hfattahi (approved):
Thanks for improving the log. I like that the message still gives an option to user to process a cross pol only data. Thanks @bhawkins

Tyler-g-hudson (approved):

PR #334, opened 2026-07-09
Update STATIC workflow for new water mask spec

nemo794:
This PR updates to the STATIC Workflow to reflect the current specification for the NISAR Water Mask, which uses the classifications (Per JPL D-107710):

```
Layer classes:
0: Not water – an area with valid data that is not open water (class 1), high-backscattering water (class 2), low-backscattering
non-water (class 3), landcover masked (class 4), height above nearest drainage masked (class 5), layover/shadow masked
(class 6), or ocean masked (class 254). Masking can result in “not water” (class 0) where land cover masking is applied.
1 -100: Ocean distance – distance from the coastline (unit: km)
101 -200: Inland water distance – distance from the boundary between inland water and land (unit: km)
255: Fill value (no data)
```

### Background

When the Static Workflow was originally implemented in #144 , the classification used by the NISAR water mask at that time was:

https://github.com/isce-framework/isce3/blob/bdf1f6fb9175ffe9e53b61839c797e69947dd268/python/packages/nisar/static/water_mask.py#L21-L24

However, for the current water mask TIFF files being used in mission operations, 0 designates not-water, and 1-200 designates water. Here is a screenshot of `NISAR_water_mask_N33_N34_W119_W118_30m_v0.5.0.tif`, provided by M Zahn: 

Screenshot 2026-07-08 at 5 06 38 PM

Due to this discrepancy, the static layers workflow was generating `waterMask` layers with the binary values 0 and 1 inverted. This PR resolves the issue.

### DIFF
--- a/python/packages/nisar/static/water_mask.py
+++ b/python/packages/nisar/static/water_mask.py
@@ -19,8 +19,8 @@ def binarize_nisar_water_mask(water_distance: ArrayLike) -> np.ndarray:
     ----------
     water_distance : array_like
         The input water distance map, in the format specified by the NISAR Water Mask
-        Product Specification\ [1]_. A value of 0 indicates a water pixel. A value of
-        255 represents a no-data (invalid) pixel. Values in 1-200 represent non-water
+        Product Specification\ [1]_. A value of 0 indicates a not-water pixel. A value
+        of 255 represents a no-data (invalid) pixel. Values in 1-200 represent water
         pixels.
 
     Returns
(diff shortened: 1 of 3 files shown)


hfattahi:
Thanks for fixing the bug @nemo794 . Would you please confirm if you or Max have run this PR on one example and the water mask is fixed?

jplzhan:
@hfattahi I was able to run this PR last week on 4 examples, which I uploaded to S3. Has anyone had the chance to take a look?

nemo794:
Using the HDF5s @jplzhan generated with this PR in the production pipeline, it appears that the water mask values are corrected for the water mask. Below are screenshots from those granules, where Blue:Water and Brown:Not Water

Screenshot 2026-07-13 at 11 36 26 AM
Screenshot 2026-07-13 at 11 31 58 AM
Screenshot 2026-07-13 at 11 32 30 AM
Screenshot 2026-07-13 at 11 32 40 AM

nemo794:
I also added a fresh commit, which makes the classification values explicit in the STATIC runconfig's comments. (No code logic change.)

@hfattahi @oberonia78 Could you please re-review when you have a chance? Thanks!

hfattahi (approved):
thank you @nemo794 . LGTM.

gshiroma (approved):
LGTM. Thank you, @nemo794 .

nemo794:
Thank you, @gshiroma and @hfattahi !

PR #205, opened 2026-01-21
Expose sensing start/end times and starting/ending ranges in the STATIC runconfig

gshiroma:
This PR follows #178 and exposes sensing start/end times and starting/ending ranges in the static layers (STATIC) product runconfig:

```
...
bounding_box:
  sensing_start: num(required=False)
  sensing_end: num(required=False)
  starting_range: num(required=False)
  ending_range: num(required=False)
...
```

PR #178, opened 2025-11-10
Enable selection of range spacing or azimuth time interval in the STATIC runconfig

gshiroma:
This PR addresses the 3rd item in https://github.com/isce-framework/isce3/issues/158. It exposes an option in the STATIC runconfig to select either the range spacing or the azimuth time interval. If one of these values is not provided, the STATIC workflow will infer it using the radar parameters and the geogrid.

PR #337, opened 2026-07-15
Bugfix: STATIC Layover Shadow Mask geocoding fill not set to `255`

nemo794:
WIP - unit tests are in-progress. Code is ready for review/discussion.

# Summary

When the STATIC workflow was first implemented in #144 , each raster layer provided complete coverage of the given track frame's extent; there was never any geocoding fill.

Then, #178 and #205 introduced the option into the STATIC workflow to only compute the science values for a smaller swath within that track frame, to e.g. reduce memory usage. This meant that several STATIC layer rasters would now include geocoding fill regions around the corners/edges of the rasters.

For raster layers that are float-valued, the geocoding fill pixels was correctly set to `NaN`.

However, for the layover shadow mask, the geocoding fill pixels were incorrectly set to `0`. This bug is due to a limitation of `GeocodeCov` not being able to natively handle geocoding `uint8` layers (Full details noted in the in-line code comments).

This PR resolves that issue so that the layover shadow mask layer has the geocoding fill correctly set to `255`.

# Technical Design

## Considerations

Related Ideas to keep in mind when selecting the correct implementation design:

1) This issue also impacts GUNW's Connected Components layer, and potentially others.
* For example, if `nisarqa`decides to generate PNGs for connected components, it would need to implement a similar fix.
* **Whatever approach is decided here will need to be adopted by other workflows.**

2) Choice of Design Pattern
* Design Option 1 - use `out_mask`
    - Currently, this PR implements a fix by having the calling function setup an `out_mask` parameter, which `GeocodeCov` then populates with the geocoding fill area. The calling function is then responsible for applying that mask to set the geocoding fill pixel values correctly to 255.
        * Pros: reduces number of rasters required to be in memory
        * Cons: requires an extra scratch Geotiff file
* Design Option 2 - convert to `float32`
    * A different approach could be to 1) Convert the uint8 mask to float32, replacing 255 with NaN, then 2) Geocode the float32 raster (NaN values are preserved), then 3) Convert back to uint8, replacing NaN with 255.
        * Cons: requires add'l copies of the raster to be in memory
* Design Option 3 - wrap `uint8` values
    * A trick with `uint8` is that if you add `255`, then instead of overflow, it will wrap values back around to e.g. `0`. This works with `numpy`, would need more research re: robustness in other situations.
    * Pseudo code: 1) Take the `uint8` mask layer and add `1` to all values. 2) Geocode; the fill values will be set to `0`. 3) Add `255` to the geocoded mask. This will set geocoding fill to `255` (correct), and all other values will get wrapped back to their original correct values.
* Design Option 4 - Choose 1, 2, or 3, but implement in a light wrapper around `GeocodeCov`
    * Require 1-line change to all calling functions
    * Pros:
        * Does not touch `GeocodeCov`, which can be finnicky
        * Reasonable to maintain long-term
    * Users with `uint8` datasets will need to remember to call the wrapper function
* Design Option 5 - Choose 1, 2, or 3, but implement internal to `GeocodeCov`
    * Should automatically resolve the issue across all products with minimal changes to their implementation
        * Pros: Reduces technical debt. Will improve user experience in the future.
        * Cons: Potentially messier refactor, which touches a core part of the ISCE3 code.

## Design Implemented

This PR implements Design Option 1 for only the STATIC Layers workflow.

Open to discussion about whether a different design would be preferred.

# Test Results

Tested locally. Once the design pattern is approved, @jplzhan could you please test this branch in the production pipeline?

## Before this PR

Screenshot 2026-07-15 at 3 35 29 PM

## After this PR

(Fill Value of 255 is a lighter grey; it is not noted in the colorbar. I'll update the plot if/when there's time)

Screenshot 2026-07-15 at 3 30 46 PM

PDF illustrating all layers, and how the radar grid varies between layers:
[NISAR_L2_STATIC_023_D_067_0800_0800_19991231T235959_A10000_J_001_report.pdf](https://github.com/user-attachments/files/30067715/NISAR_L2_STATIC_023_D_067_0800_0800_19991231T235959_A10000_J_001_report.pdf)

cc: @hfattahi

### DIFF
--- a/python/packages/nisar/static/layover_shadow_mask.py
+++ b/python/packages/nisar/static/layover_shadow_mask.py
@@ -195,19 +195,49 @@ def geocode_layover_shadow_mask(
     if geo2rdr_params is None:
         geo2rdr_params = {}
 
-    geocoded_layover_shadow_mask = make_scratch_gtiff(
-        shape=(geo_grid.length, geo_grid.width),
-        dtype=np.uint8,
+    # XXX: There's currently no way to create a "Geocode" object that operates directly
+    # on uint8 data with Fill value 255. Instead, the way this is done in the
+    # InSAR geocoding workflow is to create a `GeocodeFloat32` object, resulting
+    # in conversion of the input mask values to float32 during geocoding (and
+    # then conversion back to uint8 when they're written to the output raster).
+    # For STATIC, this shouldn't cause any loss of fidelity since the mask values
+    # are small and we're using nearest neighbor interpolation.
+
+    # However, when geocoding uint8 data with `GeocodeFloat32`, the C++ `GeocodeCov`
+    # implementation initializes the output buffer with `nan_t_out`. Then,
+    # 0 * NaN evaluates to NaN for float dtypes (correct), but for integer
+    # dtypes it evaluates to 0 (since uint8 cannot represent NaN). This leads to
+    # areas outside the radar grid extent retaining this 0 fill value instead
+    # of the correct 255.
+
+    # Workaround:
+    # 1. Convert the uint8 mask to float32, replacing 255 with NaN
+    # 2. Geocode the float32 raster (NaN values are preserved)
+    # 3. Convert back to uint8, replacing NaN with 255
+
+    layover_shadow_mask_float32 = make_scratch_gtiff(
+        shape=(radar_grid.length, radar_grid.width),
+        dtype=np.float32,
         dir_=scratch_dir,
-        prefix="geocoded-layover-shadow-mask_",
+        prefix="layover-shadow-mask-float32_",
     )
 
-    # XXX: There's currently no way to create a "Geocode" object that operates directly
-    # on uint8 data. Instead, the way this is done in the InSAR geocoding workflow is to
-    # create a `GeocodeFloat32` object, resulting in conversion of the input mask values
-    # to float32 during geocoding (and then conversion back to uint8 when they're
-    # written to the output raster). This shouldn't cause any loss of fidelity since the
-    # mask values are small and we're using nearest neighbor interpolation.
+    # Convert uint8 mask to float32, replacing fill value 255 with NaN
+    radar_data_uint8 = np.empty((radar_grid.length, radar_grid.width), dtype=np.uint8)
+    layover_shadow_mask.read_all(radar_data_uint8)
+    radar_data_float32 = radar_data_uint8.astype(np.float32)
+    radar_data_float32[radar_data_uint8 == 255] = np.nan
+    layover_shadow_mask_float32.write_all(radar_data_float32)
+
+    # Create output float32 raster for geocoded result
+    geocoded_layover_shadow_mask_float32 = make_scratch_gtiff(
+        shape=(geo_grid.length, geo_grid.width),
+        dtype=np.float32,
+        dir_=scratch_dir,
+        prefix="geocoded-layover-shadow-mask-float32_",
+    )
+
+    # Geocode the float32 mask
     geocode = isce3.geocode.GeocodeFloat32()
     geocode.orbit = orbit
     geocode.ellipsoid = get_reference_ellipsoid(dem_raster)
@@ -230,8 +260,8 @@ def geocode_layover_shadow_mask(
 
     geocode.geocode(
         radar_grid=radar_grid,
-        input_raster=layover_shadow_mask,
-        output_raster=geocoded_layover_shadow_mask,
+        input_raster=layover_shadow_mask_float32,
+        output_raster=geocoded_layover_shadow_mask_float32,
         dem_raster=dem_raster,
         output_mode=GeocodeOutputMode.INTERP,
         memory_mode=normalize_geocode_memory_mode(memory_mode),
@@ -239,6 +269,18 @@ def geocode_layover_shadow_mask(
         max_block_size=max_block_size,
         dem_interp_method=normalize_data_interp_method(dem_interp_method),
     )
+
+    # Convert geocoded float32 result back to uint8, replacing NaN with 255
+    geocoded_layover_shadow_mask = make_scratch_gtiff(
+        shape=(geo_grid.length, geo_grid.width),
+        dtype=np.uint8,
+        dir_=scratch_dir,
+        prefix="geocoded-layover-shadow-mask_",
+    )
+    geo_data_float32 = np.empty((geo_grid.length, geo_grid.width), dtype=np.float32)
+    geocoded_layover_shadow_mask_float32.read_all(geo_data_float32)
+    geo_data_uint8 = np.where(np.isnan(geo_data_float32), 255, geo_data_float32).astype(np.uint8)
+    geocoded_layover_shadow_mask.write_all(geo_data_uint8)
 
     return geocoded_layover_shadow_mask
 
--- a/python/packages/nisar/static/layover_shadow_mask.py
+++ b/python/packages/nisar/static/layover_shadow_mask.py
@@ -196,48 +196,42 @@ def geocode_layover_shadow_mask(
         geo2rdr_params = {}
 
     # XXX: There's currently no way to create a "Geocode" object that operates directly
-    # on uint8 data with Fill value 255. Instead, the way this is done in the
-    # InSAR geocoding workflow is to create a `GeocodeFloat32` object, resulting
-    # in conversion of the input mask values to float32 during geocoding (and
-    # then conversion back to uint8 when they're written to the output raster).
-    # For STATIC, this shouldn't cause any loss of fidelity since the mask values
-    # are small and we're using nearest neighbor interpolation.
+    # on uint8 data. Instead, the way this is done in the InSAR geocoding workflow is to
+    # create a `GeocodeFloat32` object, resulting in conversion of the input mask values
+    # to float32 during geocoding (and then conversion back to uint8 when they're
+    # written to the output uint8 raster). For the layover/shadow mask, this shouldn't
+    # cause any loss of fidelity since the mask values are small and we're using
+    # nearest neighbor interpolation.
 
-    # However, when geocoding uint8 data with `GeocodeFloat32`, the C++ `GeocodeCov`
-    # implementation initializes the output buffer with `nan_t_out`. Then,
-    # 0 * NaN evaluates to NaN for float dtypes (correct), but for integer
-    # dtypes it evaluates to 0 (since uint8 cannot represent NaN). This leads to
-    # areas outside the radar grid extent retaining this 0 fill value instead
-    # of the correct 255.
+    # However, when geocoding a `GeocodeFloat32`, the C++ `GeocodeCov` internally
+    # uses NaN as a fill value for the area outside the radar grid extent.
+    # uint8 cannot represent NaN, which leads to the geocoding fill area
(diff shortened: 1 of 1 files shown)


nemo794:
Tagging @xhuang-jpl and @seongsujeong , because this design pattern should be implemented for GUNW's connected components layer, too.

bhawkins:
Why not just set all pixels in the raster to 255 immediately after creation? The value 255 can be represented exactly in float32, so I don't see why conversions to/from uint8 are an issue.

nemo794:
> Why not just set all pixels in the raster to 255 immediately after creation? The value 255 can be represented exactly in float32, so I don't see why conversions to/from uint8 are an issue.

I'm not following. 🤔 "set all pixels in the raster to 255 immediately after creation" -> Which raster are you referring to? At what step in the processing?

python/packages/nisar/static/layover_shadow_mask.py
```diff
@@ -195,19 +195,43 @@ def geocode_layover_shadow_mask(
...
+    # We'll use this coverage information to mask the final geocoded layover
+    # shadow mask, setting its geocoding fill pixels' value to 255.
+
+    # Construct the output uint8 Geotiff raster
     geocoded_layover_shadow_mask = make_scratch_gtiff(
         shape=(geo_grid.length, geo_grid.width),
         dtype=np.uint8,
         dir_=scratch_dir,
         prefix="geocoded-layover-shadow-mask_",
     )
```
bhawkins, line 225:
```suggestion
     geocoded_layover_shadow_mask.fill(255)
```
This is what I mean. Presumably only the valid pixels get updated during processing, so if you initialize them all to 255 in the beginning then you don't have to muck around with temporary mask layers later.

I don't think we have a `isce3.io.Raster.fill` method yet, but it should be easy to implement. To test the approach without committing to that, you could start with
```Python
    geocoded_layover_shadow_mask.set_block(
        np.s_[:, :],
        np.full(geocoded_layover_shadow_mask.shape, 255, np.uint8))
```

jplzhan: @nemo794 Does this fix seem simpler? I can either run tests for your current commit or would you prefer to try Brian's suggestion instead?

nemo794: @bhawkins Thank you for clarifying! You mentioned the word "Presumably", and that unfortunately seems to be the gotcha. :/

That said, I did test your fix in hopes it would work. Unfortunately, it did not resolve the bug -- the geocoding fill area was populated with `0` values again.

I had Claude sleuth around ISCE3 to determine the reason, and here's its summary:

```
The `geocode.geocode()` call writes the entire `geoDataBlock` to the output raster, 
overwriting ALL pixels including the pre-filled 255s. The C++ code doesn't check 
"is this pixel already 255?" - it just blindly writes the entire block, here:
```
https://github.com/isce-framework/isce3/blob/bdf1f6fb9175ffe9e53b61839c797e69947dd268/cxx/isce3/geocode/GeocodeCov.cpp#L894-L895


**Detailed breakdown from Claude:**

The C++ Geocoding Process, from GeocodeCov.cpp lines 863-877:
```
// Initialize geoDataBlock with NaN (becomes 0 for uint8)
geoDataBlock.fill(nan_t_out);

_interpolate(...);  // <-- Writes ONLY to valid pixels

// Write the block to output raster
outputRaster.setBlock(geoDataBlock.data(), ...);
```

When you pre-fill the output raster with 255:
* Before geocode: `geocoded_layover_shadow_mask` = all 255s
* During geocode:
    - C++ creates internal `geoDataBlock` filled with `nan_t_out` (0 for uint8)
    - `_interpolate()` writes radar mask values (0-3) only to pixels inside radar grid
    - Pixels outside radar grid remain as 0 in geoDataBlock
* Writing to output: C++ calls `setBlock()` on the output raster
    - This overwrites the pre-filled 255s with the `geoDataBlock` contents
    - So outside radar grid: your 255s → 0 (the bug)

nemo794: A different design could maybe be to add a new `fill_value` parameter into `GeocodeCov`, so that internally it fills `geoDataBlock` with something other than `NaN`...?

jplzhan:
I ran tests for the current commit as is (`cd41d8e`) and placed the results here:
```bash
aws s3 ls s3://nisar-adt/zhan/layover_shadowmask_fix/
                           PRE 013_071/
                           PRE 015_056/
                           PRE 089_050/
                           PRE 107_123/
```

python/packages/nisar/static/layover_shadow_mask.py
bhawkins: Thanks. Yeah, it looks like `nan_t_out` ends up as zero in this case ([godbolt](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:45,endLineNumber:9,positionColumn:45,positionLineNumber:9,selectionStartColumn:45,selectionStartLineNumber:9,startColumn:45,startLineNumber:9),source:'%23include+%3Ccstdio%3E%0A%23include+%3Ccmath%3E%0A%23include+%3Ccstdint%3E%0A%0Aint+main()+%7B%0A++++using+T+%3D+uint8_t%3B%0A++++T+fill_value+%3D+0%3B%0A++++fill_value+*%3D+std::numeric_limits%3CT%3E::quiet_NaN()%3B%0A++++printf(%22fill_value+%3D+%25d%5Cn%22,+fill_value)%3B%0A++++return+0%3B%0A%7D'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:33.333333333333336,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:gsnapshot,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'0',intel:'0',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'-std%3Dc%2B%2B20+-O3+-ffast-math',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x86-64+gcc+(trunk)+(Editor+%231)',t:'0')),k:35.35095197499289,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+x86-64+gcc+(trunk)+(Compiler+%231)',t:'0')),k:31.31571469167377,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4)).

Just to recap my suggestion from today's meeting, I'd do it in two parts:
1) Add a new `fill_value` parameter to the function(s) that currently have a hard-coded `nan_t_out`, allowing a motivated user to control it.
2) Populate the default with the values we want for NISAR static layers. That might look something like the following:
```C++
#include <cstdint>
#include <limits>
#include <type_traits>

template<typename T>
constexpr T get_default_fill_value() {
    if constexpr (std::is_same_v<T, float>) {
        return std::numeric_limits<float>::quiet_NaN();
    } else if constexpr (std::is_same_v<T, double>) {
        return std::numeric_limits<double>::quiet_NaN();
    } else if constexpr (std::is_same_v<T, unsigned char>) {
        return static_cast<unsigned char>(0xff);
    } else if constexpr (std::is_same_v<T, unsigned short>) {
        return static_cast<unsigned short>(0xffff);  // ... or 0xff maybe?
    } else {
        static_assert(!sizeof(T), "get_default_fill_value: unsupported type");
    }
}

// prototype in header
template<typename T>
void geocode(
    // ... lots of parameters
    const T fill_value = get_default_fill_value<T>());  // fn call is okay because it's constexpr
```
That way we get the desired behavior right away and we can work on actually percolating `fill_value` up through the rest of the API as time permits.

In the meeting I was thinking about using `std::optional<T> fill_value = nullptr` which would let you choose the default with runtime logic. But you can just do it with `constexpr` logic, so that's not actually necessary and you can just stick it in the header.

gshiroma: The challenge with option 2 is that the value `255` (`0xff`) will be considered a valid value in the GeocodeCov with area projection. The algorithm accumulates the multilooked backscatter power `v` into the output geogrid pixel ([here](https://github.com/isce-framework/isce3/blob/bdf1f6fb9175ffe9e53b61839c797e69947dd268/cxx/isce3/geocode/GeocodeCov.cpp#L3505)). If the pixel has value `NaN` or `0`, the algorithm sets it to `v`. Otherwise, it adds `v` to the pixel value. If we initially set pixel to `255`, the current version of the algorithm will assume that `255` is a valid value and set the pixel to `255 + v`.

Option 1 would be preferable, but we need to be careful about how we implement it (considering the details above). Alternatively, we could use @nemo794 's implementation, as long as the additional runtime and memory usage required to compute `out_mask` are acceptable.

gshiroma: If we decide to implement option 1, we can probably reuse some of the code I have from PR #48.

nemo794: IIRC, Claude did not like Option 2 (having the code try to be smart and guess at fill values). Instead, Claude preferred implementing only Option 1 (add new fill value parameter to the function signature). Good to know that @gshiroma is in agreement on that path forward!

gshiroma: Thank you, @nemo794 , for checking it.

It turns out that the templating of `T_out` (the output data type) is done internally by the `GeocodeCov` module in this [function](https://github.com/isce-framework/isce3/blob/bdf1f6fb9175ffe9e53b61839c797e69947dd268/cxx/isce3/geocode/GeocodeCov.cpp#L127), and it only allows `float`, `double`, or complex types (of either float or double). The casting from float to int happens later, probably in `Raster.setBlock()`.

This means that we don't need to worry about handling integer values, which was the problem I raised above, and the implementation of option 1 (the new fill value parameter) should be much simpler than I initially anticipated. I should be able to open a PR with that implementation soon.

nemo794:
Update: @gshiroma has returned! After discussion with the team: @gshiroma will implement this fix from scratch and open a new PR, linking back to this PR to reference the discussion here.

In the new fix: @gshiroma will add a new parameter to the function signature called `invalid_value` (to mirror the parameter naming in the GPU version of geocode that InSAR uses). He'll then implement this for both the INTERP and RTC Area Correction implementations. If the RTC Area Projection implementation gets too complicated to be reviewed quickly, then we can do the following: Implement fully for INTERP, and if a user passes a non-NaN value and wants RTC Area Projection then the function will raise a "not implemented" exception.

Note that this change will need to be tested/reflected in InSAR for Connected Components layer. (For maintenance reasons, should probably check the code for GUNW and GOFF mask layer as well because they are also integer dtype, although I have not witnessed issues directly with the output.) @xhuang-jpl @jungkyoJung @seongsujeong

PR #346, opened 2026-07-31
Factorized backprojection

bhawkins:
This PR adds a factorized backprojection (FBP) algorithm for SAR focusing as a faster alternative to the existing direct backprojection method.

# Algorithm Design

The basic idea is from [Yegulalp 1999](https://doi.org/10.1109/NRC.1999.767270) where local, low-resolution polar grids can efficiently store partial sums of subapertures. As in other papers, the idea is extended hierarchically, so that multiple local grids can be combined together into a higher resolution grid. This process can be repeated in successive stages until the desired resolution is met. In the limit one achieves a factor of $O(N / log(N))$ reduction in the number of sums per output pixel (where $N$ is the number of pulses in the synthetic aperture). In practice there's a fair amount of overhead, but the speedup is still significant.

The FBP algorithm is usually described for the spotlight case, and the one here has a few modifications for the stripmap case relevant to NISAR. First, the azimuth extent of the local grid is extended based on the duration of the subaperture, since the beam is sliding. Second, the azimuth resolution of the local grid is not allowed to exceed the L/2 limit. The radar travel time between transmit and receive ("bistatic" correction) is accounted for both in the initial back projection stage and in the selection of grid coordinates. The latter is necessary to maintain a baseband signal in azimuth.

I'm not sure I'd call this a "fast factorized" (FFBP) algorithm. In the interest of maximal accuracy, it does full 2D interpolation and makes no assumptions about the trajectory or nesting of coordinate grids. Non-uniform FFT algorithms are used for interpolation in order to maintain high accuracy with small kernels. The zero-padded transforms do occupy a fair amount of memory, though, which may be a bottleneck. Care has been taken to ensure that only the "active" subimages reside in memory at any given time.

You might notice a few new functions that are not used in the NISAR RSLC workflow. These have been used for processing UAVSAR data from circular tracks (CSAR).

# Key Features

- Factorized backprojection algorithm: Multi-stage approach (initial → middle → final) that reduces computational complexity compared to direct backprojection
- 2D NFFT implementation: New CPU and GPU implementations of 2D Non-uniform FFT (NFFT2d) required by the FBP algorithm
- GPU acceleration: CUDA implementations for all FBP stages and NFFT2d operations
- Bug fixes in interp1d and interp2d
- Python interface: New azcomp_bp and azcomp_fbp functions exposed via isce3.focus.azcomp_bp module
- Configuration support: Added FBP parameters to focus workflow configuration schema

# Usage

The algorithm is highly configurable. You can specify the number of stages, the size of each stage, and the interpolation parameters for each stage (NFFT kernel size and oversampling ratio in both dimensions). To use a two-stage algorithm, the RSLC configuration file should look like
```YAML
runconfig:
    groups:
        processing:
            azcomp:
                factorization:
                - size: 128
```
which means sum groups of 128 pulses into polar grids before summing those into the output image. Similarly,
```YAML
# runconfig.groups.processing.azcomp
                factorization:
                - size: 64
                - size: 2
```
means a three-stage algorithm where the initial grids comprise 64 pulses each, these get merged together two at a time, and then the merged grids get summed into the output image. The default `factorization: [{"size": 1}]` is understood to mean the original direct backprojection algorithm.

In my testing I've found the two-stage n=128 algorithm is a 10x speedup of the azcomp algorithm on a g6e instance in AWS EC2.

I included a script that can generate animations like the following from the FBP debug file:

https://github.com/user-attachments/assets/293400f6-e5ef-47b2-a78e-4b9b7d025723


bhawkins:
Hmm the GPU build is failing, though it builds fine in my dev environment. I notice that I'm on CUDA 12 while the CI is building with CUDA 13. It looks like there was a change regarding visibility of `__global__` symbols (CUDA kernels) that might be relevant ([link](https://developer.nvidia.com/blog/cuda-c-compiler-updates-impacting-elf-visibility-and-linkage/)). It's a little tricky since I used a template to allow it to use Python containers without copying the contents, and I'd like to avoid instantiating Python junk in the C++ API. Will chew on this a bit more...

cxx/isce3/cuda/focus/Backproject.h
```diff
@@ -87,4 +87,158 @@ backproject(std::complex<float>* out,
...
+ * @param[out] height               Height of each pixel (m) above the
+ *                                  ellipsoid (optional, may be nullptr)
+ *
+ * @returns Non-zero error code if rdr2geo or geo2rdr fails to
+ *          converge for any pixel, and the values for these
+ *          pixels are set to NaN.
+ */
+template <class SequenceType>
+isce3::error::ErrorCode
+accumulatePolarImagesToRadarGrid(std::complex<float>* out,
```
bhawkins: Okay, CI passes now, but I'm not sure it's the best solution. The problem is basically downstream of this, where I template on `SequenceType` in the C++ API so that in the Python bindings I can directly pass in a Python sequence without having to copy the contents.

With CUDA 13, it seems like the options are
1. Add compiler flags to revert to the old behavior (current status).
2. Move basically all of Backproject.cu into the header (e.g., Backproject.icc) so that the instantiation in the Python bindings can compile in its own translation unit. 
3. Change the C++ API to use a concrete type instead of a template `SequenceType`. Then all the stuff in the .icc file could go into the .cu file instead, and the Python bindings wouldn't be calling CUDA kernels directly.

I think we might technically be okay with (1) but the issues are pretty subtle, and I might have misunderstood. The code seems to work, at least.

Approach (2) would basically have the compiler copy/paste 2k lines of code into any file that includes Backproject.h. Maybe that's okay.

I originally had something like `std::vector<NFFT2dResult<float>>` which required copying the images from the Python sequence into the `std::vector`, or maybe binding `std::vector` so you could construct it in Python. But if I made it a vector of pointers then the copy could be avoided while achieving (3). I'm kind of leaning in this direction.

bhawkins: Okay, option 3 fixed the issue, and I reverted the build script hacks.

bhawkins:
All the CI failures are in `test.cxx.isce3.io.raster.raster` with
```
[ RUN      ] RasterTest.addRasterToVRT
ERROR 3: Failed to read scanline 100.
ERROR 1: IReadBlock failed at X offset 0, Y offset 100: Failed to read scanline 100.
In isce3::io::Raster::getSetValue() - error in RasterIO.
/home/runner/work/isce3/isce3/tests/cxx/isce3/io/raster/raster.cpp:258: Failure
Expected equality of these values:
  std::isfinite(val)
    Which is: false
  true

[  FAILED  ] RasterTest.addRasterToVRT (55 ms)
```
Seems unrelated to the PR and the test passes locally.

python/packages/isce3/focus/serialization.py
```diff
@@ -0,0 +1,122 @@
...
+    @classmethod
+    def from_dict(cls, d: dict):
+        default = cls()
+        T = NonUniformFFTParameters
+        rg = T.from_dict(d["range"]) if "range" in d else default.range
+        az = T.from_dict(d["azimuth"]) if "azimuth" in d else default.azimuth
+        return cls(rg, az)
+
+@dataclass(frozen=True)
+class BackprojectionStageParameters:
```
bhawkins, line 217: Could use docstrings

python/packages/isce3/focus/azcomp_bp.py
```diff
@@ -0,0 +1,447 @@
...
+    igeom : isce3.container.RadarGeometry
+        Input radar geometry for range-compressed data.
+    rc_grid : RadarGridParameters
+        Grid parameters for range-compressed data.
+    rcdata : array-like
+        Range-compressed data, shape (azimuth, range).
+    ogrid : RadarGridParameters
+        Output zero-Doppler radar grid parameters.
+    writer : BackgroundWriter
+        Writer object for outputting focused data blocks.
```
bhawkins: Maybe document the writer better.

python/packages/isce3/focus/azcomp_bp.py
```diff
@@ -0,0 +1,447 @@
...
+        Desired azimuth resolution, in meters.
+    kernel : isce3.core.Kernel
+        Interpolation kernel for backprojection.
+    blocks_bounds : BlockPlan
+        List of tuples containing ((row_slice, col_slice), (t0, t1)) for each
+        processing block, where slices define the output grid region and (t0, t1)
+        are the required raw data time bounds in seconds.
+    igeom : isce3.container.RadarGeometry
+        Input radar geometry for range-compressed data.
+    rc_grid : RadarGridParameters
```
bhawkins: Is `rc_grid` redundant with `igeom`?

python/packages/isce3/focus/azcomp_bp.py
```diff
@@ -0,0 +1,447 @@
...
```
bhawkins: Ditto here. Redundant args?

python/packages/isce3/focus/azcomp_bp.py
```diff
@@ -0,0 +1,447 @@
...
+        merge_polar_images = isce3.focus.merge_polar_images
+        add_to_radar_grid = isce3.focus.accumulate_polar_images_to_radar_grid
+        make_image_nfft2d = isce3.signal.make_image_nfft2d
+
+    _, v = igeom.orbit.interpolate(igeom.orbit.mid_time)
+    vs = np.linalg.norm(v)
+    tq_max = isce3.focus.get_polar_angle_time_constant(fc, vs, bandwidth)
+
+    if debugfile is not None:
+        import h5py
```
bhawkins: Don't see why we'd delay this import.

bhawkins (commented):
Comments from today's review meeting.

bhawkins:
Fix for the CI failure is in #348

bhawkins:
I just pushed a change making FBP the default algorithm in order to simplify PCM/SDS testing. @hfattahi @jshimada47

jshimada47:
I'm getting a unit test failure while doing the build on dev-5 in `test.python.pkg.nisar.workflows.focus`:
```
253/280 Test #248: test.python.pkg.nisar.workflows.focus ...............................***Failed   23.63 sec
============================= test session starts ==============================
platform linux -- Python 3.12.11, pytest-8.4.1, pluggy-1.5.0
rootdir: /src
configfile: pyproject.toml
collected 3 items

../../../../../../src/tests/python/packages/nisar/workflows/focus.py .F. [100%]

=================================== FAILURES ===================================
__________________________ test_focus[factorization1] __________________________

factorization = [<nisar.workflows.focus.Struct object at 0x7eff92c080b0>]

    @pytest.mark.parametrize("factorization", (direct_bp, factorized_bp_64))
    def test_focus(factorization):
        cfg = get_test_cfg()
        cfg.runconfig.groups.processing.azcomp.factorization = factorization

>       focus.focus(cfg)

/src/tests/python/packages/nisar/workflows/focus.py:45:
/bld/install/opt/conda/lib/python3.12/site-packages/nisar/workflows/focus.py:1873: in focus
    slc = SLC(output_slc_path, mode="w", product=product,
/bld/install/opt/conda/lib/python3.12/site-packages/nisar/products/writers/SLC.py:353: in __init__
    super().__init__(*args, **kw)
/opt/conda/lib/python3.12/site-packages/h5py/_hl/files.py:564: in __init__
    fid = make_fid(name, mode, userblock_size, fapl, fcpl, swmr=swmr)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/opt/conda/lib/python3.12/site-packages/h5py/_hl/files.py:244: in make_fid
    fid = h5f.create(name, h5f.ACC_TRUNC, fapl=fapl, fcpl=fcpl)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
h5py/_objects.pyx:56: in h5py._objects.with_phil.wrapper
    ???
h5py/_objects.pyx:57: in h5py._objects.with_phil.wrapper
    ???

>   ???
E   OSError: Unable to synchronously create file (unable to truncate a file which is already open)

h5py/h5f.pyx:122: OSError
------------------------------ Captured log call -------------------------------
WARNING  focus:focus.py:331 No DEM given, using ref height 0.0 (m).
WARNING  focus:focus.py:441 Desired Doppler LUT start time is 1.0 seconds before ephemeris start. Consider adjusting ephemeris_crop_pad or providing more orbit/attitude data.
WARNING  focus:focus.py:445 Desired Doppler LUT end time is 1.0 seconds after ephemeris end. Consider adjusting ephemeris_crop_pad or providing more orbit/attitude data.
=============================== warnings summary ===============================
tests/python/packages/nisar/workflows/focus.py::test_focus[factorization0]
tests/python/packages/nisar/workflows/focus.py::test_focus[factorization1]
  /bld/install/opt/conda/lib/python3.12/site-packages/nisar/products/readers/Base/Identification.py:238: UserWarning: Could not find isJointObservation in product identification.
    warn("Could not find isJointObservation in product identification.")

tests/python/packages/nisar/workflows/focus.py::test_focus[factorization0]
tests/python/packages/nisar/workflows/focus.py::test_focus[factorization1]
  /bld/install/opt/conda/lib/python3.12/site-packages/nisar/products/readers/Base/Identification.py:247: UserWarning: Could not find hasInputDataException in product identification.  Assuming there are no anomalies.
    warn("Could not find hasInputDataException in product "

tests/python/packages/nisar/workflows/focus.py::test_focus[factorization0]
  /bld/install/opt/conda/lib/python3.12/site-packages/nisar/products/readers/Raw/Raw.py:466: UserWarning: Missing dataset /science/LSAR/RRSD/highRateTelemetry/txH/rxH/QFSP0/CHIRP_CORRELATOR_I1_H01 in /src/tests/data/focus/REE_L0B_out17.h5. Detailed error -> 'Unable to synchronously open object (component not found)'
    warn(

tests/python/packages/nisar/workflows/focus.py::test_focus[factorization0]
  /bld/install/opt/conda/lib/python3.12/site-packages/nisar/antenna/rx_channel_imbalance_helpers.py:212: UserWarning: No LNA CAL to represent RX! Use BYPASS Cal instead!
    warn('No LNA CAL to represent RX! Use BYPASS Cal instead!')

tests/python/packages/nisar/workflows/focus.py::test_focus[factorization0]
  /bld/install/opt/conda/lib/python3.12/site-packages/nisar/antenna/rx_channel_imbalance_helpers.py:217: UserWarning: No LNA or BYPASS CAL! LNA mean will be all unity. The results will be invalid!
    warn('No LNA or BYPASS CAL! LNA mean will be all unity. '

tests/python/packages/nisar/workflows/focus.py::test_focus[factorization0]
  /bld/install/opt/conda/lib/python3.12/site-packages/nisar/antenna/rx_channel_imbalance_helpers.py:496: UserWarning: All values are zero for HH-pol Caltone! They are set to untiy. Result may be invalid!
    warn(f'All values are zero for {msg}! They are set to untiy. '

tests/python/packages/nisar/workflows/focus.py::test_focus[factorization0]
  /bld/install/opt/conda/lib/python3.12/site-packages/nisar/antenna/rx_channel_imbalance_helpers.py:500: UserWarning: Some values are zero for HH-pol Caltone!
    warn(f'Some values are zero for {msg}!')

tests/python/packages/nisar/workflows/focus.py::test_focus[factorization0]
  /bld/install/opt/conda/lib/python3.12/site-packages/nisar/antenna/rx_channel_imbalance_helpers.py:48: UserWarning: The size of LNA-CALTONE ratio is 1 instead of 12!
    warn('The size of LNA-CALTONE ratio is '

tests/python/packages/nisar/workflows/focus.py::test_focus[factorization0]
  /bld/install/opt/conda/lib/python3.12/site-packages/nisar/products/readers/instrument/instrument_parser.py:347: MissingInstrumentFieldWarning: HPOL/channelAdjustment/
    return self._channel_adjustment_factors('rx', pol)

tests/python/packages/nisar/workflows/focus.py::test_focus[factorization0]
  /bld/install/opt/conda/lib/python3.12/site-packages/nisar/products/readers/instrument/instrument_parser.py:310: MissingInstrumentFieldWarning: HPOL/channelAdjustment/
    return self._channel_adjustment_factors('tx', pol)

tests/python/packages/nisar/workflows/focus.py::test_schema
  <unknown>:1: SyntaxWarning: invalid escape sequence '\d'

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
==================================== PASSES ====================================
__________________________ test_focus[factorization0] __________________________
----------------------------- Captured stdout call -----------------------------
{
  "magnitude": 0.00012279492511879653,
  "phase": -3.09582781791687,
  "azimuth": {
    "ISLR": -15.777904987335205,
    "PSLR": -17.766033411026,
    "resolution": 1.203125,
    "index": 64.0,
    "offset": 0.0007591966132167727,
    "phase ramp": -0.006026060786098242
  },
"range": {
    "ISLR": -13.799757957458496,
    "PSLR": -16.547774076461792,
    "resolution": 1.15625,
    "index": 64.0,
    "offset": 4.2870595962085645e-10,
    "phase ramp": -0.015377550385892391
  }
}
------------------------------ Captured log call -------------------------------
WARNING  focus:focus.py:331 No DEM given, using ref height 0.0 (m).
WARNING  focus:focus.py:441 Desired Doppler LUT start time is 1.0 seconds before ephemeris start. Consider adjusting ephemeris_crop_pad or providing more orbit/attitude data.
WARNING  focus:focus.py:445 Desired Doppler LUT end time is 1.0 seconds after ephemeris end. Consider adjusting ephemeris_crop_pad or providing more orbit/attitude data.
WARNING  SLCWriter:SLC.py:333 Could not determine coverage indicator given frame_polygon=<POLYGON EMPTY> and image_polygon=<POLYGON Z ((-54.584 3.173 0, -54.583 3.174 -7.081e-10, -54.582 3.174 0, -54...>
WARNING  SLCWriter:SLC.py:771 Input L0B has undesired 'units' attribute on diagnosticModeFlag dataset.  Will omit from output RSLC.
WARNING  focus:focus.py:2188 No noise-only range lines within the specified pulse interval. Skip noise estimation and set noise equivalent backscatter to zero.
=========================== short test summary info ============================
FAILED ../../../../../../src/tests/python/packages/nisar/workflows/focus.py::test_focus[factorization1]
PASSED ../../../../../../src/tests/python/packages/nisar/workflows/focus.py::test_focus[factorization0]
PASSED ../../../../../../src/tests/python/packages/nisar/workflows/focus.py::test_schema
================== 1 failed, 2 passed, 13 warnings in 23.14s ===================
journal: hasInputDataException not found at /science/LSAR/identification
journal: hasInputDataException not found at /science/LSAR/identification
journal: No reference epoch provided. Using first date time from XML file as orbit reference epoch.
journal: hasInputDataException not found at /science/LSAR/identification
journal: hasInputDataException not found at /science/LSAR/identification
journal: hasInputDataException not found at /science/LSAR/identification
journal: hasInputDataException not found at /science/LSAR/identification
journal: hasInputDataException not found at /science/LSAR/identification
journal: hasInputDataException not found at /science/LSAR/identification
journal: cube height: 20
journal:
 -- cube length: 12
 -- cube width: 12
journal: EPSG: 32621
journal: estimating the ground-track velocity using rdr2geo
journal: hasInputDataException not found at /science/LSAR/identification
journal: hasInputDataException not found at /science/LSAR/identification
journal: No reference epoch provided. Using first date time from XML file as orbit reference epoch.
journal: hasInputDataException not found at /science/LSAR/identification
journal: hasInputDataException not found at /science/LSAR/identification
journal: hasInputDataException not found at /science/LSAR/identification
journal: hasInputDataException not found at /science/LSAR/identification
journal: hasInputDataException not found at /science/LSAR/identification
journal: hasInputDataException not found at /science/LSAR/identification

        Start 249: test.python.pkg.nisar.workflows.gen_doppler_range_product
254/280 Test #249: test.python.pkg.nisar.workflows.gen_doppler_range_product ...........   Passed    3.58 sec
```

python/extensions/pybind_isce3/cuda/focus/Backproject.cu
```diff
@@ -0,0 +1,351 @@
...
+
+            // Convert Python sequence to std::vector.  Element type is
+            // pointer so this shouldn't involve any copy.
+            using T = isce3::cuda::signal::NFFT2dResult<float>;
+            auto interpolators = std::vector<const T*>(nimg);
+            std::transform(py_image_interpolators.begin(),
+                py_image_interpolators.end(), interpolators.begin(),
+                [](const py::handle& py_itp) -> const T* {
+                    return &(py_itp.cast<const T&>());
+                });
```
Tyler-g-hudson, line 296:
```suggestion
            // Keep Python objects alive - pointers remain valid until py_objects destroyed
            std::vector<py::object> py_objects;
            std::vector<const T*> interpolators;
            py_objects.reserve(nimg);      // Pre-allocate to prevent reallocation
            interpolators.reserve(nimg);

            // Store each object (increments refcount), then take pointer to stored object
            for (const auto& py_itp : py_image_interpolators) {
                py_objects.push_back(py::cast<py::object>(py_itp));
                interpolators.push_back(&(py_objects.back().cast<const T&>()));
            }
```
Small fragility that could result in a failure in later developments - the objects indicated by the pointers in `interpolators` could be prematurely destroyed if the GIL is released (e.g. someone comes along later and uses `py::gil_scoped_release release;` before the `mergePolarImages` call) and they go out of scope.

I don't think this is a bug now, but it might be good defensive coding to just store these objects in the Backproject.cu and Backproject.cpp pybinds anyway to make sure they're safe in future development

share/nisar/defaults/focus.yaml
```diff
@@ -446,6 +446,11 @@ runconfig:
                     fit: Table # null or Cheby or Table
                     fit_order: 2048
 
+                # List of aperture factorization parameters.
+                # Default to two-stage with 128 pulses per subaperture.
+                factorization:
+                - size: 128
```
Tyler-g-hudson, line 452: To clarify, do we intend to modify this list to be hierarchical later on for more efficient processing or no?

Also, for a typical input L0B ~40 seconds per RSLC containing ~76k pulses, wouldn't 256 be the nearest power of 2 to the optimum per Yegulalp? Have you had a chance to check this case? Are my numbers wrong here? (I'm basing them off of a single full-length RSLC I have, and assuming a 1.5 second pad before and after in the L0B at the L0B native PRF)

Do we need additional scaled testing for hierarchical runs? I could produce some scaled workflows on a g6e if you have one or more RSLCs you want me to reproduce and a handful of factorization lists.

Tyler-g-hudson (commented):
I'm slowly chewing my way through the PR, here are a very basic review item and question

python/packages/isce3/focus/azcomp_bp.py
```diff
@@ -0,0 +1,447 @@
...
+            igeom.doppler.save_to_h5(h5.require_group("doppler"), "doppler",
+                epoch, "Hz")
+            h5.create_dataset("epoch", data=np.bytes_(epoch))
+            h5.create_dataset("wavelength", data=igeom.radar_grid.wavelength)
+
+    # Focus to intermediate grids.
+    # NOTE We'll actually just define the tasks and only evaluate them as needed
+    # in order to reduce memory pressure.  We could process them in
+    # parallel using concurrent.futures or dask, for for now just store them in
+    # a dict keyed by the PolarGrid associated with the imagelets.
```
Tyler-g-hudson, line 296: Would Python-level parallelism even help? The individual tasks are parallel at the C++/OMP level so, at least for CPU, they should saturate the available computation resources, no?

This would probably be more complicated to figure out for GPU, but you'd need to see if the GPU usage is low enough at any step of the task to justify parallel task requests and associated complexity. Maybe worth a test later.

python/packages/isce3/focus/azcomp_bp.py
```diff
@@ -0,0 +1,447 @@
...
+    def process_pulses(iblock, nblocks, debugfile, fdata, sr, x, v,
+                polar_grid, dem, fc, kernel, atmos, rdr2geo_params):
+        log.info(f"Focusing {len(x)} pulses to polar image {iblock + 1} of {nblocks}")
+        # NOTE Atmosphere will get applied (if requested) at final stage to
+        # avoid phase modulation from DEM sampling issues across subimages.
+        # Always "nodelay" in this stage.
+        _, img, _ = bp_to_polar_grid(fdata, sr, x, v,
+            polar_grid, dem, fc, kernel, "nodelay", rdr2geo_params)
+        if debugfile is not None:
+            import h5py
```
Tyler-g-hudson, line 315: Recommend moving this to the top of the file, I don't think this is a heavy dependency relative to ISCE3 and should only be performed once as the file is uploaded if it has not already been imported (and it likely has)

python/extensions/pybind_isce3/cuda/focus/Backproject.cu
bhawkins: Good suggestion, I'll try that out.

share/nisar/defaults/focus.yaml
bhawkins: Recall that we're using stripmap mode (not spotlight), so there are only around 6000 pulses in a NISAR synthentic aperture. Then Yegulalp's `sqrt(num_aperture)` suggestion for a two-stage algorithm would work out to 77 pulses.

However, in the version here (at least in the default config) the first stage uses a 9-point 1D interpolator while the second (and higher) stage(s) use a 5x5=25 point 2D interpolator. So if interpolation dominates the runtime, then later stages are around f=2.7 times more expensive than the first stage. To split the work evenly among a two-stage algorithm, then you'd want to solve `num_aperture = (f * n) * n` which results in around `f * n = 129` pulses in the first stage. In the general case you'd want to solve
```Python
num_aperture = f**(num_stages - 1) * n**num_stages
```
for `n`. So for a NISAR 3-stage algorithm you might try the config
```YAML
factorization:
- size: 70
- size: 9
```
However, there are other considerations like the geometry calculations (including the fact that the geometry is in float64 while the interpolation is float32), the FFT overhead, GPU memory bandwidth, cache sizes, and so-on that complicate the picture. I measured runtimes for two-stage cases using size in {64, 128, 256} and didn't see very large differences between them on g6/g6e instances. It might be worth repeating that experiment since I've hacked on the code a fair amount since then.

python/extensions/pybind_isce3/cuda/focus/Backproject.cu
bhawkins: I fixed it here and also in merge_polar_images. In the latter I was able to add a GIL release, which is a nice side benefit.

cxx/isce3/signal/NFFT2d.h
```diff
@@ -0,0 +1,156 @@
...
+    private:
+        dims_t m_, sizes_, fft_sizes_;
+        std::vector<std::complex<T>> xf_;
+        std::array<std::vector<T>, 2> weights_;
+        std::array<isce3::core::NFFTKernel<T>, 2> kernels_;
+};
+
+
+template <typename T>
+class NFFT2dResult {
```
Tyler-g-hudson, line 66: Readability suggestion:
```suggestion
/**
 * Prepared interpolator for 2D image queries at arbitrary positions.
 * Holds oversampled grid enabling rapid Kaiser-Bessel interpolation
 * queries.
 */
template <typename T>
class KaiserBesselInterpolator2d {
```
I think this could use a rename and/or or some nearby documentation. My reason for suggesting this is that the data it holds is in the polar domain (Seeing NFFT in the name might make a reader mistake at a glance that this as some sort of frequency spectrum) and that, in principle, an oversampler of any kind could produce one of these interpolators (presumably in any domain?) so long as it can also produce the correct kernels.

In an ideal world I'd recommend moving this to a separate file and renaming it - failing those, a quick comment here would probably help future maintenance.

I'm interested in this method in general - do you think it's practical for InSAR resampling? Do you know any pros and cons to doing so off the top of your head?

cxx/isce3/signal/NFFT2d.h
```diff
@@ -0,0 +1,156 @@
...
+ * @param s         Minimum factors (> 1) for frequency-domain zero-padding
+ *                  along {rows, columns}.  Actual padding may be larger to
+ *                  achieve efficient inverse transform size.
+ * @param pad_input Whether to also zero-pad input data to an efficient
+ *                  forward transform size.  Requires extra memory.
+ *
+ * @return NFFT2d<T> object for interpolating the image.
+ */
+template<typename T>
+NFFT2dResult<T> makeImageNFFT2d(
```
Tyler-g-hudson, line 151: Same comment as above - this no longer returns an NFFT2d so I recommend a rename for it. Docstring parameters also look to have been changed.
```suggestion
/**
 * @brief Create an rapid interpolator using an intermediate NFFT2d.
 *
 * @tparam T        Format of real/imag pixel data, typically float or double
 * @param image     Input time-domain image.  A temporary copy will be made if
 *                  it is not row-major with a column stride of one.
 * @param params    parameters for the intermediate NFFT
 * @param pad_input Whether to also zero-pad input data to an efficient
 *                  forward transform size.  Requires extra memory.
 *
 * @return KaiserBesselInterpolator2d<T> object for interpolating the image.
 */
template<typename T>
KaiserBesselInterpolator2d<T> makeFastImageInterpolator(
```

Tyler-g-hudson (commented):
Some questions and naming/documentation recommendations

PR #347, opened 2026-08-05
Switch to MVE Noise Est For Single Noise-only RangeLine in RSLC (NSDS-4938)

rad-eng-59:
Handful of mixed-mode RSLC workflow fails as a result of not-enough noise-only range lines (simply one range line) in **MEE**-based noise estimator when either the requested output grid simply uses less than around two seconds of a L0B product or the respective L0B is too short (`< 2 seconds`). 
This quick PR simply switches the method to **MVE** and issues a warning rather than raising exception (current behavior).
The issue has been reported in [this jira ticket](https://jira.jpl.nasa.gov/browse/NSDS-4938).

As an example, this has been tested for a mixed-mode with three L0Bs that failed in previous production. 
Here is the respective **HH** **NESZ** plot for three AZ blocks of science band (aka **A**). This is a fixed PRF with three TX gaps where a large portion of the frame covers the water. 
plot_nesz_mixed_mode_three_l0b


bhawkins (approved):
LGTM. Normally I prefer not to modify user inputs, but in this case it seems reasonable enough.

hfattahi:
@rad-eng-59 Similar to @bhawkins I also don't want us to change the user configuration and would like to push the user to fix their configuration. However, such approach will significantly complicate automated system for little gain. Therefore I'm ok with this change. 

Would you please make a comment [here in the schema](https://github.com/isce-framework/isce3/blob/develop/share/nisar/schemas/focus.yaml#L372) and clarify that that if user chooses MEE and number of receive-only rangelines are less than 2, then the workflow switches back to MVE. Thanks!

hfattahi (approved):
LGTM!

PR #348, opened 2026-08-07
Fix failing unit test test.cxx.isce3.io.raster.raster

bhawkins:
This PR will attempt to fix the failing unit test `test.cxx.isce3.io.raster.raster`. I noticed it on another PR, but I think it's unrelated to those changes. I think I know what the problem is but want to reproduce the CI failure before pushing a fix.


bhawkins:
Okay, looks like I found it. The unit test was only initializing the first 100 rows of a raster and then trying to read 200 rows in another test. I guess some update in GDAL now triggers a failure in that case. It's easy to fix the tests by just initializing the full raster. I added a few extra test assertions for good measure.

gshiroma (approved):
LGTM.

Tyler-g-hudson (approved):
LGTM

PR #349, opened 2026-08-11
Update Soil Moisture SAS to v0.4.3

xhuang-jpl:
This PR is to update the soil moisture SAS version to v0.4.3, which includes the following change:

1. Update the PMI algorithm to include the backscatter min/max value
2. Update the DSG algorithm to take into account the water over the rice fields.
3. Bump the SAS version to v0.4.3 and product spec to 1.0.3.


jshimada47 (approved):
LGTM, thanks

hfattahi (approved):

hfattahi:
Approving because this is a test delivery. For next formal delivery we will address the CI failures. All unit tests have been passed with an internal docker recipe which pins different libraries.

PR #352, opened 2026-08-12
Add  the `fill_value` parameter to `GeocodeCov` CPU module

gshiroma:
This PR adds the `fill_value` parameter to the `GeocodeCov` CPU module. It is inspired by #337 and sets the fill value of the layover/shadow mask layer in the Static Layers (STATIC) product to `255`, instead of the default value of `0`. Details about the issue and the discussion can be found in that PR.

The C++ and Python interfaces have been updated with the new parameter. The two CPU geocoding algorithms, interpolation and area projection, have also been updated to account for the new parameter. For the interpolation algorithm, the array is first filled with `fill_value` before interpolation occurs. In contrast, for the area projection algorithm, `fill_value` is applied after multilooking. This is necessary because the area projection algorithm accumulates averaged values (multilooked backscatter), and applying `fill_value` before data accumulation could cause it to be interpreted as a valid value.


cxx/isce3/geocode/GeocodeCov.h
```diff
@@ -225,6 +229,11 @@ class Geocode {
      * `abs_cal_factor parameters`, which default to `false` and `1`,
      * respectively.
      * @param[in]  dem_raster          Input DEM raster
+     * @param[in]  fill_value          Fill value. Defaults to NaN. If the
+     * output is complex (e.g., off-diagonal terms), this value is used for
+     * the real part with the imaginary part set to 0
+     * @param[in]  flag_apply_rtc      Apply radiometric terrain correction
```
nemo794: I thought this PR was only adding the `fill_value` parameter.

Was `flag_apply_rtc` parameter accidentally included?

cxx/isce3/geocode/GeocodeCov.h
```diff
@@ -225,6 +229,11 @@ class Geocode {
      * `abs_cal_factor parameters`, which default to `false` and `1`,
      * respectively.
      * @param[in]  dem_raster          Input DEM raster
+     * @param[in]  fill_value          Fill value. Defaults to NaN. If the
+     * output is complex (e.g., off-diagonal terms), this value is used for
+     * the real part with the imaginary part set to 0
```
nemo794: Please update this comment to be explicit about the behavior of this new parameter for different cases. 

Some Cases:
* INTERP vs. RTC AREA PROJECTION
   - For INTERP, the fill value is only applied to geocoding fill area
   - For AREA PROJECTION, the fill value is applied to the geocoding fill area AND also any `NaN` values inside the input raster itself (Please confirm this is what is happening?)
   - Issue: What happens for a float-valued input raster that contains NaNs, but the fill value is 255?
* float-valued output, fill value of 255
* integer-valued output, fill value of NaN (default)
    - If I understand correctly, the behavior is undefined so please recommend to user to provide a fill value of e.g. `255`.
* Complex-valued output, but Fill Value is NaN (a real-valued float).
   - Does it get expanded to Nan+Nan*J? or is it also Nan+0*j?
   - The previous behavior (and per PIX specs), this should be Nan+Nan*J
* Complex-valued output, but Fill Value is non-NaN real-valued float, e.g. 255
   - Currently, the docstring says this will be expanded to e.g. 255+0*j
   - Ok by me to keep this behavior (there is no use case in NISAR for complex-valued output with a non-NaN fill value. It's future work.)

gshiroma:
We just met to review this PR, and it was decided that we should add a warning message to alert the user when casting the data or fill value to the output raster type changes the data or fill value. In addition, the docstring for `fill_value` needs to be updated to explain these changes. The new feature should also be covered by a unit test.

gshiroma:
> We just met to review this PR, and it was decided that we should add a warning message to alert the user when casting the data or fill value to the output raster type changes the data or fill value. In addition, the docstring for `fill_value` needs to be updated to explain these changes. The new feature should also be covered by a unit test.

@hfattahi , @bhawkins , and @nemo794 . I implemented all the suggestions from our PR review meeting. I can’t pinpoint specific commits because the changes are spread across several commits. The changes are:
* Added a unit test.
* Added warning messages when data or fill values are cast while being stored in the output raster files.
* Replaced the invalid complex fill value `NaN + 0j` with `NaN + NaN.j`.

Would you mind taking another look at this PR?

cxx/isce3/geocode/GeocodeCov.cpp
```diff
@@ -157,9 +179,48 @@ void Geocode<T>::geocode(const isce3::product::RadarGridParameters& radar_grid,
...
+                << pyre::journal::endl;
+    }
+
+    // Print a warning if output_raster has a complex data type.
+    if (GDALDataTypeIsComplex(outputRasterDataType)) {
+        warning << "The output raster has a complex data type, "
+                << "whereas the variable `fill_value` has a double data type. "
+                << "The real-valued fill value will be cast by "
+                << "isce3::io::Raster with a 0 imaginary part as "
+                << "<REAL_PART> + 0j."
```
bhawkins: Might want to mention special case of `NaN` here.

tests/python/extensions/pybind/geocode/geocodeCov.py
```diff
@@ -123,26 +115,92 @@ def test_geocode_cov():
...
+    sub_swath_str = ''
+
+    for geocode_algorithm in geocode_modes.keys():
+        for dtype in ['cfloat64', 'float64', 'byte', 'uint16']:
+            test_raster = \
+                f"{axis}_{geocode_algorithm}{sub_swath_str}_{dtype}.geo"
+            print(f'    file: {test_raster}')
+
+            ds = gdal.Open(test_raster, gdal.GA_ReadOnly)
+            data_fill_value = ds.GetRasterBand(1).ReadAsArray(0, 0, 1, 1)[0, 0]
```
bhawkins, line 183:
```suggestion
            # we know that first pixel is fill_value
            data_fill_value = ds.GetRasterBand(1).ReadAsArray(0, 0, 1, 1)[0, 0]
```
clarify for posterity

bhawkins (approved):
LGTM

cxx/isce3/geocode/GeocodeCov.cpp
```diff
@@ -157,9 +179,48 @@ void Geocode<T>::geocode(const isce3::product::RadarGridParameters& radar_grid,
...
+    pyre::journal::warning_t warning("isce.geocode.GeocodeCov.geocode");
+    GDALDataType outputRasterDataType = output_raster.dtype();
+
+    // Print a warning if output_raster has an integer data type.
+    if (GDALDataTypeIsInteger(outputRasterDataType)) {
+        warning << "The output raster has an integer data type, "
+                << "which is not natively supported by the GeocodeCov "
+                << "module. The data values and fill value will therefore "
+                << "be cast to an integer type by isce3::io::Raster. "
+                << "NaN values will be cast to 0."
```
hfattahi, line 192: Can you clarify what do you mean exactly here? Do you intend to provide a warning that the geocoding is done by casting integer to float and then the final geocoded value will be cast to integer by isce3::io::Raster, during writing the output?

hfattahi (commented):
looks good to me overall. Good functionality added. I have one minor question below

cxx/isce3/geocode/GeocodeCov.cpp
gshiroma: That's correct, @hfattahi . @bhawkins asked that the user be informed about this casting. This is the same type of casting that caused the “problem” in the STATIC workflow, where the `NaN` values were silently cast to `0`.

gshiroma:
@bhawkins , just letting you know that I'm setting the fill value for the GCOV & GSLC `inputDataExceptionMask` to `65535` in the commit https://github.com/isce-framework/isce3/pull/352/changes/c2e7f6e094a337d7cac6312d525de20cf8ada2fc (instead of issuing a new PR).

hfattahi (approved):
LGTM. Thanks @gshiroma

PR #357, opened 2026-08-13
Limit water mask coverage to radar grid extents

gshiroma:
This PR follows #356 and ensures that the water mask layer has consistent coverage with the other static layers, limited to the extents of the radar grid (`RadarGridParameters`). This is achieved by using the layover/shadow mask as a “reference fill value layer”. The geocoding of this mask assigns the fill value to areas outside the radar grid, which is then used to identify the corresponding fill values in the water mask layer.


tests/python/packages/nisar/static/water_mask.py
```diff
@@ -18,15 +18,27 @@ def test_binarize_nisar_water_mask():
...
+    # Add fill values over non water
+    reference_fill_value[:, 5:10] = 255
+
+    # Add fill values over water
+    reference_fill_value[:, 15:20] = 255
+
     # Convert the water distance map into a binary water mask
-    water_mask = binarize_nisar_water_mask(water_distance)
+    water_mask = binarize_nisar_water_mask(
+        water_distance, reference_fill_value=reference_fill_value)
```
Tyler-g-hudson:
```suggestion
        water_distance, reference_fill_value=reference_fill_value
    )
```

python/packages/nisar/static/water_mask.py
```diff
@@ -37,14 +41,15 @@ def binarize_nisar_water_mask(water_distance: ArrayLike) -> np.ndarray:
...
+    water = ((water_distance >= 1) & (water_distance <= 200)).astype(np.uint8)
 
-    # Get a binary mask of invalid pixels (i.e. pixels whose value is equal to the fill
-    # value of 255).
+    # Set pixels to the fill value if either the water distance or the
+    # reference image contains a fill value.
     fill_value = 255
-    invalid = water_distance == fill_value
+    water[((water_distance == fill_value) |
+           (reference_fill_value == fill_value))] = fill_value
```
Tyler-g-hudson:
```suggestion
    water[
        ((water_distance == fill_value) | (reference_fill_value == fill_value))
    ] = fill_value
```

Tyler-g-hudson (commented):
Some nitpicks;

I'm mainly a little confused about the difference between what this PR does and what the name suggests; The PR seems to mask the layover shadow mask onto the watermask, but the title makes claims about using the extents of the RadarGridParameters which don't seem to be used in the code changes.

python/packages/nisar/static/water_mask.py
gshiroma: Thank you, @Tyler-g-hudson , I think these are style changes, right? I tend to limit the number of columns to 80.

Tyler-g-hudson: These are style changes, yes - the change I recommended here has a max line length of 79 characters.

Tyler-g-hudson:
Please also rebase to develop - Brian just pushed a branch that will get rid of these CI failures

gshiroma:
> This is achieved by using the layover/shadow mask as a “reference fill value layer”. The geocoding of this mask assigns the fill value to areas outside the radar grid, which is then used to identify the corresponding fill values in the water mask layer.

Thank you, @Tyler-g-hudson ,  I updated the PR description to make the intent of this PR clearer:

“This is achieved by using the layover/shadow mask as a ‘reference fill value layer.’ The geocoding of this mask assigns the fill value to areas outside the radar grid, which is then used to identify the corresponding fill values in the water mask layer.”

Tyler-g-hudson:
> “This is achieved by using the layover/shadow mask as a ‘reference fill value layer.’ The geocoding of this mask assigns the fill value to areas outside the radar grid, which is then used to identify the corresponding fill values in the water mask layer.”

Thanks for this clarification - I think it would help to have a brief comment about this in the code near where the layover/shadow mask is passed into the function to explain why it is there.

python/packages/nisar/static/water_mask.py
gshiroma: I see.. so you want me to use your style? lol

gshiroma: Just accepted your suggestions.

gshiroma:
> > “This is achieved by using the layover/shadow mask as a ‘reference fill value layer.’ The geocoding of this mask assigns the fill value to areas outside the radar grid, which is then used to identify the corresponding fill values in the water mask layer.”
> 
> Thanks for this clarification - I think it would help to have a brief comment about this in the code near where the layover/shadow mask is passed into the function to explain why it is there.

Thank you, @Tyler-g-hudson , I added this comment in https://github.com/isce-framework/isce3/pull/357/commits/79d3a2fac03ae9d01f3fcb98777cd9bec44c83cc.

Issue #354, opened 2026-08-13
Performance: vectorizing generate_insar_mask() makes prepare_insar_hdf5 5x faster


s-sasaki-earthsea-wizard:
## Summary

[`generate_insar_mask()`](https://github.com/isce-framework/isce3/blob/0a8df45ddf780d22e41bf5358502121e6f09b7df/python/packages/nisar/products/insar/utils.py#L497) builds the InSAR `mask` layer with a pure-Python double loop over every output pixel — two scalar pybind `SubSwaths.get_sample_sub_swath` calls, NumPy scalar indexing and a `list.append` per pixel, ~5 µs/px. The per-pixel work is elementwise with no loop-carried dependency, so the same computation vectorizes with NumPy using only the existing `SubSwaths.get_valid_samples_array` bulk API. On a frame-scale grid (72.4 Mpx) the function drops from ~383 s to ~7 s on purely synthetic data (**57x**; self-contained reproducer below), and the containing `prepare_insar_hdf5` stage runs **5x faster with bitwise-identical output** on a real production run. The rewrite also simplifies the function body (guard clauses and a named membership helper instead of five levels of nesting). A candidate implementation is linked below; I can submit a PR with focused regression tests.

## Mechanism

The inner loop ([utils.py#L561-L616](https://github.com/isce-framework/isce3/blob/0a8df45ddf780d22e41bf5358502121e6f09b7df/python/packages/nisar/products/insar/utils.py#L561-L616)) executes per output pixel:

- two scalar pybind calls to `SubSwaths.get_sample_sub_swath` (reference, and offset-shifted secondary) via [`_compute_subswath_mask_id`](https://github.com/isce-framework/isce3/blob/0a8df45ddf780d22e41bf5358502121e6f09b7df/python/packages/nisar/products/insar/utils.py#L317);
- NumPy scalar indexing into the two `inputDataExceptionMask` arrays plus Python-level bit packing;
- a `list.append`.

[...]

Issue #335, opened 2026-07-09
`generate_insar_mask()`: inputDataExceptionMask bits silently dropped from InSAR mask layer under NumPy ≥ 2.0 (uint8 scalar left-shift overflows to 0)


xhuang-jpl:
## Summary

`generate_insar_mask()` in `python/packages/nisar/products/insar/utils.py` packs the reference/secondary RSLC `inputDataExceptionMask` values into the InSAR `mask` layer by left-shifting **`np.uint8` scalars** by 16 and 8 bits:

```python
# utils.py, L592
ref_input_exception_mask_id = ref_input_exception_mask[int(i),int(j)] << 16
...
# utils.py, L599
sec_input_exception_mask_id = sec_input_exception_mask[sec_i,sec_j] << 8
```

Both masks are loaded as `np.uint8` ([L551](https://github.com/isce-framework/isce3/blob/bdf1f6fb9175ffe9e53b61839c797e69947dd268/python/packages/nisar/products/insar/utils.py#L551)), so indexing yields a `np.uint8` scalar.

* With **NumPy < 2.0**, legacy value-based promotion widens `np.uint8 scalar << python int` to `int64`, so the shifts produce the intended values.
* With **NumPy ≥ 2.0** (NEP 50 promotion rules), the result stays `uint8`. Shifting a uint8 by ≥ 8 bits pushes every bit out and yields **0** — silently, with no warning or error.

Consequence under NumPy ≥ 2.0: bits 8–15 (secondary exception mask) and 16–23 (reference exception mask) of the output `mask` dataset are always zero; only the sub-swath id in the low bits survives. This corrupts the `mask` layer written via `InSAR_L1_writer.py` ([L535](https://github.com/isce-framework/isce3/blob/bdf1f6fb9175ffe9e53b61839c797e69947dd268/python/packages/nisar/products/insar/InSAR_L1_writer.py#L535), [L766](https://github.com/isce-framework/isce3/blob/bdf1f6fb9175ffe9e53b61839c797e69947dd268/python/packages/nisar/products/insar/InSAR_L1_writer.py#L766)) and everything downstream that inherits it. Note the corruption only manifests when `inputDataExceptionMask` is non-zero — i.e., exactly the scenes those bits exist to flag.

[...]

PR #358, opened 2026-08-17
Performance tune for prepare_insar_hdf5, resample_slc_v2, and rubbersheet

lijun99:
This PR aims to speed up three modules in insar workflows, prepare_insar_hdf5, resample_slc_v2, and rubbersheet. I put them together for easier testing. Please let me know if separate PRs are preferred. 

1) prepare_insar_hdf5 
Replace the per-pixel Python loop (and its per-pixel calls into SubSwaths.get_sample_sub_swath) with vectorized numpy operations, reading offset raster rows only as needed to keep memory bounded. Tests show ~3.5-10x speedup on the prepare_insar_hdf5 step, and output verified bit-for-bit identical to the original.

2) resample_slc_v2
Replace the legacy HDF5 reader with HDF5OptimizedReader used in unwrap.py, crossmul.py, and others. This cuts the resample step's SLC I/O read time by ~5-10x. Output is verified bit-for-bit identical to the original.

3) rubbersheet
Update rubbersheet polyfit critical_value from 0.1 to Baarda's convention 3.29. 
0.1 (keeping only 8%) is far tighter than the standard w-test threshold, causing excessive outlier-removal iterations without a corresponding accuracy benefit. 3.29 (alpha=0.001) is the conventional geodetic data-snooping critical value. Recommended range 1.69-3.29. 
Here is a test on a scene from Afar, Ethiopia, 
> crit_value    Iterations      Time    % of data removed       Removed points with corr_peak > 0.7
> 0.1 (current) 23,187  72.35s  58.0%   68.3% (mostly good matches!)
> 1.0   3,197   16.30s  8.0%    3.4%
> 2.0   3,111   15.76s  7.8%    2.8%
> 3.29 (Baarda's standard)      3,008   14.66s  7.5%    2.8%

It turns out that 0.1 removes 68.3% of good pixels for polyfit, not only makes the fitting less representative, but also leads to long iterations to finish.  
In tests, choosing 3.29 leads to ~5-10x speedup for the polyfit. (Note that the rubbersheet procedure has heavy I/Os, the actual I/O performance depends on types of storage systems.) There is a small difference, ~0.2 rad across the scene, from a ramp, since the fitting parameters are slightly different. 

BTW, I am also working on a [gpu phase unwrapping](https://github.com/earthdef/cuPHU). It shows some early success; but needs more test. Please give it a try when you have time.

### DIFF
--- a/cxx/isce3/geometry/TopoLayers.cpp
+++ b/cxx/isce3/geometry/TopoLayers.cpp
@@ -110,9 +110,18 @@ TopoLayers::TopoLayers(const size_t linesPerBlock, isce3::io::Raster* xRaster,
 
 void TopoLayers::writeData(size_t xidx, size_t yidx)
 {
-    std::vector<std::variant<double*, float*, short*>> valarrays {&_x[0],
-            &_y[0], &_z[0], &_inc[0], &_hdg[0], &_localInc[0], &_localPsi[0],
-            &_sim[0], &_mask[0], &_groundToSatEast[0], &_groundToSatNorth[0]};
+    // Layers whose raster wasn't requested are left with size-0 valarrays
+    // (see setBlockSize), so indexing element 0 unconditionally would be
+    // out-of-bounds. Only take the address when the valarray is non-empty.
+    auto ptr = [](auto& valarray) {
+        using T = typename std::remove_reference_t<decltype(valarray)>::value_type;
+        return valarray.size() ? &valarray[0] : static_cast<T*>(nullptr);
+    };
+
+    std::vector<std::variant<double*, float*, short*>> valarrays {ptr(_x),
+            ptr(_y), ptr(_z), ptr(_inc), ptr(_hdg), ptr(_localInc),
+            ptr(_localPsi), ptr(_sim), ptr(_mask), ptr(_groundToSatEast),
+            ptr(_groundToSatNorth)};
 
     std::vector<isce3::io::Raster*> rasters {_xRaster, _yRaster, _zRaster,
             _incRaster, _hdgRaster, _localIncRaster, _localPsiRaster,
--- a/cxx/isce3/geometry/TopoLayers.cpp
+++ b/cxx/isce3/geometry/TopoLayers.cpp
@@ -113,15 +113,16 @@ void TopoLayers::writeData(size_t xidx, size_t yidx)
     // Layers whose raster wasn't requested are left with size-0 valarrays
     // (see setBlockSize), so indexing element 0 unconditionally would be
     // out-of-bounds. Only take the address when the valarray is non-empty.
-    auto ptr = [](auto& valarray) {
+    auto safe_deref = [](auto& valarray) {
         using T = typename std::remove_reference_t<decltype(valarray)>::value_type;
         return valarray.size() ? &valarray[0] : static_cast<T*>(nullptr);
     };
 
-    std::vector<std::variant<double*, float*, short*>> valarrays {ptr(_x),
-            ptr(_y), ptr(_z), ptr(_inc), ptr(_hdg), ptr(_localInc),
-            ptr(_localPsi), ptr(_sim), ptr(_mask), ptr(_groundToSatEast),
-            ptr(_groundToSatNorth)};
+    std::vector<std::variant<double*, float*, short*>> valarrays {
+            safe_deref(_x), safe_deref(_y), safe_deref(_z), safe_deref(_inc),
+            safe_deref(_hdg), safe_deref(_localInc), safe_deref(_localPsi),
+            safe_deref(_sim), safe_deref(_mask), safe_deref(_groundToSatEast),
+            safe_deref(_groundToSatNorth)};
 
     std::vector<isce3::io::Raster*> rasters {_xRaster, _yRaster, _zRaster,
             _incRaster, _hdgRaster, _localIncRaster, _localPsiRaster,
--- a/environment.yml
+++ b/environment.yml
@@ -20,7 +20,7 @@ dependencies:
   - ninja
   - numpy>=1.20
   - pyaps3>=0.3
-  - pybind11>=2.5
+  - pybind11>=2.5,!=3.1.0
   - pyre>=1.12.5
   - pysolid>=0.2
   - pytest
--- a/environment.yml
+++ b/environment.yml
@@ -20,7 +20,7 @@ dependencies:
   - ninja
   - numpy>=1.20
   - pyaps3>=0.3
-  - pybind11>=2.5,!=3.1.0
+  - pybind11>=2.5
   - pyre>=1.12.5
   - pysolid>=0.2
   - pytest
--- a/extern/CMakeLists.txt
+++ b/extern/CMakeLists.txt
@@ -113,7 +113,7 @@ macro(getpackage_pybind11)
         if(NOT pybind11_FOUND)
             fetch_extern_repo(pybind11
                 GIT_REPOSITORY  https://github.com/pybind/pybind11
-                GIT_TAG         v2.5.0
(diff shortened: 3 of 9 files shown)


lijun99:
The test failures seem to be pre-existing. Please take a look at this PR, @hfattahi @bhawkins @xhuang-jpl.

Tyler-g-hudson:
@lijun99 Could you rebase to the develop branch? That should resolve these test failures.

Tyler-g-hudson:
It looks like Linux Latest Debug and (particularly) MacOS Latest Debug have large numbers of test failures outside of the ones that @bhawkins fixed

Tyler-g-hudson:
We're getting a weird mix of test failures here.

**Minimum Python Supported:**
- ❌ Failed test.cxx.isce3.geometry.geometry.geometry
- ✅ No SEGFAULT in test.python.pybind.geometry.pntintersect

**Linux Latest Release:**
- ❌ Failed test.cxx.isce3.geometry.geometry.geometry
- ❌ SEGFAULT in test.python.pybind.geometry.pntintersect

**Linux Latest Debug:**
- ✅ No fail in test.cxx.isce3.geometry.geometry.geometry
- ✅ No SEGFAULT in test.python.pybind.geometry.pntintersect
- ❌ Lots of aborted subprocesses in python plus one in cxx
- ❌ Strange errors with files and headers in some python workflows

**macOS Latest Release:**
- ✅ No fail in test.cxx.isce3.geometry.geometry.geometry
- ❌ SEGFAULT in test.python.pybind.geometry.pntintersect

**macOS Latest Debug:**
- ❌ All Python tests fail with errors importing the pybinds:
```
../../../../../../install/packages/isce3/__init__.py:2: in <module>
    from .ext import extisce3
../../../../../../install/packages/isce3/ext/__init__.py:3: in <module>
    from . import isce3 as extisce3
E   ImportError: Internal error while parsing type signature (2)
```

**EDIT:** These appear to be universal to recent PR's. I will open an issue about it.

Tyler-g-hudson:
This pattern of errors is recorded now in #368. I don't see anything in the content of the errors that would be special to these changes, but the resample_slc and resample_slc_v2 Python workflows are effected in Linux Latest Debug so there's a small chance that an error in this PR for Linux Latest Debug would be hidden by those errors.

@hfattahi how do you feel about proceeding with review with this in mind? I think such a hidden error is very improbable, so my position is that we can proceed.

Tyler-g-hudson:
The CI error issue is fixed by #366 - please rebase when that PR merges and rebase again.

lijun99:
Great, all test passed! Thank you, @bhawkins and @Tyler-g-hudson !

Tyler-g-hudson (approved):
With regards to the Resample update, this looks good!

hfattahi:
@lijun99 thanks for this PR. I am ok with Resample and h5 prep changes but couple of us think that the polyfitting changes requires intense testing. To make progress faster, I suggest to decouple this PR to three separate PRs. Resample can be merged quickly. we will review h5 prep and we will need to discuss and test polyfitting further.

lijun99:
I have spilt the PR into three separate parts. 

2. resample_slc_v2, resubmitted as #372. 

3. rubbersheet, per discussion with @hfattahi, changing the default critical_value should be cautious. Let's do more tests before changing it. But from my research into the topic, 0.1 doesn't make sense to me. 

1. prepare_insar_hdf5, I noticed #359 from @s-sasaki-earthsea-wizard, which is on the same line. I will check his approach before deciding whether I will need to submit a new one. 
 
Thanks!

s-sasaki-earthsea-wizard:
@lijun99 Thanks for the ping — and sorry for the overlap: I should have spotted #358 before filing #359 (I had been tracking this loop via #354 / #335 since Aug 13, but missed your PR when I filed mine two days after yours).

Since the two `utils.py` changes are independent implementations of the same fix, I ran yours through the benchmark harness I built for #354 to make the comparison concrete. Setup: `utils.py` from the #358 head (`bdb5974`), from the #359 head (`dbb4e8d`), and pristine `develop` (blob `0643bde0c`) as baseline, all against the same compiled isce3 (`develop`, numpy 1.26.4, py3.12).

**Timing** — synthetic reproducer from #354 (6840 x 10581 = 72.4 Mpx RIFG-like grid, three sub-swaths, no NISAR granules needed):

| utils.py variant | wall time (median of 3) | vs scalar | peak RSS |
|---|---|---|---|
| develop (scalar per-pixel loop) | 572.6 s (single run) | 1x | — |
| #358 (`_get_sample_subswath_grid`) | 15.8 s | ~36x | 9.9 GB |
| #359 (`_subswath_numbers`) | 10.0 s | ~57x | 1.6 GB |

The output mask CRC is identical (`0x295ba19b`) across all three variants and all runs on this fixture, so on correctness of the hot path and on the size of the win we are in full agreement — both rewrites eliminate the per-pixel pybind calls, which is where ~all the time goes. The residual differences come from temporaries: #358 materializes several full-grid float64/int64 arrays via meshgrid (hence the peak RSS), while #359 processes row-wise. For reference, on a quieter host I previously measured the scalar loop at 383.3 s with the same ~57x ratio for #359, so the ratios are stable across host-load conditions.

**A minor contract-fidelity note on the #358 version.** I also ran both implementations against a self-contained scalar oracle that calls `SubSwaths.get_sample_sub_swath` directly (adversarial fixtures: exact `k + 0.5` offsets where the two rounding rules diverge, empty sub-swath arrays, out-of-swath pushes). Both match the oracle on every file-loadable input — including an RSLC without sub-swath datasets, which `Serialization.h` deserializes as `num_sub_swaths == 1` with an empty valid-samples array, a state both PRs handle correctly via the empty-array short-circuit. The one divergence I found is `num_sub_swaths == 0`: there the scalar API returns 1 for in-bounds samples (`SubSwaths::getSampleSubSwath`: *"If the dataset does not have sub-swaths information, consider samples valid and belonging to the first sub-swath"*), while `_get_sample_subswath_grid`'s per-sub-swath loop never executes and leaves 0, dropping the sub-swath digits. To be clear about reachability: this state cannot come from RSLC deserialization (the `numSubSwaths` setter even throws for `n <= 0`), only from programmatic construction such as the pybind list constructor `SubSwaths(length, width, [])` — so it is a fidelity gap against the documented scalar API rather than a production-path bug, and it matters mostly for tests and library consumers that build `SubSwaths` objects directly. If you keep your implementation it is a one-line guard to make the helper a full drop-in.

**Tests.** #359 also carries a pytest suite (~350 lines, `tests/python/packages/nisar/products/insar/utils.py`) that validates the vectorized mask against the scalar pybind API as oracle, including the two distinct rounding rules (`int(x + 0.5)` truncation for the sub-swath lookup vs `round()` half-even for the exception-mask lookup), the empty-interval short-circuit, the `num_sub_swaths == 0` fallback above, and the uint32 bit packing that #335 is about.

Happy to resolve the overlap either way: if you and the maintainers prefer the #358 implementation for the prepare part, I can rebase #359 down to a tests-only PR — the suite validates either implementation (modulo the fallback above). Otherwise #359 already carries implementation + tests. Whichever way, the cross-check script and logs are public in my benchmark repo ([`s-sasaki-earthsea-wizard/isce3-benchmark`](https://github.com/s-sasaki-earthsea-wizard/isce3-benchmark), `scripts/crosscheck_insar_mask_oracle.py` + `artifacts/pr358-crosscheck-20260827/`).

s-sasaki-earthsea-wizard:
For completeness: the real-data leg of the comparison above (NISAR frame; bitwise + peak-RSS results for both implementations) is now posted in https://github.com/isce-framework/isce3/pull/359#issuecomment-5434042896 — keeping the data in one place since that is where the prepare-part decision lands.

Issue #304, opened 2026-06-18
(CF Compliance) `valid_min` and `valid_max` Attribute Dtype Mismatch


nemo794:
**Description:** The `valid_min` and `valid_max` attributes have different data types than the datasets they describe. This is a related (but separate) issue from #297 .

**CF Requirement (Section 2.5.1):**

> "The attributes `valid_min`, `valid_max` and `valid_range` ... are all of **the same type as the variable** to which they are attached."
>
> — CF Conventions Version 1.7, Section 2.5.1: Missing Data
> 
> Source: http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#_missing_data

**Note:** This issue primarily impacts RSLC and the five InSAR. (GSLC and GCOV resolved this issue via #213 ).

**Impact:**
- Inconsistent behavior across different CF-compliant software
- Tools may reject values that should be valid or accept invalid values

**Fix Required:** Ensure `valid_min` and `valid_max` use the same dtype as their parent dataset.

**Common Patterns:**
1. **float32 datasets with float64 valid_min/max** (most common)
   - Geolocation/radarGrid metadata cubes
   - Affects: RSLC, RIFG, RUNW, ROFF, GUNW, GOFF

2. **uint32 datasets with int64 valid_min** (no valid_max)
   - Mask datasets
   - Affects: RIFG, RUNW, ROFF, GUNW, GOFF

#### Occurrences by Product:

**RSLC**
```
NISAR_L1_PR_RSLC_001_149_D_080_2005_QPDH_A_20251003T130952_20251003T131016_X05016_N_P_J_001.h5

Dataset: /science/LSAR/RSLC/metadata/geolocationGrid/alongTrackUnitVectorX
  Dataset dtype: float32
  valid_min (actual): -1.0 (dtype: float64)
  valid_min (should be): np.float32(-1.0)
  valid_max (actual): 1.0 (dtype: float64)
  valid_max (should be): np.float32(1.0)

Other affected datasets (same pattern):
  - /science/LSAR/RSLC/metadata/geolocationGrid/alongTrackUnitVectorY
  - /science/LSAR/RSLC/metadata/geolocationGrid/elevationAngle (valid_min: 0.0, valid_max: 90.0)
  - /science/LSAR/RSLC/metadata/geolocationGrid/incidenceAngle (valid_min: 0.0, valid_max: 90.0)
[...]

PR #361, opened 2026-08-19
Fix InSAR Product Spec Issue #304

xhuang-jpl:
Fix the InSAR product spec issue #304


python/packages/nisar/workflows/compute_stats.py
```diff
@@ -53,11 +53,11 @@ def compute_stats_real_data(raster, h5_ds):
         stats_obj = isce3.math.compute_raster_stats_float64(raster)[0]
     else:
         stats_obj = isce3.math.compute_raster_stats_float32(raster)[0]
```
hfattahi, line 55: @bhawkins brought up that this function may not cover 32 bits integer datasets. Let's make sure we are safe
```suggestion
    if raster.datatype() == gdal.GDT_Float32:
         stats_obj = isce3.math.compute_raster_stats_float32(raster)[0]
     else:
         stats_obj = isce3.math.compute_raster_stats_float64(raster)[0]
```

We still may need to raise an exception for int 64 bits.

hfattahi (approved):
one minor comment.

python/packages/nisar/workflows/compute_stats.py
xhuang-jpl: > [@bhawkins](https://github.com/bhawkins) brought up that this function may not cover 32 bits integer datasets. Let's make sure we are safe
> 
> Suggested change
>         stats_obj = isce3.math.compute_raster_stats_float32(raster)[0]
>     if raster.datatype() == gdal.GDT_Float32:
>          stats_obj = isce3.math.compute_raster_stats_float32(raster)[0]
>      else:
>          stats_obj = isce3.math.compute_raster_stats_float64(raster)[0]
> We still may need to raise an exception for int 64 bits.

just checked that this function is not applied to any integer data type dataset in the insar workflows.

Image

hfattahi (approved):
LGTM!

hfattahi:
@bhawkins please take another look.

bhawkins (approved):

Issue #318, opened 2026-06-18
(RIFG/RUNW/GUNW) `isOffsetsBlendingApplied` metadata value is "None"; should be a boolean


nemo794:
`isOffsetsBlendingApplied` metadata value is "None"; should be "True" or "False"

Path: `/science/LSAR/RIFG/metadata/processingInformation/parameters/pixelOffsets/frequencyA/isOffsetsBlendingApplied`
-  Current value: fixed-length string `'None'`
 - Expected value: "True" or "False" (per NISAR convention for boolean flags)

Dataset description:
`Flag to indicate if pixel offsets are the results of blending multi-resolution layers of pixel offsets`

InSAR granules checked:
`NISAR_L1_PR_RIFG_013_118_A_171_014_2000_SH_20260222T102514_20260222T102520_20260306T102515_20260306T102521_X05016_P_P_J_001.h5`
`NISAR_L1_PR_RUNW_007_106_A_140_008_7700_SH_20251211T140849_20251211T140926_20251223T140850_20251223T140926_X05016_N_P_J_001.h5`
`NISAR_L2_PR_GUNW_001_062_A_019_006_2000_SH_20250927T114508_20250927T114542_20251126T114516_20251126T114545_X05016_P_F_J_001.h5`

cc: @hfattahi @xhuang-jpl @seongsujeong @oberonia78

PR #363, opened 2026-08-20
Fix InSAR Product Spec Issue #318

xhuang-jpl:
This PR is to fix the product spec issue #318


python/packages/nisar/products/insar/InSAR_base_writer.py
```diff
@@ -748,7 +748,7 @@ def add_pixeloffsets_to_procinfo_params_group(self):
             ),
             DatasetParams(
                 "isOffsetsBlendingApplied",
-                str(merge_gross_offset),
+                is_roff,
```
hfattahi:
```suggestion
                str(is_roff),
```

hfattahi (commented):
@gshiroma brought up that you need to ensure boolean is string as that is what the spec expects.

hfattahi (approved):
LGTM.

bhawkins (approved):

PR #366, opened 2026-08-20
Solve CI failures

### LINKED ISSUE #368: Strange mix of errors in recent PR's
Starting at PR #333 (though this is not culprit since it is not merged), there have been several pull requests with a peculiar pattern of errors appearing on our dev machines, with most of the recent PR's having the errors listed below. Debug builds seem to have wholly different sets of issues from each other and from the other builds, while the Release builds seem to just have errors with the `geometry` cxx/pybind tests.

**Minimum Python Supported:**
- ❌ Failed test.cxx.isce3.geometry.geometry.geometry
- ✅ No SEGFAULT in test.python.pybind.geometry.pntintersect

**Linux Latest Release:**
- ❌ Failed test.cxx.isce3.geometry.geometry.geometry
- ❌ SEGFAULT in test.python.pybind.geometry.pntintersect

**Linux Latest Debug:**
- ✅ No fail in test.cxx.isce3.geometry.geometry.geometry
- ✅ No SEGFAULT in test.python.pybind.geometry.pntintersect
- ❌ Lots of aborted subprocesses in python plus one in cxx
- ❌ Strange errors with files and headers in some python workflows

**macOS Latest Release:**
- ✅ No fail in test.cxx.isce3.geometry.geometry.geometry
- ❌ SEGFAULT in test.python.pybind.geometry.pntintersect

**macOS Latest Debug:**
- ❌ All Python tests fail with errors importing the pybinds:
```
../../../../../../install/packages/isce3/__init__.py:2: in <module>
    from .ext import extisce3
../../../../../../install/packages/isce3/ext/__init__.py:3: in <module>
    from . import isce3 as extisce3
E   ImportError: Internal error while parsing type signature (2)
```

bhawkins:
CI is failing for every PR right now. I've narrowed the problem with `test.python.pybind.geometry.pntintersect` to the latest version of pybind11 (v3.1.0), while the previous version (3.0.4) seems to work okay. Claude thinks it has something to do with the new `call_impl` method, but I didn't go too deep down that rabbit hole. Changing the function signature in the Python binding of `slantrange_from_lookvec` to pass the ellipsoid by value instead of by `const&` seems to work around the issue, but that seems like a hack. ~For now I just blacklisted the latest pybind11 version in hopes it gets fixed upstream.~ Edit: As noted below, it's not actually necessary to blacklist the latest pybind11 version.

I think there may be some other CI failures as well and will use this branch to work on them.

### DIFF
--- a/cxx/isce3/geometry/TopoLayers.cpp
+++ b/cxx/isce3/geometry/TopoLayers.cpp
@@ -110,9 +110,18 @@ TopoLayers::TopoLayers(const size_t linesPerBlock, isce3::io::Raster* xRaster,
 
 void TopoLayers::writeData(size_t xidx, size_t yidx)
 {
-    std::vector<std::variant<double*, float*, short*>> valarrays {&_x[0],
-            &_y[0], &_z[0], &_inc[0], &_hdg[0], &_localInc[0], &_localPsi[0],
-            &_sim[0], &_mask[0], &_groundToSatEast[0], &_groundToSatNorth[0]};
+    // Layers whose raster wasn't requested are left with size-0 valarrays
+    // (see setBlockSize), so indexing element 0 unconditionally would be
+    // out-of-bounds. Only take the address when the valarray is non-empty.
+    auto ptr = [](auto& valarray) {
+        using T = typename std::remove_reference_t<decltype(valarray)>::value_type;
+        return valarray.size() ? &valarray[0] : static_cast<T*>(nullptr);
+    };
+
+    std::vector<std::variant<double*, float*, short*>> valarrays {ptr(_x),
+            ptr(_y), ptr(_z), ptr(_inc), ptr(_hdg), ptr(_localInc),
+            ptr(_localPsi), ptr(_sim), ptr(_mask), ptr(_groundToSatEast),
+            ptr(_groundToSatNorth)};
 
     std::vector<isce3::io::Raster*> rasters {_xRaster, _yRaster, _zRaster,
             _incRaster, _hdgRaster, _localIncRaster, _localPsiRaster,
--- a/cxx/isce3/geometry/TopoLayers.cpp
+++ b/cxx/isce3/geometry/TopoLayers.cpp
@@ -113,15 +113,16 @@ void TopoLayers::writeData(size_t xidx, size_t yidx)
     // Layers whose raster wasn't requested are left with size-0 valarrays
     // (see setBlockSize), so indexing element 0 unconditionally would be
     // out-of-bounds. Only take the address when the valarray is non-empty.
-    auto ptr = [](auto& valarray) {
+    auto safe_deref = [](auto& valarray) {
         using T = typename std::remove_reference_t<decltype(valarray)>::value_type;
         return valarray.size() ? &valarray[0] : static_cast<T*>(nullptr);
     };
 
-    std::vector<std::variant<double*, float*, short*>> valarrays {ptr(_x),
-            ptr(_y), ptr(_z), ptr(_inc), ptr(_hdg), ptr(_localInc),
-            ptr(_localPsi), ptr(_sim), ptr(_mask), ptr(_groundToSatEast),
-            ptr(_groundToSatNorth)};
+    std::vector<std::variant<double*, float*, short*>> valarrays {
+            safe_deref(_x), safe_deref(_y), safe_deref(_z), safe_deref(_inc),
+            safe_deref(_hdg), safe_deref(_localInc), safe_deref(_localPsi),
+            safe_deref(_sim), safe_deref(_mask), safe_deref(_groundToSatEast),
+            safe_deref(_groundToSatNorth)};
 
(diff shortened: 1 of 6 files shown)


bhawkins:
There were two dereferencing bugs caught thanks to the newer compilers/runtime libraries. One was in the isce3 library while the other was just in the unit test implementation.

There was another bug in the geo2rdr test related to using an uninitialized aztime. The code actually does check whether the time falls inside the start/end time of the orbit and has a strategy to deal with that. However, in this case the extent of the LUT2d doesn't cover the full orbit, it's not set to extrapolate, and the compiler happened to set the aztime to a value inside the orbit but outside the LUT2d. Setting the aztime guess explicitly in the test prevents the failure.

Now it's just the macos debug build that's failing.

bhawkins:
Finally, success! There actually wasn't a problem in pybind11, but the update merely exposed a problem in the isce3 bindings. Basically two methods had the argument list `(Vec3, Vec3, Ellipsoid)` but one of the .cpp files forgot to include the pybind11/eigen.h header. This lead to templates getting instantiated differently in different .o files and a violation of the One Definition Rule (ODR) that didn't show up until runtime, with different behavior for different build types.

I must say, after a period of relative complacency, this episode really reminded me of what I don't like about C++!

tools/imagesets/oracle8conda/distrib_nisar/Dockerfile
bhawkins, line 15: I'm not sure why this change from #349 is showing up in the diff...

Tyler-g-hudson: It looks like commit [a1c757b](https://github.com/isce-framework/isce3/pull/366/commits/a1c757b106b54b95c19a43cc7edb754aba825dd9) from that PR somehow ended up in your git history for this one. Maybe a rebase would fix it?

bhawkins: Weird. I'm thinking what happened is that I merged origin/develop into this branch and then rebased onto my local develop, which was one commit behind. I just rebased onto origin/develop and force pushed. Diff seems sensible now.

Tyler-g-hudson (approved):
LGTM!

jplzhan:
LGTM, I only added some comments that are both just nitpicks.

cxx/isce3/geometry/TopoLayers.cpp
```diff
@@ -110,9 +110,18 @@ TopoLayers::TopoLayers(const size_t linesPerBlock, isce3::io::Raster* xRaster,
 
 void TopoLayers::writeData(size_t xidx, size_t yidx)
 {
-    std::vector<std::variant<double*, float*, short*>> valarrays {&_x[0],
-            &_y[0], &_z[0], &_inc[0], &_hdg[0], &_localInc[0], &_localPsi[0],
-            &_sim[0], &_mask[0], &_groundToSatEast[0], &_groundToSatNorth[0]};
+    // Layers whose raster wasn't requested are left with size-0 valarrays
+    // (see setBlockSize), so indexing element 0 unconditionally would be
+    // out-of-bounds. Only take the address when the valarray is non-empty.
+    auto ptr = [](auto& valarray) {
```
jplzhan: I only have a nitpick here that the name `ptr` for a lambda is pretty vague. The intention of the variable is to safely dereference some pointers, so maybe `safe_deref` is a better name. The point of the function is clear from reading the code though.

tests/cxx/isce3/geometry/geometry/geometry.cpp
```diff
@@ -133,7 +133,12 @@ TEST_F(GeometryTest, GeoToRdr)
             -115.72466801139711 * radians, 34.65846532785868 * radians, 1772.0};
 
     // Run geo2rdr
-    double aztime, slantRange;
+    // aztime is used as the initial azimuth time guess, so it must be
+    // initialized to a value within the orbit's time span rather than left
+    // undefined.  It also needs to fall within the domain of the Doppler
+    // LUT2d, or the LUT2d must be configured to allow extrapolation.
+    // In this case, orbit.midTime() satisfies both constraints.
+    double aztime = orbit.midTime(), slantRange;
```
jplzhan: In general, variables in C++ should always be initialized according to `ES.20` in the C++ Core Guidelines: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#res-always

So, I think even though it is not currently a problem, `slantRange` should also be initialized here, to avoid a code smell.

Or rather, in general, it would be good to see where else this occurs in the ISCE3 repo.

hfattahi (approved):
LGTM. Thanks @bhawkins for identifying all those edge cases. Very much appreciated.

tests/cxx/isce3/geometry/geometry/geometry.cpp
bhawkins: I agree it's a good idea to follow ES.20 and changed this occurrence. I think hunting down other occurrences would require some static analysis tool and is outside the scope of this PR.

PR #367, opened 2026-08-24
Add whirlwind unwrapping algorithm

xhuang-jpl:
This PR adds support for the whirlwind phase unwrapping algorithm to the InSAR workflow.


Tyler-g-hudson:
Hey @xhuang-jpl - the PR to fix these tests has been merged, you should be able to resolve them by rebasing to develop.

share/nisar/defaults/insar.yaml
```diff
@@ -810,6 +810,92 @@ runconfig:
...
+                    # Effective number of looks used to form the sample coherence. Must be at least 1.
+                    # A higher number of looks means higher confidence in coherence and a narrower
+                    # coherence cost model. Very large values (above ~80 looks) are capped for the
+                    # cost model since the phase PDF has effectively converged by then.
+                    # If not specified, nlooks (n_e) is estimated as:
+                    # $$
+                    # n_e = \frac{d_r d_a}{\rho_r \rho_a}
+                    # $$
+                    # where $d_r$ and $d_a$ are the sample spacing in range and azimuth,
+                    # and $\rho_r$ and $\rho_a$ are the range and azimuth resolution.
```
hfattahi, line 823: @scottstanie can you clarify if there is an upper bound when n_e is computed by whirlwind?

share/nisar/defaults/insar.yaml
```diff
@@ -810,6 +810,92 @@ runconfig:
...
+                    nlooks:
+                    # File path to a binary mask of valid pixels. Zeros in this raster indicate
+                    # interferogram pixels that should be masked out. If provided, it must have
+                    # the same dimensions as the input interferogram and boolean or 8-bit integer datatype.
+                    # If not provided, defaults to (igram != 0) & (coherence > 0).
```
hfattahi: what is the format of the mask file? tif, etc?

share/nisar/defaults/insar.yaml
```diff
@@ -810,6 +810,92 @@ runconfig:
...
+                    mask:
+                    # Post-processing step that re-levels regions the valid mask splits into
+                    # disconnected pieces (e.g., land slabs separated by low-coherence river).
+                    # Estimates each region's level from unwrapped phase at region boundaries
+                    # and snaps to integer cycles, propagated along MST rooted at largest region.
+                    bridge: True
+                    # Coarse-solve factor for noisy scenes. When >1, the interferogram is
+                    # coherently averaged into downsample x downsample blocks and unwrapped at
+                    # coarse resolution. Only the integer cycle is borrowed back to full resolution,
+                    # so every per-pixel value is preserved. Detail finer than block scale aliases.
```
hfattahi: what does "Detail finer than block scale aliases." mean?

share/nisar/defaults/insar.yaml
```diff
@@ -810,6 +810,92 @@ runconfig:
...
+                    # May help noisy scenes, provides little benefit for clean scenes.
+                    downsample: 1
+                    # Spiral persistent-scatterer interpolation pre-pass. When True, every valid
```
hfattahi, line 834: what is pre-pass?

hfattahi (commented):
Thanks @xhuang-jpl . Looks good to me overall. See below for few minor comments. @scottstanie appreciate if you browse over this PR.

hfattahi:
Rebasing with develop should solve the unit test failures.

share/nisar/defaults/insar.yaml
scottstanie: agreed that it's worth adding to the comment.
it's any GDAL-readable raster, same as the snaphu mask option, since both use `open_raster`. 
whirlwind coerces with `np.ascontiguousarray(mask, dtype=bool)`, so 
- uint8 works 
- also a float mask is accepted, with nonzero is `True`

share/nisar/defaults/insar.yaml
scottstanie: i think this was taken from internal whirlwind jargon so we could probably even take out this downsample option. i'm not a big fan of it now that the main path is so fast. the downsample,unwrap,upsample is something other people have done before, so i added it early on

share/nisar/defaults/insar.yaml
```diff
@@ -810,6 +810,92 @@ runconfig:
...
+                    # connected components. "auto" uses 0.32/sqrt(nlooks); pass float for fixed cutoff,
+                    # or None to disable cutoff and use conncomp_reliability instead. Takes precedence
+                    # over conncomp_reliability. Only used when conncomp_algorithm="snaphu".
+                    conncomp_min_coherence: auto
+                    # Lower-level conservativeness knob for default ("snaphu") connected components,
+                    # in inverse-variance (1/sigma2) units, so values are small. Used only when
+                    # conncomp_min_coherence is None. 0 labels essentially every reliably unwrapped
+                    # pixel; raise to cut more low-coherence interior edges. Only used when
+                    # conncomp_algorithm="snaphu".
+                    conncomp_reliability: 0.0
```
scottstanie:
```suggestion
                    conncomp_reliability: 0.5
```

python/packages/nisar/workflows/unwrap.py
```diff
@@ -481,6 +530,63 @@ def set_phass_attributes(cfg: dict):
...
+    if cfg.get('max_ncomps') is not None:
+        ww_kwargs['max_ncomps'] = cfg['max_ncomps']
+    if cfg.get('goldstein_alpha') is not None:
+        ww_kwargs['goldstein_alpha'] = cfg['goldstein_alpha']
+    if cfg.get('goldstein_psize') is not None:
+        ww_kwargs['goldstein_psize'] = cfg['goldstein_psize']
+
+    return ww_kwargs
+
+
```
scottstanie, line 578: i think this whole thing could be condensed to list the keys in a module level tuple and then make it 

```suggestion
def set_whirlwind_attributes(cfg: dict):
    """
    Return dictionary with whirlwind parameters from user-defined config

    Parameters
    ----------
    cfg: dict
        Dictionary containing user-defined whirlwind parameters

    Returns
    -------
    ww_kwargs: dict
        Dictionary of whirlwind parameters for whirlwind.unwrap()
    """
    return {key: cfg[key] for key in WHIRLWIND_OPTIONS if cfg[key] is not None}


```


adding something like this to the top
```python
  +# `whirlwind.unwrap`. `nlooks` and `mask` are excluded because they need
  +# conversion (looks estimation, raster read) and are set by the caller.
  +WHIRLWIND_OPTIONS = (
  +    'bridge', 'downsample', 'interpolate', 'interp_cutoff',
  +    'interp_num_neighbors', 'interp_max_radius', 'interp_min_radius',
  +    'interp_alpha', 'conncomp_min_coherence','conncomp_reliability', 
  +    'goldstein_psize', 'goldstein_alpha', 'min_size_px', 'max_ncomps',
  +)
 ```

File: python/packages/nisar/workflows/unwrap.py [hunk 7]
```diff
@@ -313,6 +313,55 @@ def run(cfg: dict, input_hdf5: str, output_hdf5: str):
...
+                        err_str = ("whirlwind-insar is not installed. "
+                                   "Install it with 'pip install whirlwind-insar' or "
+                                   "'conda install -c conda-forge whirlwind-insar'")
+                        error_channel.log(err_str)
+                        raise ImportError(err_str)
+
+                    # Get whirlwind parameters using helper function
+                    ww_kwargs = set_whirlwind_attributes(whirlwind_cfg)
+                    ww_kwargs['nlooks'] = float(nlooks)
+                    ww_kwargs['mask'] = mask_array if mask_array is not None else None
```
scottstanie: since above we always make a mask_array,
```suggestion
                    ww_kwargs['mask'] = mask_array
```

scottstanie:
initial things-

- do we need to add `whirlwind-insar >= 0.9.0` to somewhere? environment.yml? tools/imagesets/oracle8conda/runtime/environment.yml? the spec file?
   - related- seems like the `try/except ImportError` at the top is only partially making it an optional dependency

scottstanie (commented):
Here's some suggestions, and https://github.com/xhuang-jpl/isce3/compare/add_whirlwind...scottstanie:isce3:whirlwind-followups is a branch off of this which has a few more. i suggest cutting some of these options, since i believe it's very unlikely to be useful (especially the alternatve `conncomp_algorithm`. the snaphu one is good, the other maybe we should cut).

scottstanie:
also, didn't notice at first, but the bridging needs to be deduplicated (i would pick the whirlwind one like this https://github.com/xhuang-jpl/isce3/commit/bf6aeaeaf40a5ee82de6d88fcaa129520f8faae4 ), or else it runs twice

share/nisar/defaults/insar.yaml
xhuang-jpl: removed the downsample option

xhuang-jpl:
> initial things-
> 
> * do we need to add `whirlwind-insar >= 0.9.0` to somewhere? environment.yml? tools/imagesets/oracle8conda/runtime/environment.yml? the spec file?
>   
>   * related- seems like the `try/except ImportError` at the top is only partially making it an optional dependency

@Tyler-g-hudson can you help with this? Thanks

share/nisar/defaults/insar.yaml
xhuang-jpl: fixed

xhuang-jpl:
Thank you @hfattahi and @scottstanie , I have made the changes based on you comments, However, for adding the whirlwind as a dependency, I might need help from @Tyler-g-hudson since we also need to update the spec.txt in addition to the [environment.yml](https://github.com/xhuang-jpl/isce3/compare/add_whirlwind...scottstanie:isce3:whirlwind-followups#diff-19218fb660c772635155f01a0a22b4933344f48dcfcef5d3b1ef887022459adc)

python/packages/nisar/workflows/unwrap.py
```diff
@@ -313,6 +318,59 @@ def run(cfg: dict, input_hdf5: str, output_hdf5: str):
...
+                    # Get effective number of looks
+                    if whirlwind_cfg['nlooks'] is not None:
+                        nlooks = whirlwind_cfg['nlooks']
+                    else:
+                        rg_spacing = src_h5[f"{src_freq_group_path}/interferogram/slantRangeSpacing"][()]
+                        az_spacing = src_h5[f"{src_freq_group_path}/interferogram/sceneCenterAlongTrackSpacing"][()]
+                        rg_bw = src_h5[f"{src_freq_bandwidth_group_path}/rangeBandwidth"][()]
+                        az_bw = src_h5[f"{src_freq_bandwidth_group_path}/azimuthBandwidth"][()]
+                        nlooks = get_effective_looks(ref_slc, ref_orbit, rg_spacing,
+                                                     az_spacing, rg_bw, az_bw, freq=freq)
```
hfattahi, line 347: Looks like this part is duplicated code compared with existing code for Snaphu option. Would you please try to combine the two. Maybe something like this:

if algorithin in ["anphu", "whirlwind"]:
    code to prep the mask
    if alg == snaphu:
       run snaphu
    else:
        run whirlwind

hfattahi (commented):
I have one more comments for cleaner code and please add whirlwind to CI so that unit test pass.

hfattahi:
To pass CI, please add whirlwind to this environment file https://github.com/isce-framework/isce3/blob/develop/environment.yml

Please also add whirlwind to the docker env here: https://github.com/isce-framework/isce3/blob/develop/tools/imagesets/oracle8conda/runtime/environment.yml

When the PR is merged @jshimada47 may want to update this spec file: https://github.com/isce-framework/isce3/blob/develop/tools/imagesets/oracle8conda/runtime/spec-file.txt

xhuang-jpl:
@scottstanie there is no whirlwind-insar package in the conda-forge channel for the **osx** platform? 

cc: @hfattahi

scottstanie:
do you mean osx, but not arm? 
image
https://anaconda.org/channels/conda-forge/packages/whirlwind-insar/files
i thought they were deprecating that?

xhuang-jpl:
> do you mean osx, but not arm? image https://anaconda.org/channels/conda-forge/packages/whirlwind-insar/files i thought they were deprecating that?

It is the osx-arm64 not the osx-64. The CI failure is because we use the macos-latest runner, and it point at the osx-arm64.  @scottstanie

scottstanie:
> > do you mean osx, but not arm? image https://anaconda.org/channels/conda-forge/packages/whirlwind-insar/files i thought they were deprecating that?
> 
> It is the osx-arm64 not the osx-64. The CI failure is because we use the macos-latest runner, and it point at the osx-arm64. @scottstanie

https://anaconda.org/channels/conda-forge/packages/whirlwind-insar/files version 0.10.0 has it now

scottstanie:
note that 0.10 also added these 2 options which were helping in the fixed PRF cases to have fewer jumps in between the disconnected subswaths: https://github.com/isce-framework/dolphin/pull/742/changes#diff-b4aa92505a629feacbbd6c3d57b0944330f64c51bc19a7274dc7af58aec8457cR45-R46

hfattahi (approved):
Thank you @xhuang-jpl and @scottstanie. LGTM!

bhawkins (approved):

PR #369, opened 2026-08-25
Improvement for TEC correction

seongsujeong:
This PR adds several options to ionospheric TEC correction

- Add option into runconfig to use total TEC only
- Add option to turn on / off the slant range / azimuth correction individually
- Add logic to ignore topside TEC (i.e. use total TEC only) based on the data flag inside IMAGEN TEC
- Apply polynomial fitting to TEC profile to get rid of spikes
- Add the additional TEC profiles to GCOV and InSAR workflows

EDIT
09/10/2026: RANSAC-based outlier detection is implemented and being applied before polynomial fitting. Test is ongoing internally.

### DIFF
--- a/python/packages/isce3/atmosphere/tec_product.py
+++ b/python/packages/isce3/atmosphere/tec_product.py
@@ -2,6 +2,8 @@
 Package to compute TEC LUT from JSON file
 '''
 from datetime import datetime, timedelta
+from itertools import combinations
+from math import comb
 import json
 import os
 
@@ -56,7 +58,10 @@ def _compute_ionospheric_range_delay(utc_time: np.ma.MaskedArray,
         Ellipsoid with same EPSG as DEM interpolator
     total_tec_only: bool
         If True, use total TEC only without subtracting the topside TEC.
-        Otherwise use the suborbital TEC (total TEC minus topside TEC).
+        If False, the topside TEC is evaluated using the corresponding data flag
+        and if all data are valid covering the scenes are found then
+        the suborbital TEC (total TEC minus topside TEC).
+        If not then only total TEC is used.
     polyfit: bool
         If True, fit a polynomial of degree `polyfit_degree` to the suborbital
         TEC profile and use the fitted (smoothed) values. Otherwise use the
@@ -83,6 +88,117 @@ def _compute_ionospheric_range_delay(utc_time: np.ma.MaskedArray,
     delta_r = K * sub_orbital_tec * TECU / center_freq**2 / np.cos(incidence)
 
     return delta_r
+
+
+def _ransac_polyfit(x: np.ndarray, y: np.ndarray, degree: int,
+                    inlier_scale: float=3.0,
+                    max_combinations: int=1000) -> np.ndarray:
+    '''
+    Robustly fit a polynomial to a TEC profile by rejecting noisy samples using
+    an exhaustive RANSAC-style search.
+
+    A polynomial of the given degree is fit to every combination of the minimal
+    number of points (`degree` + 1). For each candidate fit, the consensus set
+    is all samples whose absolute residual falls within an inlier threshold
+    derived from a robust (MAD-based) estimate of the noise scale. The candidate
+    with the largest consensus set is selected (ties broken by smaller sum of
+    squared inlier residuals).
+
+    Because the initial noise scale comes from a full-data fit that the outliers
+    themselves corrupt, the threshold is then recomputed from the consensus set
+    and the search repeated until the consensus set stabilizes. A final
+    polynomial is fit to the resulting consensus set and evaluated at every `x`.
+
+    Parameters
+    ----------
+    x: np.ndarray
+        Sample x-axis (assumed distinct, e.g. sample indices).
+    y: np.ndarray
+        Sample values (the TEC profile) to fit.
+    degree: int
+        Degree of the polynomial to fit.
+    inlier_scale: float
+        Multiplier `k` on the robust noise scale that sets the inlier
+        threshold: `k * 1.4826 * MAD`. Default 3.0.
+    max_combinations: int
+        Upper bound on the number of minimal subsets to enumerate. If the
+        exhaustive search would exceed this, fall back to a plain polynomial
+        fit over all samples. Default 1000.
+
+    Returns
+    -------
+    np.ndarray
+        The fitted (smoothed) values evaluated at every `x`.
+    '''
+    warning_channel = journal.warning("tec_product._ransac_polyfit")
+    info_channel = journal.info("tec_product._ransac_polyfit")
+
+    n = len(y)
+    n_min = degree + 1
+
+    # Not enough points to separate inliers from outliers; plain fit.
+    if n <= n_min:
+        coeffs = np.polyfit(x, y, degree)
+        return np.polyval(coeffs, x)
+
+    # Guard against combinatorial blow-up for very long TEC profiles.
+    if comb(n, n_min) > max_combinations:
+        warning_channel.log(
+            f'Number of minimal subsets C({n}, {n_min}) exceeds '
+            f'max_combinations={max_combinations}; falling back to a plain '
+            'polynomial fit without outlier rejection.')
+        coeffs = np.polyfit(x, y, degree)
+        return np.polyval(coeffs, x)
+
+    def _robust_scale(inliers):
+        # k * 1.4826 * MAD of residuals about the fit over `inliers`.
+        coeffs = np.polyfit(x[inliers], y[inliers], degree)
+        resid = (y - np.polyval(coeffs, x))[inliers]
+        mad = np.median(np.abs(resid - np.median(resid)))
+        return inlier_scale * 1.4826 * mad
+
+    def _largest_consensus(threshold):
+        # Exhaustive search over minimal subsets for the largest consensus set.
+        best_inliers = None
+        best_count = -1
+        best_ssr = np.inf
+        for combo in combinations(range(n), n_min):
+            combo = list(combo)
+            coeffs = np.polyfit(x[combo], y[combo], degree)
+            resid = np.abs(y - np.polyval(coeffs, x))
+            inliers = resid <= threshold
+            count = int(np.count_nonzero(inliers))
+            ssr = float(np.sum(resid[inliers] ** 2))
+            if count > best_count or (count == best_count and ssr < best_ssr):
+                best_count = count
+                best_ssr = ssr
+                best_inliers = inliers
+        return best_inliers
+
+    # Initial noise scale from a full-data fit.
+    threshold = _robust_scale(np.ones(n, dtype=bool))
+
+    # Degenerate noise scale (near-perfect fit); nothing to reject.
+    if not np.isfinite(threshold) or threshold <= 0:
+        coeffs = np.polyfit(x, y, degree)
+        return np.polyval(coeffs, x)
+
+    # Iterate: the full-data threshold is inflated by the outliers, so recompute
+    # the scale from the (cleaner) consensus set and reselect until it settles.
+    inliers = _largest_consensus(threshold)
+    for _ in range(5):
+        new_threshold = _robust_scale(inliers)
+        if not np.isfinite(new_threshold) or new_threshold <= 0:
+            break
+        new_inliers = _largest_consensus(new_threshold)
+        if np.array_equal(new_inliers, inliers):
+            break
+        inliers = new_inliers
+
+    # Final fit on the largest consensus set.
+    coeffs = np.polyfit(x[inliers], y[inliers], degree)
+    info_channel.log('RANSAC outlier detection and polynimial fitting completed.')
+    return np.polyval(coeffs, x)
 
 
 def _get_suborbital_tec(tec_json_dict: dict,
@@ -130,10 +246,10 @@ def _get_suborbital_tec(tec_json_dict: dict,
     sub_orbital_tec = sub_orbital_tec[~tec_time_mask]
 
     if polyfit:
-        # Fit a polynomial over the TEC profile to smooth out noise.
+        # Fit a polynomial over the TEC profile to smooth out noise, rejecting
+        # noisy samples first via an exhaustive RANSAC-style search.
         x = np.arange(len(sub_orbital_tec))
-        coeffs = np.polyfit(x, sub_orbital_tec, polyfit_degree)
-        sub_orbital_tec = np.polyval(coeffs, x)
+        sub_orbital_tec = _ransac_polyfit(x, sub_orbital_tec, polyfit_degree)
 
     return sub_orbital_tec
 
@@ -422,6 +538,9 @@ def _get_tec_time(tec_json_dict: dict,
     _check_tec_grid_contains_radargrid(radar_grid, t_since_ref_epoch,
                                        staggered_tec_grid)
 
+    # check if the TEC time grids are equally spaced
+    _check_tec_grid_equally_spaced(t_since_ref_epoch)
+
     return t_since_ref_epoch
 
 
@@ -474,6 +593,37 @@ def _check_tec_grid_contains_radargrid(radar_grid: isce3.product.RadarGridParame
                f'Relative timing w.r.t. Sensing start:\ntec_start={tec_t[0] - radar_grid.sensing_start}, '
                f'tec_end={tec_t[-1] - radar_grid.sensing_start}\n'
                f'radargrid start={0}, radargrid_stop={radar_grid.sensing_stop - radar_grid.sensing_start}')
+
+    error_channel.log(err_msg)
+    raise ValueError(err_msg)
+
+
+def _check_tec_grid_equally_spaced(t_since_ref_epoch: np.ma.MaskedArray) -> None:
+    '''
+    Helper function to check if the valid TEC time grid is equally spaced.
+
+    Parameters
+    ----------
+    t_since_ref_epoch: np.ma.MaskedArray
+        Masked array of the TEC time grid in seconds since reference epoch.
+
+    Raises
+    ------
+    ValueError: When the valid TEC time grid is not equally spaced.
+    '''
+    tec_t = t_since_ref_epoch.compressed()
+
+    spacings = np.diff(tec_t)
+    mean_spacing = spacings.mean()
+
+    if np.allclose(spacings, mean_spacing):
+        return
+
+    error_channel = journal.error(
+        "tec_product._check_tec_grid_equally_spaced")
+    err_msg = ('TEC time grid is not equally spaced.\n'
+               f'mean spacing={mean_spacing}\n'
+               f'min spacing={spacings.min()}, max spacing={spacings.max()}')
 
     error_channel.log(err_msg)
     raise ValueError(err_msg)
--- a/python/packages/isce3/atmosphere/tec_product.py
+++ b/python/packages/isce3/atmosphere/tec_product.py
@@ -22,7 +22,8 @@ def _compute_ionospheric_range_delay(utc_time: np.ma.MaskedArray,
                                      doppler_lut: isce3.core.LUT2d,
                                      radar_grid: isce3.product.RadarGridParameters,
                                      dem_interp: isce3.geometry.DEMInterpolator,
-                                     ellipsoid: isce3.core.Ellipsoid) -> np.ndarray:
+                                     ellipsoid: isce3.core.Ellipsoid,
+                                     total_tec_only: bool=False) -> np.ndarray:
     '''
     Compute near or far TEC delta range
 
@@ -51,6 +52,9 @@ def _compute_ionospheric_range_delay(utc_time: np.ma.MaskedArray,
         Digital elevation model, m above ellipsoid. Defaults to h=0.
     ellipsoid: isce3.core.Ellipsoid
         Ellipsoid with same EPSG as DEM interpolator
+    total_tec_only: bool
+        If True, use total TEC only without subtracting the topside TEC.
+        Otherwise use the suborbital TEC (total TEC minus topside TEC).
 
     Returns
     -------
@@ -59,7 +63,8 @@ def _compute_ionospheric_range_delay(utc_time: np.ma.MaskedArray,
     '''
     # compute sub orbital TEC from total and top TEC in JSON
     sub_orbital_tec = _get_suborbital_tec(tec_json_dict, nr_fr,
-                                          utc_time.mask)
+                                          utc_time.mask,
+                                          total_tec_only=total_tec_only)
 
     incidence = [compute_incidence_angle(t, nr_fr_rg, orbit, doppler_lut,
                                          radar_grid, dem_interp, ellipsoid)
@@ -72,7 +77,10 @@ def _compute_ionospheric_range_delay(utc_time: np.ma.MaskedArray,
 
 def _get_suborbital_tec(tec_json_dict: dict,
                         nr_fr: str,
-                        tec_time_mask: np.ndarray) -> np.ndarray:
+                        tec_time_mask: np.ndarray,
+                        total_tec_only: bool=False,
+                        polyfit: bool=False,
+                        polyfit_degree: int=2) -> np.ndarray:
     '''
     Get the suborbital TEC from IMAGEN TEC product parsed as a dictionary by
     subtracting the total TEC by top (i.e. above the satellite) TEC
@@ -87,6 +95,15 @@ def _get_suborbital_tec(tec_json_dict: dict,
         Mask of TEC values that fall within radar grid or orbit time span. Mask
         follows NumPy masked array convention where True is masked and False is
         not.
+    total_tec_only: bool
+            If True, use total TEC only without subtracting the topside TEC.
+            Otherwise use the suborbital TEC (total TEC minus topside TEC).
+    polyfit: bool
+        If True, fit a polynomial of degree `polyfit_degree` to the suborbital
+        TEC profile and return the fitted (smoothed) values. Otherwise return
+        the suborbital TEC as is.
+    polyfit_degree: int
+        Degree of the polynomial fit applied when `polyfit` is True. Default 2.
 
     Returns
     -------
@@ -96,8 +113,17 @@ def _get_suborbital_tec(tec_json_dict: dict,
     # compute sub orbital TEC from total and top TEC in JSON
     tot_tec = np.array(tec_json_dict[f'totTec{nr_fr}'])
     top_tec = np.array(tec_json_dict[f'topTec{nr_fr}'])
-    sub_orbital_tec = tot_tec - top_tec
+    if total_tec_only:
+        sub_orbital_tec = tot_tec
+    else:
+        sub_orbital_tec = tot_tec - top_tec
     sub_orbital_tec = sub_orbital_tec[~tec_time_mask]
+
+    if polyfit:
+        # Fit a polynomial over the TEC profile to smooth out noise.
+        x = np.arange(len(sub_orbital_tec))
+        coeffs = np.polyfit(x, sub_orbital_tec, polyfit_degree)
+        sub_orbital_tec = np.polyval(coeffs, x)
 
     return sub_orbital_tec
 
@@ -106,7 +132,8 @@ def tec_lut2d_from_json_srg(json_path: str, center_freq: float,
                             orbit: isce3.core.Orbit,
                             radar_grid: isce3.product.RadarGridParameters,
                             doppler_lut: isce3.core.LUT2d, dem_path: str,
-                            margin: float=40.0) -> isce3.core.LUT2d:
+                            margin: float=40.0,
+                            total_tec_only: bool=False) -> isce3.core.LUT2d:
     '''
     Create a TEC LUT2d for slant range correction from a JSON source
 
@@ -127,6 +154,9 @@ def tec_lut2d_from_json_srg(json_path: str, center_freq: float,
     margin: float
         Margin (seconds) to pad to sensing start and stop times when extracting
         TEC data. Default 40 seconds.
+    total_tec_only: bool
+        If True, use total TEC only without subtracting the topside TEC.
+        Otherwise use the suborbital TEC (total TEC minus topside TEC).
 
     Returns
     -------
@@ -172,7 +202,8 @@ def tec_lut2d_from_json_srg(json_path: str, center_freq: float,
                                                           doppler_lut,
                                                           radar_grid,
                                                           dem_interp,
-                                                          ellipsoid)
+                                                          ellipsoid,
+                                                          total_tec_only)
                          for nr_fr, rg in zip(['Nr', 'Fr'], rg_vec)]).T
 
     return isce3.core.LUT2d(rg_vec, t_since_epoch_masked.compressed(), delta_r)
@@ -181,7 +212,10 @@ def tec_lut2d_from_json_srg(json_path: str, center_freq: float,
 def tec_lut2d_from_json_az(json_path: str, center_freq: float,
                            orbit: isce3.core.Orbit,
                            radar_grid: isce3.product.RadarGridParameters,
-                           margin: float=40.0) -> isce3.core.LUT2d:
+                           margin: float=40.0,
+                           total_tec_only: bool=False,
+                           polyfit: bool=False,
+                           polyfit_degree: int=2) -> isce3.core.LUT2d:
     '''
     Create a TEC LUT2d for azimuth time correction from a JSON source
 
@@ -198,6 +232,15 @@ def tec_lut2d_from_json_az(json_path: str, center_freq: float,
     margin: float
         Margin (seconds) to pad to sensing start and stop times when extracting
         TEC data. Default 40 seconds.
+    total_tec_only: bool
+        If True, use total TEC only without subtracting the topside TEC.
+        Otherwise use the suborbital TEC (total TEC minus topside TEC).
+    polyfit: bool
+        If True, fit a polynomial of degree `polyfit_degree` to the suborbital
+        TEC profile before computing the azimuth gradient. Otherwise use the
+        suborbital TEC as is.
+    polyfit_degree: int
+        Degree of the polynomial fit applied when `polyfit` is True. Default 2.
 
     Returns
     -------
@@ -229,7 +272,10 @@ def tec_lut2d_from_json_az(json_path: str, center_freq: float,
     # Transpose stacked output to get shape to be consistent with coordinates
     tec_suborbital = np.vstack([_get_suborbital_tec(tec_json_dict,
                                                     nr_fr,
-                                                    t_since_epoch_masked.mask)
+                                                    t_since_epoch_masked.mask,
+                                                    total_tec_only=total_tec_only,
+                                                    polyfit=polyfit,
+                                                    polyfit_degree=polyfit_degree)
                                 for nr_fr in ['Nr', 'Fr']]).T
 
     # set up up the LUT grids for az. iono. delay
@@ -299,7 +345,7 @@ def _get_tec_time(tec_json_dict: dict,
     # Get string UTC times from JSON as isce3.core.DateTime objects.
     json_utc_datetimes = [isce3.core.DateTime(iso_t_str)
                           for iso_t_str in tec_json_dict['utc']]
-    
+
     # Adjust the radar grid margin in case of the staggered grid.
     # The staggered grid needs at least half of the TEC spacing at each side.
     # When `margin` is not big enough, then increase it to half the spacing.
@@ -352,7 +398,7 @@ def _get_tec_time(tec_json_dict: dict,
 
     t_since_ref_epoch = np.ma.MaskedArray(data=t_since_ref_epoch,
                                           mask=np.logical_not(time_mask))
-    
+
     _check_tec_grid_contains_radargrid(radar_grid, t_since_ref_epoch,
                                        staggered_tec_grid)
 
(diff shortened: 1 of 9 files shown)


share/nisar/schemas/gslc.yaml
```diff
@@ -436,3 +438,10 @@ output_options:
 lut_options:
     # Boolean flag to activate/deactivate model-based solid earth tide corrections
     solid_earth_tides_enabled: bool(required=False)
+
+tec_correction_options:
+    apply_slant_range_correction: bool(required=False)
+    apply_azimuth_correction: bool(required=False)
+    use_total_tec_only: bool(required=False)
+    ignore_low_quality_topside_tec: bool(required=False)
```
hfattahi: how about:
```suggestion
    ignore_invalid_topside_tec: bool(required=False)
```

share/nisar/schemas/gcov.yaml
```diff
@@ -555,6 +558,11 @@ geocode_options:
     # (polarizations) to process.
     max_block_size: num(min=0, required=False)
 
+tec_correction_options:
+    use_total_tec_only: bool(required=False)
+    ignore_low_quality_topside_tec: bool(required=False)
```
hfattahi:
```suggestion
    ignore_invalid_topside_tec: bool(required=False)
```

share/nisar/schemas/insar.yaml
```diff
@@ -1235,6 +1237,13 @@ lut_options:
     # Boolean flag to activate/deactivate model-based solid earth tide corrections
     solid_earth_tides_enabled: bool(required=False)
 
+tec_correction_options:
+    apply_slant_range_correction: bool(required=False)
+    apply_azimuth_correction: bool(required=False)
+    use_total_tec_only: bool(required=False)
+    ignore_low_quality_topside_tec: bool(required=False)
```
hfattahi:
```suggestion
    ignore_invalid_topside_tec: bool(required=False)
```

share/nisar/schemas/gcov.yaml
```diff
@@ -555,6 +558,11 @@ geocode_options:
...
+    polyfit_tec_profile: bool(required=False)
+
```
hfattahi, line 565: Why gcov does not have extra options for choosing between range and azimuth correction.

share/nisar/defaults/gcov.yaml
```diff
@@ -277,6 +277,13 @@ runconfig:
                 #     'algorithm_type' equal to 'area_projection'
                 area_beta_mode: auto
 
+            # TEC correction options. The apply-range/azimuth ionospheric delay
+            # flags live under the `geocode` group above.
+            tec_correction:
+                use_total_tec_only: False
+                ignore_low_quality_topside_tec: False
```
hfattahi: Let's make default True and let's please make sure it is turned on clearly in PSF for SDS.

```suggestion
                ignore_low_quality_topside_tec: True
```

share/nisar/defaults/gslc.yaml
```diff
@@ -331,6 +331,13 @@ runconfig:
                 # Boolean flag to activate/deactivate model-based solid earth tide corrections
                 solid_earth_tides_enabled: True
 
+            tec_correction:
+                apply_slant_range_correction: True
+                apply_azimuth_correction: True
+                use_total_tec_only: False
+                ignore_low_quality_topside_tec: False
```
hfattahi:
```suggestion
                ignore_low_quality_topside_tec: True
```

share/nisar/defaults/insar.yaml
```diff
@@ -830,6 +830,13 @@ runconfig:
                 # Boolean flag to activate/deactivate model-based solid earth tide corrections
                 solid_earth_tides_enabled: True
 
+            tec_correction:
+                apply_slant_range_correction: True
+                apply_azimuth_correction: True
+                use_total_tec_only: False
+                ignore_low_quality_topside_tec: False
```
hfattahi:
```suggestion
                ignore_low_quality_topside_tec: True
```

python/packages/nisar/workflows/geocode_corrections.py
```diff
@@ -166,13 +172,27 @@ def _get_iono_azimuth_corrections(cfg, slc, frequency, orbit):
...
+                                            total_tec_only=use_totaltec_only,
+                                            polyfit=polyfit_tec_profile)
+
+    if cfg['processing']['tec_correction']['apply_azimuth_correction'] is False:
+        # NOTE: Just returning an empty LUT2d() will disrupt the downstream procedure,
+        # especially comparing the LUT's axis with the RSLC's radargrid
+        zero_arr = np.zeros(tec_correction.data.shape)
+        tec_correction = isce3.core.LUT2d(tec_correction.x_axis,
+                                          tec_correction.y_axis,
+                                          zero_arr)
```
hfattahi, line 189: @gshiroma brought up Is it safer if we would just use isce3.core.LUT2d() without limiting extents and turn off error boundary check?

python/packages/nisar/workflows/geocode_corrections.py
```diff
@@ -208,12 +231,103 @@ def _get_iono_srange_corrections(cfg, slc, frequency, orbit):
...
-                                             radar_grid, doppler, dem_file)
+                                             radar_grid, doppler, dem_file,
+                                             total_tec_only=use_totaltec_only,
+                                             polyfit=polyfit_tec_profile)
+
+    if cfg['processing']['tec_correction']['apply_slant_range_correction'] is False:
+        # NOTE: Just returning an empty LUT2d() will disrupt the downstream procedure,
+        # especially comparing the LUT's axis with the RSLC's radargrid
+        zero_arr = np.zeros(tec_correction.data.shape)
+        tec_correction = isce3.core.LUT2d(tec_correction.x_axis,
```
hfattahi, line 246: Same comment as above.

python/packages/nisar/workflows/geocode_corrections.py
```diff
@@ -208,12 +231,103 @@ def _get_iono_srange_corrections(cfg, slc, frequency, orbit):
...
+    warning_channel = journal.warning("geocode_corrections.should_use_only_total_tec")
+
+    tec_path = cfg['dynamic_ancillary_file_group']['tec_file']
+
+    if tec_path is None:
+        warning_channel.log(f'TEC path was not provided: {tec_path}')
+        return False
+
+    if not os.path.exists(tec_path):
+        warning_channel.log(f'TEC path provided does not exist: {tec_path}')
```
hfattahi: raise an exception.

python/packages/nisar/workflows/geocode_corrections.py
```diff
@@ -208,12 +231,103 @@ def _get_iono_srange_corrections(cfg, slc, frequency, orbit):
...
+    # For compatibility, skip the check if `topTecNrDataFlag` or `topTecFrDataFlag`
+    # do not exist.
+    required_keys = {"topTecNrDataFlag", "topTecFrDataFlag"}
+
+    if not required_keys.issubset(tec_dict):
+        warning_channel.log(f'TopTEC Dataflag do not exist in IMAGEN TEC file: {tec_path}')
+        return False
+
+    # Start the investigation.
+    if cfg['processing']['tec_correction']['ignore_low_quality_topside_tec'] is True:
```
hfattahi:
```suggestion
    if cfg['processing']['tec_correction']['ignore_invalid_topside_tec']:
```

python/packages/nisar/workflows/geocode_corrections.py
```diff
@@ -208,12 +231,103 @@ def _get_iono_srange_corrections(cfg, slc, frequency, orbit):
...
+        within_radar_grid = ((tec_t_since_epoch >= az_start) &
+                             (tec_t_since_epoch <= az_stop))
+
+        top_tec_flags = np.concatenate(
+            [np.array(tec_dict['topTecNrDataFlag'])[within_radar_grid],
+             np.array(tec_dict['topTecFrDataFlag'])[within_radar_grid]])
+
+        # A flag value of 1 marks a low-quality topside TEC sample. If any
+        # exist within the radar grid, fall back to total TEC only.
+        if np.any(top_tec_flags == 1):
```
hfattahi:
```suggestion
        if np.any(top_tec_flags):
```

python/packages/isce3/atmosphere/tec_product.py
```diff
@@ -51,6 +54,15 @@ def _compute_ionospheric_range_delay(utc_time: np.ma.MaskedArray,
         Digital elevation model, m above ellipsoid. Defaults to h=0.
     ellipsoid: isce3.core.Ellipsoid
         Ellipsoid with same EPSG as DEM interpolator
+    total_tec_only: bool
+        If True, use total TEC only without subtracting the topside TEC.
+        Otherwise use the suborbital TEC (total TEC minus topside TEC).
```
hfattahi: Do you want to be more specific that actually is False, the topside TEC quality is evaluated and if all valid samples covering the scene are found, then suborbital TEC is used. If not then total TEC is used. 
```suggestion
        Otherwise the topside TEC quality is evaluated and if all valid samples covering the scene are found, then suborbital TEC is used. If not then total TEC is used.
```
or something like that

python/packages/nisar/workflows/geocode_corrections.py
```diff
@@ -208,12 +231,103 @@ def _get_iono_srange_corrections(cfg, slc, frequency, orbit):
...
+        # Convert the TEC UTC times to seconds since the radar grid reference
+        # epoch so they share the same time base as the azimuth time span.
+        tec_t_since_epoch = np.array(
+            [(isce3.core.DateTime(t_str) - ref_epoch).total_seconds()
+             for t_str in tec_dict['utc']])
+
+        # Crop the topside TEC data flags to the TEC samples that fall within
+        # the azimuth time span of the (decimated) radar grid.
+        within_radar_grid = ((tec_t_since_epoch >= az_start) &
+                             (tec_t_since_epoch <= az_stop))
```
rad-eng-59: Given times are monotonically increasing (sorted but no need to be uniform), we can avoid unnecessary checks
```suggestion
        i_start = np.searchsorted(tec_t_since_epoch, az_start, side='left')
        i_stop = np.searchsorted(tec_t_since_epoch, az_stop, side='right')
        within_radar_grid = slice(i_start, i_stop)
```

EDIT:
If stop is excludive point then use the following:
```python
i_start_stop = np.searchsorted(tec_t_since_epoch, (az_start, az_stop))
i_start_stop[1] -= 1
within_radar_grid = slice(*i_start_stop)
```

python/packages/nisar/workflows/geocode_corrections.py
rad-eng-59: Perhaps a bit more pythonic
```suggestion
        if np.any(top_tec_flags):
```

Note that if the mask array `top_tec_flags` is empty, the `numpy.any` will fail! This is why it would be good to raise an exception  earlier of the start/stop AZ time is out of TEC validity range.

python/packages/nisar/workflows/geocode_corrections.py
```diff
@@ -208,12 +231,103 @@ def _get_iono_srange_corrections(cfg, slc, frequency, orbit):
...
+
+    tec_path = cfg['dynamic_ancillary_file_group']['tec_file']
+
+    if tec_path is None:
+        warning_channel.log(f'TEC path was not provided: {tec_path}')
+        return False
+
+    if not os.path.exists(tec_path):
+        warning_channel.log(f'TEC path provided does not exist: {tec_path}')
+        return False
```
rad-eng-59: Should not this be checked at the higher level or at least raise exception? If we want to make this function more independent then we should also check the input `[az_start, az_stop]` against TEC start/end valid datetime below otherwise we get empty valid mask and "np.any" will raise an error.

hfattahi (commented):
some comments up to what we reviewed today.

python/packages/isce3/atmosphere/tec_product.py
```diff
@@ -96,17 +123,29 @@ def _get_suborbital_tec(tec_json_dict: dict,
...
-    sub_orbital_tec = tot_tec - top_tec
+    if total_tec_only:
+        sub_orbital_tec = tot_tec
+    else:
+        sub_orbital_tec = tot_tec - top_tec
     sub_orbital_tec = sub_orbital_tec[~tec_time_mask]
 
+    if polyfit:
+        # Fit a polynomial over the TEC profile to smooth out noise.
+        x = np.arange(len(sub_orbital_tec))
```
bhawkins, line 251: Might be better to use the time-axis of the TEC data in case there are missing samples or something. Or maybe it's already been resampled to a uniform grid at this point?

python/packages/isce3/atmosphere/tec_product.py
```diff
@@ -172,7 +220,10 @@ def tec_lut2d_from_json_srg(json_path: str, center_freq: float,
...
                                                           radar_grid,
                                                           dem_interp,
-                                                          ellipsoid)
+                                                          ellipsoid,
+                                                          total_tec_only,
+                                                          polyfit,
+                                                          polyfit_degree)
                          for nr_fr, rg in zip(['Nr', 'Fr'], rg_vec)]).T
 
     return isce3.core.LUT2d(rg_vec, t_since_epoch_masked.compressed(), delta_r)
```
bhawkins, line 345: It seems like this runs into trouble if the time spacing is unequal or there are missing data. Maybe it'd be a good idea to add an explicit check somewhere to validate that assumption.

bhawkins (commented):
Polynomial fitting might  not be robust to outliers, so another approach like median filter or RANSAC algorithm might be better.

@gshiroma suggests checking that the related output layers (az and rg corrections, but not TEC, which get geocoded) in the GCOV product still make sense in all cases.

seongsujeong:
> @gshiroma suggests checking that the related output layers (az and rg corrections, but not TEC, which get geocoded) in the GCOV product still make sense in all cases.

In GCOV, disabling either correction (i.e., the azimuth or slant-range correction) causes the corresponding correction LUT to be omitted from the output. @gshiroma Please confirm if this is expected behavior.

python/packages/nisar/workflows/geocode_corrections.py
seongsujeong: Sound good. Let me take your suggestion.

> i_start_stop[1] -= 1

I would rather not manipulate the indices (`i_start`, `i_stop`) as you suggested in the edit because I cannot guarantee that they correspond to the leftmost and rightmost elements in the array.

python/packages/nisar/workflows/geocode_corrections.py
seongsujeong: Good point. The validity of TEC products with respect to the radargrid is checked in the helper function below:
https://github.com/isce-framework/isce3/blob/67bccb0ce9245d2f7f4bbf87816d412e3746a634/python/packages/isce3/atmosphere/tec_product.py#L362

share/nisar/schemas/gcov.yaml
seongsujeong: GCOV already has a flag to control the azimuth/slant range correction independently. Below is the snippet of GSLC default runconfig:

https://github.com/isce-framework/isce3/blob/67bccb0ce9245d2f7f4bbf87816d412e3746a634/share/nisar/defaults/gcov.yaml#L304-L312

python/packages/isce3/atmosphere/tec_product.py
seongsujeong: Added a checker `_check_tec_grid_equally_spaced()` in the code. The function will raise an error when any missing sample is found. Therefore the equal-sample interval is guaranteed at this point.

python/packages/isce3/atmosphere/tec_product.py
seongsujeong: Same as above.

python/packages/isce3/atmosphere/tec_product.py
```diff
@@ -70,9 +90,123 @@ def _compute_ionospheric_range_delay(utc_time: np.ma.MaskedArray,
...
+    # Guard against combinatorial blow-up for very long TEC profiles.
+    if comb(n, n_min) > max_combinations:
+        warning_channel.log(
+            f'Number of minimal subsets C({n}, {n_min}) exceeds '
+            f'max_combinations={max_combinations}; falling back to a plain '
+            'polynomial fit without outlier rejection.')
+        coeffs = np.polyfit(x, y, degree)
+        return np.polyval(coeffs, x)
+
+    def _robust_scale(inliers):
```
rad-eng-59, line 153: FYI, below scipy will compute MAD for you with normal distribution scalar:

```python
return inlier_scale * scipy.stats.median_abs_deviation(resid, scale='normal')
```

python/packages/isce3/atmosphere/tec_product.py
```diff
@@ -70,9 +90,123 @@ def _compute_ionospheric_range_delay(utc_time: np.ma.MaskedArray,
     return delta_r
 
 
+def _ransac_polyfit(x: np.ndarray, y: np.ndarray, degree: int,
```
rad-eng-59, line 93: IMO, this is overkill. You could have just use MAD with either unity scalar (strict filtering) or factor of 2 scalar (looser filtering) to remove outliers from your data population prior to polyfitting for a certain degree.
It seems like it is assumed that the noise in measured TEC values has normal distribution by default and thus outliers are treated as tails beyond `3xSigma`. Not sure whether this is correct assumption about TEC measurements and their source of error given small number of samples.

python/packages/isce3/atmosphere/tec_product.py
```diff
@@ -70,9 +90,123 @@ def _compute_ionospheric_range_delay(utc_time: np.ma.MaskedArray,
...
+    ----------
+    x: np.ndarray
+        Sample x-axis (assumed distinct, e.g. sample indices).
+    y: np.ndarray
+        Sample values (the TEC profile) to fit.
+    degree: int
+        Degree of the polynomial to fit.
+    inlier_scale: float
+        Multiplier `k` on the robust noise scale that sets the inlier
+        threshold: `k * 1.4826 * MAD`. Default 3.0.
```
rad-eng-59, line 122: As  minor suggestion, It would be good to mention that the default scalar is based on `3xSigma` of normal distribution. It took me a while to figure that out.

rad-eng-59 (approved):
LGTM. Thanks for the updates. Just some minor comments.

python/packages/nisar/workflows/geocode_corrections.py
seongsujeong: Good point. For the TEC product existence check, I added the routine in the commit below:
https://github.com/isce-framework/isce3/pull/369/changes/eea2ecf30bea6e7e81ca76fa43fc9a8c598a7992

Regarding the temporary coverage check we already have that in place:
https://github.com/isce-framework/isce3/blob/355f6b35519f4172ae6a24b90bf5a4ce14a34b7c/python/packages/isce3/atmosphere/tec_product.py#L362-L364

python/packages/isce3/atmosphere/tec_product.py
seongsujeong: After testing and tuning the algorithm, it look like MAD-based filtering gives better results. I changed the implementation to MAD-based.

python/packages/isce3/atmosphere/tec_product.py
seongsujeong: Equation for computing MAD is simple, and I just wanted to minimize the modules to import to the `tec_product.py`

python/packages/isce3/atmosphere/tec_product.py
seongsujeong: Changed `k` into `sigma` to better reveal its meaning.

python/packages/nisar/workflows/geocode_corrections.py
bhawkins:
```suggestion
                                          zero_arr, b_error=False)
```
Like @hfattahi mentioned, this would prevent an error occurring when our dummy extent isn't wide enough for whatever reason, replicating the behavior of `LUT2d()` (no arg ctor)

python/packages/isce3/atmosphere/tec_product.py
```diff
@@ -70,9 +88,68 @@ def _compute_ionospheric_range_delay(utc_time: np.ma.MaskedArray,
...
+
+    Parameters
+    ----------
+    x: np.ndarray
+        Sample x-axis (assumed distinct, e.g. sample indices).
+    y: np.ndarray
+        Sample values (the TEC profile) to fit.
+    degree: int
+        Degree of the polynomial to fit.
+    sigma: float
```
bhawkins:
```suggestion
    num_sigma: float
```
Name is a little confusing since it's a sigma multiplier (unitless) rather than something with the units of the data.

python/packages/isce3/atmosphere/tec_product.py
```diff
@@ -70,9 +88,68 @@ def _compute_ionospheric_range_delay(utc_time: np.ma.MaskedArray,
...
+
+    # Robust noise scale: k * 1.4826 * MAD of the residuals.
+    mad = np.median(np.abs(resid - np.median(resid)))
+    threshold = sigma * mad
+    info_channel.log(f'TEC MAD threshold={threshold}')
+
+    # Degenerate noise scale (near-perfect fit); nothing to reject.
+    if not np.isfinite(threshold) or threshold <= 0:
+        info_channel.log('TEC MAD Threshold is not valid. Keeping all TEC samples for polynomial fitting.')
+        return np.polyval(coeffs, x)
```
bhawkins, line 137:
```suggestion
        raise RuntimeError()
```
If we get here I'm not sure it's safe to evaluate the polynomial. So I'd suggest either `return y` or just throw an exception.

python/packages/isce3/atmosphere/tec_product.py
```diff
@@ -70,9 +88,68 @@ def _compute_ionospheric_range_delay(utc_time: np.ma.MaskedArray,
     return delta_r
 
 
+def _mad_polyfit(x: np.ndarray, y: np.ndarray, degree: int,
+                 sigma: float=1.5) -> np.ndarray:
+    '''
+    Robustly fit a polynomial to a TEC profile by rejecting noisy samples using
+    a MAD-based threshold.
```
gshiroma: Please, introduced the acronym MAD (Median Absolute Deviation).

python/packages/isce3/atmosphere/tec_product.py
```diff
@@ -70,9 +88,68 @@ def _compute_ionospheric_range_delay(utc_time: np.ma.MaskedArray,
...
+    # Degenerate noise scale (near-perfect fit); nothing to reject.
+    if not np.isfinite(threshold) or threshold <= 0:
+        info_channel.log('TEC MAD Threshold is not valid. Keeping all TEC samples for polynomial fitting.')
+        return np.polyval(coeffs, x)
+
+    # Keep only inliers, but fall back to the plain fit if too few survive.
+    inliers = np.abs(resid) <= threshold
+    if np.count_nonzero(inliers) <= degree:
+        info_channel.log('Not enough iners from TEC sample. Keeping all TEC samples for polynomial fitting.')
+        return np.polyval(coeffs, x)
```
bhawkins, line 143:
```suggestion
        return y
```
This condition suggests the fit is bad, so maybe it's better not to apply polynomial smoothing and just return the data.  Should mention the behavior in the docstring.

python/packages/isce3/atmosphere/tec_product.py
```diff
@@ -70,9 +88,68 @@ def _compute_ionospheric_range_delay(utc_time: np.ma.MaskedArray,
...
+    np.ndarray
+        The fitted (smoothed) values evaluated at every `x`.
+    '''
+    info_channel = journal.info("tec_product._mad_polyfit")
+
+    # Initial fit over all samples and its residuals.
+    coeffs = np.polyfit(x, y, degree)
+    resid = y - np.polyval(coeffs, x)
+
+    # Robust noise scale: k * 1.4826 * MAD of the residuals.
```
gshiroma: What's this "k * 1.4826"?

python/packages/isce3/atmosphere/tec_product.py
```diff
@@ -70,9 +88,68 @@ def _compute_ionospheric_range_delay(utc_time: np.ma.MaskedArray,
...
+
+    # Degenerate noise scale (near-perfect fit); nothing to reject.
+    if not np.isfinite(threshold) or threshold <= 0:
+        info_channel.log('TEC MAD Threshold is not valid. Keeping all TEC samples for polynomial fitting.')
+        return np.polyval(coeffs, x)
+
+    # Keep only inliers, but fall back to the plain fit if too few survive.
+    inliers = np.abs(resid) <= threshold
+    if np.count_nonzero(inliers) <= degree:
+        info_channel.log('Not enough iners from TEC sample. Keeping all TEC samples for polynomial fitting.')
```
gshiroma: "iners"? Should it be "inliers"?

python/packages/isce3/atmosphere/tec_product.py
```diff
@@ -70,9 +88,68 @@ def _compute_ionospheric_range_delay(utc_time: np.ma.MaskedArray,
     return delta_r
 
 
+def _mad_polyfit(x: np.ndarray, y: np.ndarray, degree: int,
```
bhawkins:
```suggestion
def _smooth_mad_polyfit(x: np.ndarray, y: np.ndarray, degree: int,
```
The function doesn't return fit coefficients like `numpy.polyfit`, so I think it's clearer if the name is more different. The function performs smoothing, so `_smooth_mad_polyfit` makes more sense to me, at least.

python/packages/isce3/atmosphere/tec_product.py
```diff
@@ -70,9 +88,68 @@ def _compute_ionospheric_range_delay(utc_time: np.ma.MaskedArray,
...
+    info_channel.log(f'TEC MAD threshold={threshold}')
+
+    # Degenerate noise scale (near-perfect fit); nothing to reject.
+    if not np.isfinite(threshold) or threshold <= 0:
+        info_channel.log('TEC MAD Threshold is not valid. Keeping all TEC samples for polynomial fitting.')
+        return np.polyval(coeffs, x)
+
+    # Keep only inliers, but fall back to the plain fit if too few survive.
+    inliers = np.abs(resid) <= threshold
+    if np.count_nonzero(inliers) <= degree:
```
gshiroma, line 141: Shouldn't it be "degrees + 1" instead? A polynomial with degree-n should require at least n + 1 samples.

python/packages/isce3/atmosphere/tec_product.py
```diff
@@ -70,9 +88,68 @@ def _compute_ionospheric_range_delay(utc_time: np.ma.MaskedArray,
...
+    ----------
+    x: np.ndarray
+        Sample x-axis (assumed distinct, e.g. sample indices).
+    y: np.ndarray
+        Sample values (the TEC profile) to fit.
+    degree: int
+        Degree of the polynomial to fit.
+    sigma: float
+        Multiplier `k` on the robust noise scale that sets the inlier
+        threshold: `k * MAD`. Default 1.5.
```
gshiroma: `k` non longer needed?

share/nisar/defaults/gcov.yaml
```diff
@@ -277,6 +277,13 @@ runconfig:
                 #     'algorithm_type' equal to 'area_projection'
                 area_beta_mode: auto
 
+            # TEC correction options. The apply-range/azimuth ionospheric delay
+            # flags live under the `geocode` group above.
+            tec_correction:
+                use_total_tec_only: False
+                ignore_invalid_topside_tec: True
+                polyfit_tec_profile: False
```
bhawkins, line 285:
```suggestion
                polyfit_tec_profile: False
                num_sigma: 1.5
```
Might be a good idea to make the outlier check tunable.

share/nisar/defaults/gslc.yaml
```diff
@@ -331,6 +331,13 @@ runconfig:
                 # Boolean flag to activate/deactivate model-based solid earth tide corrections
                 solid_earth_tides_enabled: True
 
+            tec_correction:
+                apply_slant_range_correction: True
+                apply_azimuth_correction: True
+                use_total_tec_only: False
+                ignore_invalid_topside_tec: True
+                polyfit_tec_profile: False
```
bhawkins, line 339:
```suggestion
                polyfit_tec_profile: False
                num_sigma: 1.5
```

share/nisar/defaults/insar.yaml
```diff
@@ -830,6 +830,13 @@ runconfig:
                 # Boolean flag to activate/deactivate model-based solid earth tide corrections
                 solid_earth_tides_enabled: True
 
+            tec_correction:
+                apply_slant_range_correction: True
+                apply_azimuth_correction: True
+                use_total_tec_only: False
+                ignore_invalid_topside_tec: True
+                polyfit_tec_profile: False
```
bhawkins, line 838:
```suggestion
                polyfit_tec_profile: False
                num_sigma: 1.5
```

share/nisar/schemas/gcov.yaml
```diff
@@ -555,6 +558,11 @@ geocode_options:
     # (polarizations) to process.
     max_block_size: num(min=0, required=False)
 
+tec_correction_options:
+    use_total_tec_only: bool(required=False)
+    ignore_invalid_topside_tec: bool(required=False)
+    polyfit_tec_profile: bool(required=False)
```
bhawkins, line 564:
```suggestion
    polyfit_tec_profile: bool(required=False)
    num_sigma: num(min=0.0, required=False)
```

share/nisar/schemas/gslc.yaml
```diff
@@ -436,3 +438,10 @@ output_options:
 lut_options:
     # Boolean flag to activate/deactivate model-based solid earth tide corrections
     solid_earth_tides_enabled: bool(required=False)
+
+tec_correction_options:
+    apply_slant_range_correction: bool(required=False)
+    apply_azimuth_correction: bool(required=False)
+    use_total_tec_only: bool(required=False)
+    ignore_invalid_topside_tec: bool(required=False)
+    polyfit_tec_profile: bool(required=False)
```
bhawkins, line 447:
```suggestion
    polyfit_tec_profile: bool(required=False)
    num_sigma: num(min=0.0, required=False)
```

share/nisar/schemas/insar.yaml
```diff
@@ -1235,6 +1237,13 @@ lut_options:
     # Boolean flag to activate/deactivate model-based solid earth tide corrections
     solid_earth_tides_enabled: bool(required=False)
 
+tec_correction_options:
+    apply_slant_range_correction: bool(required=False)
+    apply_azimuth_correction: bool(required=False)
+    use_total_tec_only: bool(required=False)
+    ignore_invalid_topside_tec: bool(required=False)
+    polyfit_tec_profile: bool(required=False)
```
bhawkins, line 1245:
```suggestion
    polyfit_tec_profile: bool(required=False)
    num_sigma: num(min=0.0, required=False)
```

python/packages/isce3/atmosphere/tec_product.py
bhawkins: Now I'm having some doubts about what to do in this case.

python/packages/isce3/atmosphere/tec_product.py
seongsujeong: No it's no longer needed. It's the remnant on the doctoring of the older RANSAC implementation. I removed this in the revised commit.

python/packages/isce3/atmosphere/tec_product.py
seongsujeong: This is a constant to convert MAD to standard deviation of the normal distribution
https://en.wikipedia.org/wiki/Median_absolute_deviation
This constant is no longer valid in the revised implementation, so I removed the comment.

python/packages/isce3/atmosphere/tec_product.py
seongsujeong: `<=` will make trigger the log message below when the number of inliers are the same as the order. For example, if the order is `2` and we have `2` inliers, the if statement becomes `True`. If 3 or more, it satisfies the condition to perform 2nd order polynomial fitting.

python/packages/isce3/atmosphere/tec_product.py
seongsujeong: yes it should. Good catch.

python/packages/isce3/atmosphere/tec_product.py
seongsujeong: I think it would make sense to return the original values when any smoothing attempt fails. Not only this one but also the one in the below (not enough inliers)

python/packages/isce3/atmosphere/tec_product.py
seongsujeong: Not having enough inliers mean the TEC profile is messed up (which is rare). In that case, it would be hard to tell which one performs better (i.e. fitted or not fitted). In that case I think it's safe to leave the TEC values as they are.

python/packages/nisar/workflows/geocode_corrections.py
seongsujeong: `b_error=False` was added to the zero-filled LUT

python/packages/isce3/atmosphere/tec_product.py
seongsujeong: After the discussion, we will return `y` in this case.

python/packages/isce3/atmosphere/tec_product.py
seongsujeong: Same as above (return y)

python/packages/isce3/atmosphere/tec_product.py
```diff
@@ -70,9 +93,67 @@ def _compute_ionospheric_range_delay(utc_time: np.ma.MaskedArray,
...
+        Sample values (the TEC profile) to fit.
+    degree: int
+        Degree of the polynomial to fit.
+    num_sigma: float
+        Multiplier on the MAD to scale the threshold. Default 1.5.
+
+    Returns
+    -------
+    np.ndarray
+        The fitted (smoothed) values evaluated at every `x`.
```
hfattahi, line 122: Would you please clarify that at certain situation the original data is returned and not the fitted values.

hfattahi (approved):
LGTM! Thanks @seongsujeong . Please address the minor comment below

python/packages/isce3/atmosphere/tec_product.py
```diff
@@ -70,9 +93,67 @@ def _compute_ionospheric_range_delay(utc_time: np.ma.MaskedArray,
...
+    if np.count_nonzero(inliers) <= degree:
+        info_channel.log('Not enough inliers from TEC sample. Returning the original TEC values.')
+        return y
+
+    coeffs = np.polyfit(x[inliers], y[inliers], degree)
+    info_channel.log('MAD outlier detection and polynomial fitting completed.')
+    return np.polyval(coeffs, x)
+
+
 def _get_suborbital_tec(tec_json_dict: dict,
```
gshiroma: Should this function still be called `_get_suborbital_tec()`? Now that you’re allowing the total TEC to be returned, should the function be renamed?

python/packages/isce3/atmosphere/tec_product.py
```diff
@@ -96,17 +189,31 @@ def _get_suborbital_tec(tec_json_dict: dict,
     # compute sub orbital TEC from total and top TEC in JSON
     tot_tec = np.array(tec_json_dict[f'totTec{nr_fr}'])
     top_tec = np.array(tec_json_dict[f'topTec{nr_fr}'])
-    sub_orbital_tec = tot_tec - top_tec
+    if total_tec_only:
+        sub_orbital_tec = tot_tec
```
gshiroma: The variable name `sub_orbital_tec` seems to be wrong in this case. Maybe you can rename it to `orbital_tec`?

python/packages/nisar/workflows/geocode_corrections.py
```diff
@@ -208,12 +233,110 @@ def _get_iono_srange_corrections(cfg, slc, frequency, orbit):
...
+        # Convert the TEC UTC times to seconds since the radar grid reference
+        # epoch so they share the same time base as the azimuth time span.
+        tec_t_since_epoch = np.array(
+            [(isce3.core.DateTime(t_str) - ref_epoch).total_seconds()
+             for t_str in tec_dict['utc']])
+
+        # Crop the topside TEC data flags to the TEC samples that fall within
+        # the azimuth time span of the (decimated) radar grid.
+        i_start = np.searchsorted(tec_t_since_epoch, az_start, side='left')
+        i_stop = np.searchsorted(tec_t_since_epoch, az_stop, side='right')
```
gshiroma: The azimuth start and end times here don't seem to include the additional time margin (e.g., 40s). I just want to make sure that if `topTecNrDataFlag` or `topTecFrDataFlag` are set outside of the radargrid but within the additional time margin, it won't cause any problems.

gshiroma (commented):
Thank you @seongsujeong , for addressing my comments. I have a few additional comments below. The should be easy to address.

bhawkins (approved):
LGTM thanks

python/packages/isce3/atmosphere/tec_product.py
seongsujeong: Good point. I think we should change the name to avoid the confusion down the line.
The function was renamed `_get_tec_profile`

python/packages/isce3/atmosphere/tec_product.py
seongsujeong: `sub_orbital_tec` will be renamed to `tec_profile`

PR #370, opened 2026-08-25
RSLC missing data mask

bhawkins:
This is work in progress for improving the RSLC valid data mask. Currently, the RSLC mask only considers the transmit gaps and the range location of the swath (which may change every 10 s). We want it to also reflect missing data, typically echoes that were recorded on board but were lost in the downlink. Here's the contents of `validSamplesSubSwath1` for a missing data case using this branch as of today (green is start, red is stop):

image

Based on today's discussion the code needs substantial changes, but I thought it'd be helpful to open PR so it could be added to the delivery milestone, if nothing else. The core problem is that the existing metadata is not polarization-specific, so it can't represent polarization-specific missing data masks. Here are the changes we discussed:

- [ ] L0B should change how it populates its `validSamplesSubSwath1`. If the gap metadata is missing for one RX polarization but not the other, then it should be populated with the valid one.
- [ ] L0B should add new RX polarization specific 1D datasets called `pulseHasValidSamples` that report missing pulses. That saves users the trouble of reading the whole image and checking whether all the samples are fill_value in each row.
- [ ] RSLC `validSamplesSubSwath1` should basically just reflect the TX gaps and range timing as it currently does.
- [ ] RSLC should add new polarization-specific mask images to report missing data.

### DIFF
--- a/python/packages/nisar/focus/valid_regions.py
+++ b/python/packages/nisar/focus/valid_regions.py
@@ -221,6 +221,7 @@ def get_focused_sub_swaths(rawlist, out_chan, grid, orbit, doppler, dem, azres,
     """
     min_segment_length = _get_min_segment_length(min_segment_fraction, azres,
         grid, orbit, dem.ellipsoid, _get_prf(rawlist))
+    log.info(f"Using {min_segment_length=} for valid sub swaths")
 
     raw_bbox_lists, chirp_durations = get_raw_sub_swath_bboxes(rawlist,
         out_chan, orbit, num_ignore=num_ignore,
@@ -385,6 +386,7 @@ def save_valid_data_mask(rawlist, out_chan, grid, orbit, doppler, dem, azres,
 
     min_segment_length = _get_min_segment_length(min_segment_fraction, azres,
         grid, orbit, dem.ellipsoid, _get_prf(rawlist))
+    log.info(f"Using {min_segment_length=} for valid data mask")
 
     raw_bbox_lists, chirp_durations = get_raw_sub_swath_bboxes(rawlist,
         out_chan, orbit, num_ignore=num_ignore,
--- a/python/packages/nisar/focus/valid_regions.py
+++ b/python/packages/nisar/focus/valid_regions.py
@@ -13,14 +13,14 @@ class _PolBit(enum.IntEnum):
     Bit index assigned to each polarization channel in the valid-data mask
     written by save_valid_data_mask.
     """
-    HH = 0
-    HV = 1
-    VH = 2
-    VV = 3
-    LH = 4
-    LV = 5
-    RH = 6
-    RV = 7
+    HH = 8
+    HV = 9
+    VH = 10
+    VV = 11
+    LH = 12
+    LV = 13
+    RH = 14
+    RV = 15
 
 
 class PolValidMask(enum.IntFlag):
--- a/python/packages/nisar/focus/valid_regions.py
+++ b/python/packages/nisar/focus/valid_regions.py
@@ -1,6 +1,7 @@
 #!python3
 import enum
 import logging
+from math import ceil
 import isce3
 import numpy as np
 from nisar.mixed_mode import find_overlapping_channel
@@ -261,11 +262,44 @@ def _mark_all_valid(image, bit, blocksize=None):
     return nrows * ncols
 
 
+def _get_prf(rawlist, reduction=max):
+    """
+    Get a representative pulse repetition frequency for a group of Raw files.
+
+    Parameters
+    ----------
+    rawlist : Iterable[Raw]
+        List of input raw data (NISAR L0B) files.
+    reduction : Callable[[Iterable[float]], float], optional
+        How to combine results from multiple files.  Default=max
+    
+    Returns
+    -------
+    prf : float
+        Pulse repetition frequency in Hz
+    """
+    return reduction(1.0 / np.mean(np.diff(raw.getPulseTimes()))
+        for raw in rawlist)
+
+
+def _get_min_segment_length(min_segment_fraction, azres, grid, orbit, ellipsoid,
+                            prf):
+    if not (0 < min_segment_fraction <= 1.0):
+        raise ValueError("Expected 0 < min_segment_fraction <= 1 but got "
+            f"{min_segment_fraction}")
+
+    sardur = isce3.focus.get_sar_duration(grid.sensing_mid, grid.end_range,
+        orbit, ellipsoid, azres, grid.wavelength)
+
+    min_segment_length = ceil(min_segment_fraction * sardur * prf)
+    return max(1, min_segment_length)
+
+
 def save_valid_data_mask(rawlist, out_chan, grid, orbit, doppler, dem, azres,
                          image, rdr2geo_params=dict(), geo2rdr_params=dict(),
                          ignore_failure=False, polygon_segment_length=50.0,
                          num_ignore=25, max_observation_gap=0.002,
-                         min_segment_length=2000, max_pulse_gap=1,
+                         min_segment_fraction=0.25, max_pulse_gap=1,
                          blocksize=None):
     """
     Determine fully-focused regions of the image and write the result as a
@@ -321,16 +355,18 @@ def save_valid_data_mask(rawlist, out_chan, grid, orbit, doppler, dem, azres,
         and the first pulse of the following observation for the two to be
         considered seamless.  Larger raw data gaps may result in a synthetic
         aperture being marked invalid in the RSLC.
-    min_segment_length : int, optional
+    min_segment_fraction : float, optional
+        Lower bound on the number of consecutive valid pulses, expressed as a
+        fraction of the synthetic aperture length.
         Segments with fewer than this number of consecutive valid pulses
         will not be returned.  This helps avoid unnecessary bookkeeping when
         lots of missing pulses are sprinkled throughout an observation. Must
         be >= 1 pulse.
     max_pulse_gap : int, optional
-        Segments (each of which must be at least min_segment_length pulses)
-        separated by max_pulse_gap or fewer invalid pulses will be joined
-        together.  This provides an easy way to ignore isolated gaps of a
-        single invlid pulse, for example.  Set to 0 to disable merging.
+        Segments (each lasting at least min_segment_fraction) separated by
+        max_pulse_gap or fewer invalid pulses will be joined together.  This
+        provides an easy way to ignore isolated gaps of a single invlid pulse,
+        for example.  Set to 0 to disable merging.
     blocksize : int, optional
         Number of rows to rasterize and write at a time.  Defaults to the
         chunk size of `image` if it is an HDF5 dataset, otherwise 512.
@@ -341,6 +377,9 @@ def save_valid_data_mask(rawlist, out_chan, grid, orbit, doppler, dem, azres,
         Total number of valid pixels in the image.
     """
     bit = int(_PolBit[out_chan.pol])
+
+    min_segment_length = _get_min_segment_length(min_segment_fraction, azres,
+        grid, orbit, dem.ellipsoid, _get_prf(rawlist))
 
     raw_bbox_lists, chirp_durations = get_raw_sub_swath_bboxes(rawlist,
         out_chan, orbit, num_ignore=num_ignore,
--- a/python/packages/nisar/focus/valid_regions.py
+++ b/python/packages/nisar/focus/valid_regions.py
@@ -1,0 +1,140 @@
+#!python3
+import logging
+import isce3
+from nisar.mixed_mode import find_overlapping_channel
+
+log = logging.getLogger("focus")
+
+def get_focused_sub_swaths(rawlist, out_chan, grid, orbit, doppler, dem, azres,
+                           rdr2geo_params=dict(), geo2rdr_params=dict(),
+                           ignore_failure=False, polygon_segment_length=50.0,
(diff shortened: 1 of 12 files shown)


bhawkins:
This work probably supersedes #344

hfattahi:
Thank you @bhawkins for the PR . The listed items reflects what we discussed offline. Once the masks are more mature we will derive couple of scalar parameters representing missing data in the observation used for focusing the image.

bhawkins:
Alice implemented the necessary L0B changes. That wound up revealing that HV is mostly valid for this test image, making it an even better test than I thought. I implemented the RSLC features, though there are a couple subtleties.

One is that the validSamplesSubSwath mask now excludes regions that are invalid on all polarizations. That turned out to be easier than trying guess about missing information (e.g., handling the case where data is missing at a DWP update). I assume that's unobjectionable.

The other is that I implemented a single mask layer, assigning one bit to each of the possible polarizations. I figured that's a bit more efficient than having a separate mask image for each pol. I just called the dataset "mask" which is consistent with GSLC and GCOV products, at least. Open to suggestions. If we weren't concerned about backwards compatibility we could combine this new mask with the inputDataExceptionMask and also delete the validSamplesSubSwathX datasets.

rslc_image_and_mask

rslc_mask_h5web

bhawkins:
Oh yeah, I uploaded this RSLC product to s3://nisar-adt/bhawkins/scratch/rslc_mask_example.h5 if anyone wants a sample product to play with.

bhawkins:
Following this morning's discussion, I modified the code to change inputDataExceptionMask from uint8 to uint16 and use the new bits to store the valid data mask. This reduces the number of mask layers users have to read and minimizes changes to downstream workflows. I updated the description to read

> Bitwise OR of input data exception codes for each image pixel (0: no anomaly, 2: NISAR LSAR qFSP-H1 sample slip). Also includes OR of validity mask bits, where each bit is set on for fully-focused data or off for partially focused or missing data (256:HH, 512:HV, 1024:VH, 2048:VV, 4096:LH, 8192:LV, 16384:RH, 32768:RV)

I posted a sample data product to s3://nisar-adt/bhawkins/scratch/rslc_mask_example_20260831.h5

hfattahi:
Thanks @bhawkins for updating this PR based on few offline discussions. 
I just looked at the sample product and looks great. Now I wonder about a "scalar" metric that can represent the valid data in the product. I tried to compute the percentage of valid data in two different ways:

1) just parsed the new mask in this PR and computed the ratio of "number of non zero pixels" to  "total number of pixels" for each polarization and that gives:

HH: ~31% valid
HV: ~90% valid

2) I first evaluated subswathValidMask and then parsed the new mask. I computed the ratios for only valid part of the new mask based on the  subSwathValid mask.
HH: ~34% valid
HV: 100% valid

We know that for this product HV is fine and HH has missing data. Therefore, I think the second approach is more accurate to demonstrate the percentage of valid data in term of missing/corrupted data. The impact will be even more pronounced when we have fixed PRF with multiple transmit gaps. What do you think? Is it possible to add this metric?

python/packages/nisar/products/readers/Raw/Raw.py
```diff
@@ -1126,6 +1196,153 @@ class Raw(RawBase, family='nisar.productreader.raw'):
...
+    # Likewise, it ends when we go from having data to not having data.
+    starts = np.where(dx > 0)[0]
+    ends = np.where(dx < 0)[0]
+    assert len(starts) == len(ends)
+
+    return [(start, end) for (start, end) in zip(starts, ends)
+        if (end - start) >= min_segment_length]
+
+
+def split_segments(segments, breaks):
```
Tyler-g-hudson, line 1262: Should this function also take a `min_segment_length` argument? It seems like there could be a rare occasion where the segments and breaks are aligned such that a length-1 segment is produced here.

python/packages/nisar/workflows/focus.py
```diff
@@ -1974,6 +1853,17 @@ def focus(runconfig, runconfig_path=""):
             log.warning("Internal calibration (INT_CAL) file was not provided "
                 "so unable to populate inputDataExceptionMask")
 
+        # Set missing/valid data mask (same dataset).
+        for pol in pols:
+            log.info(f"Geting and saving valid data mask for {frequency}{pol}")
```
Tyler-g-hudson:
```suggestion
            log.info(f"Getting and saving valid data mask for {frequency}{pol}")
```

python/packages/nisar/focus/valid_regions.py
```diff
@@ -0,0 +1,363 @@
...
+    -------
+    num_valid : int
+        Total number of valid pixels in the image.
+    """
+    bit = int(_PolBit[out_chan.pol])
+
+    raw_bbox_lists, chirp_durations = get_raw_sub_swath_bboxes(rawlist,
+        out_chan, orbit, num_ignore=num_ignore,
+        min_segment_length=min_segment_length, max_pulse_gap=max_pulse_gap,
+        max_observation_gap=max_observation_gap, use_rx_pulse_mask=True)
```
Tyler-g-hudson, line 421: Elsewhere in the file (line 219) this function is also called, but with `use_rx_pulse_mask` defaulted to False, is it intentional that the `get_focused_sub_swaths` function and this are different in this regard?

Tyler-g-hudson (commented):
Some quick review

python/packages/nisar/focus/valid_regions.py
```diff
@@ -0,0 +1,363 @@
...
+        Number of rows to rasterize and write at a time.  Defaults to the
+        chunk size of `image` if it is an HDF5 dataset, otherwise 512.
+
+    Returns
+    -------
+    num_valid : int
+        Total number of valid pixels in the image.
+    """
+    bit = int(_PolBit[out_chan.pol])
+
```
bhawkins, line 413: Can calculate `min_segment_length` based on `azres` here.

python/packages/nisar/focus/valid_regions.py
Tyler-g-hudson: Offline discussion: The RSLC subswath vectors should be agnostic w/r to the receive pulses, but the valid data mask cares about this. Might warrant a comment but I'm not worried about it if not.

python/packages/nisar/products/readers/Raw/Raw.py
```diff
@@ -1126,6 +1196,153 @@ class Raw(RawBase, family='nisar.productreader.raw'):
...
+        times, as returned by `RawBase.getSubSwaths`.
+
+    Returns
+    -------
+    np.ndarray(bool)
+        1-D boolean array of length nt that is True for pulses where at
+        least one sub-swath has a non-empty valid interval and False for
+        pulses that are entirely within a transmit gap.
+    """
+    return (subswaths[..., 1] - subswaths[..., 0]).sum(axis=0) > 0
```
bhawkins:
```suggestion
    return (subswaths[..., 1] > subswaths[..., 0]).sum(axis=0) > 0
```
We need to be careful since `subswaths` could be an unsigned integer type, and subtraction could overflow. Just doing a comparison would be safe.

Tyler-g-hudson:
@xhuang-jpl this is the PR that I brought up to you offline just now

python/packages/nisar/products/readers/Raw/Raw.py
Tyler-g-hudson: Offline discussion: This doesn't necessarily need to be a problem - there is one line downstream:
```python
                    assert ipulse0 < (nt - 1)  # since min_segment_length > 1
```
which enforces a >2 line minimum segment length but @bhawkins thinks it might be artificial and fixable.

python/packages/nisar/focus/valid_regions.py
Tyler-g-hudson: Do you think maybe this minimum segment length concept could be inverted? That is to say, would it be better to have a "maximum missing rangelines per aperture" with the additional stipulation that those rangelines can't be adjacent?

bhawkins: I don't think that would sufficiently constrain the distribution of missing pulses. For example, if all the missing pulses were clustered in one spot but interleaved with one pulse of valid data, the effect on image quality is a lot worse than if they were distributed evenly throughout the aperture. Requiring contiguous blocks of valid data provides a stronger guarantee that regions marked as "valid" are actually uncorrupted.

python/packages/nisar/products/readers/Raw/Raw.py
bhawkins: I modified the code so that the "peek ahead" trick is only performed if the segment has at least two pulses. Then I went through and changed the lower bound on min_segment_length to 1 (instead of 2) and now it's okay if the call to `split_segments` results in a single-pulse segment.

bhawkins:
Okay, I think I've addressed all the comments from yesterday's review. The main changes were
* Allow single-pulse segments since splitting at DWP changes could cause those to happen.
* Calculate min_segment_length based on synthetic aperture size (user input is now fraction of SA)

@hfattahi I guess you weren't kidding about not seeing attributes 😆 . The mask dataset already has attributes that calculate what you described as option 1. Note that the sub-swath vectors will have null areas wherever both channels have missing data, so with option 2 one of the polarizations would have 100% valid by construction, even if tons of data were missing. I implemented the option I mentioned a week or two ago where the metric is the ratio of valid pulses to total pulses in the subset of the L0B that is used to create the RSLC. Let me know what you think:

mask attributes

hfattahi:
> Okay, I think I've addressed all the comments from yesterday's review. The main changes were
> 
> * Allow single-pulse segments since splitting at DWP changes could cause those to happen.
> * Calculate min_segment_length based on synthetic aperture size (user input is now fraction of SA)
> 
> @hfattahi I guess you weren't kidding about not seeing attributes 😆 . The mask dataset already has attributes that calculate what you described as option 1. Note that the sub-swath vectors will have null areas wherever both channels have missing data, so with option 2 one of the polarizations would have 100% valid by construction, even if tons of data were missing. I implemented the option I mentioned a week or two ago where the metric is the ratio of valid pulses to total pulses in the subset of the L0B that is used to create the RSLC. Let me know what you think:
> 
> mask attributes

Haha yeah I was not joking! Thanks @bhawkins I think the metrics are great and representative. 
Now the only part left is if we want to add 1-2 datasets to the identification or not. If we want to add, one option could be:
validRawPulseFractionMax
validRawPulseFractionMin
validMaskPixelFractionMax
validMaskPixelFractionMin

For each of those we would give the min/max of the attributes you have added for all polarizations. Therefore, regardless of how many frequency or polarizations a product may have, the identification will always have these four datasets. what do you think?

bhawkins:
I verified that this PR covers the case in NSDS-5279 where invalid data at the beginning of a short file causes failures and closed #344. Without this patch, the granule
```
NISAR_L0_PR_RRSD_026_041_A_156S_20260723T021153_20260723T021213_P05023_F_J_001.h5
```
fails but with this patch we get an image and valid data masks as follows (subswath start=blue, end=red):
short image with valid subswath overlaid

inputDataExceptionMask

Note that the radiometric artifact outside the valid data region in the top-right corner occurs because the L0B doesn't mark enough data invalid. I'll report this finding back to SDS.

hfattahi:
> I verified that this PR covers the case in NSDS-5279 where invalid data at the beginning of a short file causes failures and closed #344. Without this patch, the granule
> 
> ```
> NISAR_L0_PR_RRSD_026_041_A_156S_20260723T021153_20260723T021213_P05023_F_J_001.h5
> ```
> 
> fails but with this patch we get an image and valid data masks as follows (subswath start=blue, end=red): short image with valid subswath overlaid
> 
> inputDataExceptionMask
> Note that the radiometric artifact outside the valid data region in the top-right corner occurs because the L0B doesn't mark enough data invalid. I'll report this finding back to SDS.

That is great news!

hfattahi:
> > Okay, I think I've addressed all the comments from yesterday's review. The main changes were
> > 
> > * Allow single-pulse segments since splitting at DWP changes could cause those to happen.
> > * Calculate min_segment_length based on synthetic aperture size (user input is now fraction of SA)
> > 
> > @hfattahi I guess you weren't kidding about not seeing attributes 😆 . The mask dataset already has attributes that calculate what you described as option 1. Note that the sub-swath vectors will have null areas wherever both channels have missing data, so with option 2 one of the polarizations would have 100% valid by construction, even if tons of data were missing. I implemented the option I mentioned a week or two ago where the metric is the ratio of valid pulses to total pulses in the subset of the L0B that is used to create the RSLC. Let me know what you think:
> > mask attributes
> 
> Haha yeah I was not joking! Thanks @bhawkins I think the metrics are great and representative. Now the only part left is if we want to add 1-2 datasets to the identification or not. If we want to add, one option could be: validRawPulseFractionMax validRawPulseFractionMin validMaskPixelFractionMax validMaskPixelFractionMin
> 
> For each of those we would give the min/max of the attributes you have added for all polarizations. Therefore, regardless of how many frequency or polarizations a product may have, the identification will always have these four datasets. what do you think?

We discussed this suggestion in an offline meeting and we agreed to wait on these few extra variables and instead to push for carrying the existing and more clear attributes (per frequency and per polarization) all the way to CMR. If there will be any technical obstacle, we will revise back this idea in a different small PR.

hfattahi (approved):
LGTM! Thank you @bhawkins for this great work and very useful addition.

Tyler-g-hudson (approved):
Apologies for the delay - LGTM given the offline discussion today

PR #372, opened 2026-08-26
Use optimized HDF5 reader for resample_slc_v2

lijun99:
resample_secondary_rslc_onto_reference() read the secondary RSLC via RSLC.getSlcDatasetAsNativeComplex(), which opens the file with h5py's default chunk cache instead of one sized for the dataset's own chunking. Open the secondary RSLC directly via HDF5OptimizedReader instead, matching the pattern already used in unwrap.py, crossmul.py, and other workflows. This cuts the resample step's SLC I/O read time by ~5-10x. Output is verified bit-for-bit identical to the original.

Moved from #358.

### DIFF
--- a/python/packages/nisar/workflows/resample_slc_v2.py
+++ b/python/packages/nisar/workflows/resample_slc_v2.py
@@ -10,7 +10,9 @@ import numpy as np
 from journal.Channel import Channel
 
 from isce3.core.gpu_check import use_gpu
+from isce3.core.types import ComplexFloat16Decoder, is_complex32
 from isce3.image.v2.resample_slc import resample_slc_blocks
+from isce3.io import HDF5OptimizedReader
 from isce3.io.gdal.gdal_raster import GDALRaster
 
 from nisar.products.readers import RSLC
@@ -32,7 +34,7 @@ def run(cfg: dict, resample_type: str) -> None:
         of the runconfig.
         "fine" uses ".off.vrt" files under the rubbersheet_offsets subdirectory of the
         offsets directory of the runconfig.
-    """ 
+    """
     sec_file_path = cfg["input_file_group"]["secondary_rslc_file"]
     ref_file_path = cfg["input_file_group"]["reference_rslc_file"]
     scratch_path = Path(cfg["product_path_group"]["scratch_path"])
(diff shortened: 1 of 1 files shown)


python/packages/nisar/workflows/resample_slc_v2.py
```diff
@@ -205,29 +207,38 @@ def resample_secondary_rslc_onto_reference(
...
+    # than via RSLC.getSlcDatasetAsNativeComplex()'s default h5py cache.
+    with HDF5OptimizedReader(
+        name=os.fspath(sec_file_path), mode='r', libver='latest', swmr=True
+    ) as sec_h5:
+        sec_readers = []
+        for pol in pols:
+            ds_path = sec_slc_obj.imageDatasetPath(freq, pol)
+            dataset = sec_h5[ds_path]
+            if is_complex32(dataset):
+                dataset = ComplexFloat16Decoder(dataset)
```
hfattahi, line 224: It's been a long time since we moved on from float 16 for RSLC. We have not produced any NISAR real data RSLC with float 16. I think we don't need this check. Do you agree @bhawkins ?

Tyler-g-hudson: Is any of our test data float16?

Tyler-g-hudson (approved):
LGTM pending the resolution of Heresh's concern regarding float16.

python/packages/nisar/workflows/resample_slc_v2.py
lijun99: I would recommend keeping the check for now, 1) the original call [`getSlcDatasetAsNativeCompl`](https://github.com/isce-framework/isce3/blob/23f99329d71515516a72d9e9b03167fea4644628/python/packages/nisar/products/readers/SLC/RSLC.py#L82) offers the same check and conversion, 2) it seems that there are some other places which do the same, such as calls to [`is_dataset_complex32`](https://github.com/isce-framework/isce3/blob/23f99329d71515516a72d9e9b03167fea4644628/python/packages/nisar/products/readers/SLC/RSLC.py#L93), [`read_c4_dataset_as_c8`](https://github.com/isce-framework/isce3/blob/23f99329d71515516a72d9e9b03167fea4644628/python/packages/isce3/core/types.py#L46). If we decide to drop complex32, it's better to handle them altogether.

hfattahi: Fine with me. let's keep it.

hfattahi (approved):
LGTM. Thank you @lijun99 for the contribution.

Tyler-g-hudson:
Thanks @lijun99 for the contribution!

PR #373, opened 2026-08-27
Improve definition of isMixedMode for RSLC

bhawkins:
We noticed an edge case where RSLC reports isMixedMode=True in the metadata, but the file name (determined by PCM) did not indicate a mixed mode product. It turned out PCM was correct. The test in RSLC was to check if any input mode differs from the output mode. Usually that works fine, but in cases like QD5+5 we have to re-label one of the frequency B images to frequency A so that the RSLC can correctly represent their distinct center frequencies. The fix in this PR is to just compare the input modes to each other rather than to the output mode.

### DIFF
--- a/python/packages/nisar/mixed_mode/logic.py
+++ b/python/packages/nisar/mixed_mode/logic.py
@@ -256,3 +256,13 @@ def find_overlapping_channel(raw: RawBase, desired: PolChannel) -> PolChannel:
         if chan.intersection(desired).isvalid:
             return chan
     raise ValueError(f"Raw file does not contain channel intersecting {desired}.")
+
+
+def check_mixed_mode(modes: list[PolChannelSet]) -> bool:
+    """
+    Determine whether a list of modes represents a mixed-mode case.  True if any
+    mode doesn't match one of the others.
+    """
+    if len(modes) == 0:
+        return False
+    return any(mode != modes[0] for mode in modes)
--- a/python/packages/nisar/workflows/focus.py
+++ b/python/packages/nisar/workflows/focus.py
@@ -13,7 +13,7 @@ from nisar.antenna import AntennaPattern, get_calib_range_line_idx
 from nisar.noise.noise_estimation_from_raw import (
     est_noise_power_in_focus, NoiseEquivalentBackscatterProduct)
 from nisar.mixed_mode import (PolChannel, PolChannelSet, Band,
-    find_overlapping_channel)
+    find_overlapping_channel, check_mixed_mode)
 from nisar.products.readers.antenna import AntennaParser
 from nisar.products.readers.instrument import InstrumentParser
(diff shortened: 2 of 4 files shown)


python/packages/nisar/workflows/focus.py
```diff
@@ -1896,8 +1896,8 @@ def focus(runconfig, runconfig_path=""):
         is_full_frame=is_full_frame, frame_coverage=overlap,
         coverage_threshold=cfg.geometry.full_coverage_threshold_percent / 100,
         is_dithered=is_dithered, granule_id=granule_id,
-        is_mixed_mode=any(PolChannelSet.from_raw(raw) != common_mode
-            for raw in rawlist),
+        is_mixed_mode=check_mixed_mode([PolChannelSet.from_raw(raw)
```
rad-eng-59, line 1899: Just out of curiosity, how an empty-set common mode, where there is no intersection among input raw files,  is handled prior to this.

python/packages/nisar/mixed_mode/logic.py
```diff
@@ -256,3 +256,13 @@ def find_overlapping_channel(raw: RawBase, desired: PolChannel) -> PolChannel:
...
             return chan
     raise ValueError(f"Raw file does not contain channel intersecting {desired}.")
+
+
+def check_mixed_mode(modes: list[PolChannelSet]) -> bool:
+    """
+    Determine whether a list of modes represents a mixed-mode case.  True if any
+    mode doesn't match one of the others.
+    """
+    if len(modes) == 0:
```
rad-eng-59, line 266: Should not an exception being raised here when there is no mode, no raw inputs? I have been thinking at least one mode/raw file is needed and if none stop the workflow asap. Perhaps, when there is no input raw file or no common mode, the exception being already raised somewhere in focus.py?

rad-eng-59 (approved):
LGTM. Thank you.

Just for the sake of my own recap of focus.py, the output grid time span has no role in common and mixed mode logic. Simply it is assumed all raw files /modes will make contribution to some extent in the output grid no matter what being set by users in respective RSLC runconfig and thus the common mode selection and mixed mode logic is purely based on chirp spectrum and TxRx polarization.

hfattahi (approved):
LGTM. The PR fixes an edge case with Quasi-dual!

python/packages/nisar/workflows/focus.py
bhawkins: Good point. I don't really know what happens in that case (probably an obscure failure) and opened #377 to track progress on that.

python/packages/nisar/mixed_mode/logic.py
bhawkins: I think it's okay for the low-level function to say the empty set is not mixed mode. But I agree the high-level workflow should fail with an informative error.

PR #374, opened 2026-08-27
Fix SNAPHU unwrap incorrect connected components 

xhuang-jpl:
This PR is to fix the connected components from the SNAPHU in the unwrap.py where the valid mask dose not pass to the SNAPHU causing the incorrect connected components. This PR is to pass the valid mask combing the preprocess step and SNAPHU mask to the SNAPHU algorithm.


hfattahi (commented):
Let's close this one as it is addressed in the whirwind PR https://github.com/isce-framework/isce3/pull/367

xhuang-jpl:
closed because it is addressed in the whirlwind PR https://github.com/isce-framework/isce3/pull/367

PR #379, opened 2026-09-10
Add polarization-specific `validDataMask` datasets to InSAR product

xhuang-jpl:
### Summary

Add polarization-specific `validMask` datasets to InSAR products for tracking reference and secondary image validity per polarization.

### Changes

- **New Dataset**: `validMask` (uint8) with bit encoding - bit 1: reference validity, bit 0: secondary validity
- **Geocoding**: Added `get_valid_mask_input_output()` and updated CPU/GPU workflows to geocode validMask
- **Writers**: Modified GUNW, GOFF, ROFF, and L1 writers to create validMask in pixelOffsets, unwrappedInterferogram, and wrappedInterferogram groups


s-sasaki-earthsea-wizard:
Thanks for integrating the row-wise `generate_insar_mask` vectorization from #359 into this PR — and the sliding-window `_RSLCInputDataExceptionMask` reader is a genuine improvement over #359's full-array loads, as the numbers below show.

I ran this PR's head (`0c2ff0af5`) against its merge-base (`0d1600d8b`) on a real NISAR L-band frame (ascending 139/019 pair, frequency A / HH, RIFG interferogram grid 6840 x 10581 = 72.4 Mpx), as a standalone `prepare_insar_hdf5` run against a pre-seeded `geo2rdr` scratch — the same harness as the #358/#359 comparison posted [here](https://github.com/isce-framework/isce3/pull/359#issuecomment-5434042896). Only the 7 files this PR changes differ between the runs.

| variant | prepare_insar_hdf5 (journal) | vs scalar | peak RSS |
|---|---|---|---|
| merge-base (scalar loop) | 418.0 s | 1x | 12.5 GiB |
| this PR | 84.8 s | 4.9x | **6.6 GiB** |

**No regressions on this dataset.** All 558 common datasets across the GUNW skeleton and the RIFG/RUNW scratch skeletons are byte-identical to the merge-base run except the usual run-varying metadata (`processingDateTime`, `runConfigurationContents`); that includes every existing `mask` layer. The only structural additions are the 7 new `validMask` datasets, and the radar-grid ones check out per pixel: HH validMask equals the sub-swath validity bits of the co-located `mask` (`2*(ref digit > 0) + (sec digit > 0)`) with zero mismatches on all four RIFG/RUNW layers (RIFG interferogram: 69,341,858 px with either image valid, 68,664,482 with both; pixelOffsets: 368,611 / 366,145). The GUNW validMask planes are creation-state skeletons at the prepare stage, so their geocoding is not covered by this check. Peak RSS lands below both the scalar baseline and #359 (10.1 GiB on the same harness), thanks to the windowed exception-mask reads. Full logs, provenance, and the comparison scripts: [`artifacts/pr379-ab-20260913/`](https://github.com/s-sasaki-earthsea-wizard/isce3-benchmark/tree/e8bb692c2ee27c2d6c5d93ad847554d7b1148618/artifacts/pr379-ab-20260913).

The performance and the tested-configuration results look good. Reading the integrated code turned up two validMask correctness notes in input configurations beyond the tested one, plus two minor cleanups — details and suggested fixes are folded below.

<details>
<summary>Two validMask correctness notes (multi-frequency products; non-HH polarizations with a uint8 or missing inputDataExceptionMask) — worth addressing before merge; neither manifests on the frame above</summary>

1. **The per-pol validMask update for pixelOffsets runs outside the frequency loop.** In `InSAR_L1_writer.add_pixel_offsets_to_swaths_group`, the `for pol in pol_list:` block that writes `pixelOffsets/{pol}/validMask` ([InSAR_L1_writer.py#L570-L583](https://github.com/isce-framework/isce3/blob/0c2ff0af5e03e7560bc5aaffda72c1de35a7a2f6/python/packages/nisar/products/insar/InSAR_L1_writer.py#L570-L583)) sits at method level, while `pol_valid_mask`, `pol_list`, and `offset_group_name` are loop variables of the `for freq, pol_list, _ in get_cfg_freq_pols(...)` loop starting at [L371](https://github.com/isce-framework/isce3/blob/0c2ff0af5e03e7560bc5aaffda72c1de35a7a2f6/python/packages/nisar/products/insar/InSAR_L1_writer.py#L371). For a multi-frequency product only the last frequency's validMask is written; frequency A's stays at its creation bytes (all zeros = "both invalid", with `_FillValue = 255`). The interferogram counterpart in `add_interferogram_to_swaths_group` is correctly inside the loop. Note that indenting the update block alone is not quite enough: `_add_datasets_to_pixel_offset_group()`, which creates the per-pol validMask datasets, also runs after the loop ([L568](https://github.com/isce-framework/isce3/blob/0c2ff0af5e03e7560bc5aaffda72c1de35a7a2f6/python/packages/nisar/products/insar/InSAR_L1_writer.py#L568)), so each frequency's dataset creation and validMask write need to happen inside the same frequency scope, in that order.

2. **The fallback for RSLCs without per-pol validity bits populates only the HH position.** For RSLCs with a pre-#370 uint8 `inputDataExceptionMask` — and equally when the dataset is absent — the fallback sets bit 8 (reference) / bit 0 (secondary) from sub-swath validity ([utils.py#L893-L894](https://github.com/isce-framework/isce3/blob/0c2ff0af5e03e7560bc5aaffda72c1de35a7a2f6/python/packages/nisar/products/insar/utils.py#L893-L894), [L914-L915](https://github.com/isce-framework/isce3/blob/0c2ff0af5e03e7560bc5aaffda72c1de35a7a2f6/python/packages/nisar/products/insar/utils.py#L914-L915)). Those are the `_PolBit.HH` positions, so `extract_pol_valid_mask` returns all-invalid for every other polarization of such granules (e.g. HV of a DHDH pair). A minimal repro (in the artifact bundle linked above) confirms it at runtime: on a fully-valid grid with zero offsets, `mask` is 11 everywhere while the extracted validMask is HH=3 and HV/VH/VV=0, for both a uint8 and an absent exception mask. If the intent is "no per-pol information, fall back to sub-swath validity for all pols", broadcasting the fallback to all eight bit positions (`* 0xFF00` / `* 0x00FF`) would preserve that.

</details>

<details>
<summary>Minor cleanups (fine as a follow-up)</summary>

3. **`_subswath_numbers` is defined twice in utils.py.** The first definition ([L497](https://github.com/isce-framework/isce3/blob/0c2ff0af5e03e7560bc5aaffda72c1de35a7a2f6/python/packages/nisar/products/insar/utils.py#L497), the verbatim #359 version) is shadowed by the second at [L703](https://github.com/isce-framework/isce3/blob/0c2ff0af5e03e7560bc5aaffda72c1de35a7a2f6/python/packages/nisar/products/insar/utils.py#L703) and can be deleted.

4. `pixel_offsets_ds_params` in `InSAR_L1_writer._add_datasets_to_pixel_offset_group` lists `slantRangeOffset` twice ([L328](https://github.com/isce-framework/isce3/blob/0c2ff0af5e03e7560bc5aaffda72c1de35a7a2f6/python/packages/nisar/products/insar/InSAR_L1_writer.py#L328), [L335](https://github.com/isce-framework/isce3/blob/0c2ff0af5e03e7560bc5aaffda72c1de35a7a2f6/python/packages/nisar/products/insar/InSAR_L1_writer.py#L335)) — harmless through `require_dataset`, but presumably unintended.

</details>

On #359: since this PR carries the implementation, I'll downshift #359 to its regression tests only (the offer from the #358/#359 threads). The suite passes against this PR's head as-is — 11/11, including the adversarial rounding and bit-packing cases — so it can serve as the regression gate for the uint32 `mask` semantics of the integrated implementation (it also sanity-checks the per-pol validity word against the mask's sub-swath digits; the finer per-pol bit semantics are this PR's own and are best covered by its tests).

python/packages/nisar/products/insar/GOFF_writer.py
```diff
@@ -149,6 +149,35 @@ def add_grids_to_hdf5(self):
...
+                    self.require_group(pixeloffsets_pol_name)
+
+                yds, xds = set_get_geo_info(
+                    self,
+                    pixeloffsets_pol_name,
+                    goff_geogrids,
+                )
+                self._create_2d_dataset(
+                    pixeloffsets_pol_group,
+                    "validMask",
```
hfattahi: Since we already have subswath Valid Data Mask suggest

```suggestion
                    "validDataMask",
```

python/packages/nisar/products/insar/GOFF_writer.py
```diff
@@ -149,6 +149,35 @@ def add_grids_to_hdf5(self):
...
+                    goff_shape,
+                    np.uint8,
+                    (f"Valid mask for the {pol} layers: "
+                     "bit 1 = reference (1=valid, 0=invalid), bit 0 = secondary (1=valid, 0=invalid)"),
```
hfattahi: Can we expand and make the description more informative. Maybe something like this:
```suggestion
                     "bit 1 = reference (1=valid, 0=invalid), bit 0 = secondary (1=valid, 0=invalid). Valid represents fully focused data and invalid represents partially focused or missing data"),
```
@bhawkins please take a look if that extra description is consistent with RSLC.

python/packages/nisar/products/insar/InSAR_L1_writer.py
```diff
@@ -312,30 +312,52 @@ def _add_datasets_to_pixel_offset_group(self):
...
+                        "slantRangeOffset",
+                        np.float32,
+                        "Slant range offset",
+                        Units.meter,
+                        None,
+                    ),
+                    (
+                        "validMask",
+                        np.uint8,
+                        f"Valid mask for the {pol} layers: bit 1 = reference (1=valid, 0=invalid), bit 0 = secondary (1=valid, 0=invalid)",
```
hfattahi: Similar comment as above about more clear description.

hfattahi (commented):
Very minor comments before we go to the details of this PR. 

Also one note that we need to discuss in a PR review:

what is the behavior when the workflow runs on older/existing RSLCs with 8 bits inputDataExceptionMask. Ideally we should probably populate the InSAR validDataMask inside polarization group with the same subswathValidDataMask in "mask" dataset.

python/packages/nisar/products/insar/utils.py
```diff
@@ -494,6 +436,209 @@ def generate_dem_rdr(radar_grid_obj,
...
+    ValueError
+        If the dataset shape differs from (lines, samples)
+
+    Notes
+    -----
+    Line indices are bounds-checked on every request; sample indices
+    are not, so negative sample indices would wrap around as in NumPy.
+    """
+
+    def __init__(self, dataset, lines, samples, block_lines=256):
```
bhawkins, line 475:
```suggestion
    def __init__(self, dataset, lines, samples, block_lines=None):
        if block_lines is None:
            if dataset is not None and dataset.chunks is not None:
                block_lines = dataset.chunks[0]
            else:
                block_lines = 512
```
Let's default to using the chunk size and only use a different value if the user requested it.

python/packages/nisar/products/insar/utils.py
```diff
@@ -494,6 +436,209 @@ def generate_dem_rdr(radar_grid_obj,
...
+            # Zero-stride view of a single zero: spans the whole grid
+            # without allocating it, so no request ever triggers a read
+            self._block = np.broadcast_to(np.uint8(0), (lines, samples))
+            return
+        if dataset.shape != (lines, samples):
+            raise ValueError(
+                f"inputDataExceptionMask shape {dataset.shape} differs "
+                f"from the swath shape {(lines, samples)}")
+        self._chunk_lines = dataset.chunks[0] if dataset.chunks else 1
+        self._block_lines = max(block_lines, self._chunk_lines)
```
bhawkins, line 489:
```suggestion
        self._block_lines = block_lines
```

python/packages/nisar/products/insar/utils.py
```diff
@@ -503,114 +648,202 @@ def generate_insar_mask(ref_rslc_obj,
...
+
+        # Secondary RSLC input exception mask bits at the nearest
+        # secondary sample (round() of the scalar code, i.e. half to
+        # even, as np.rint); out-of-swath samples are zeroed by gather()
+        sec_i = np.rint(i + az_off).astype(np.int64)
+        sec_j = np.rint(rg_idx_arr + rg_off).astype(np.int64)
+        sec_in_swath = ((sec_i >= 0) & (sec_i < sec_swath.lines) &
+                        (sec_j >= 0) & (sec_j < sec_swath.samples))
+
+        sec_exception_mask_row = sec_exception_mask.gather(sec_i, sec_j, sec_in_swath)
```
bhawkins, line 793:
```suggestion
        sec_exception_mask_row = np.where(
            sec_in_swath,
            sec_exception_mask.gather(sec_i[sec_in_swath], sec_j[sec_in_swath]),
            fill_value)
```

python/packages/nisar/products/insar/utils.py
```diff
@@ -494,6 +436,209 @@ def generate_dem_rdr(radar_grid_obj,
...
+
+        Raises
+        ------
+        IndexError
+            If i is not within [0, lines)
+        """
+        self._ensure(i, i)
+        return self._block[i - self._start, rg_idx]
+
+    def gather(self, az_idx, rg_idx, valid):
```
bhawkins, line 555:
```suggestion
    def gather(self, az_idx, rg_idx):
```

python/packages/nisar/products/insar/utils.py
```diff
@@ -494,6 +436,209 @@ def generate_dem_rdr(radar_grid_obj,
...
+        lo : int
+            First line requested
+        hi : int
+            Last line requested, hi >= lo
+
+        Raises
+        ------
+        IndexError
+            If the line range is not within [0, lines)
+        """
```
bhawkins, line 507:
```suggestion
        """
        if hi < lo:
            raise ValueError()
```

bhawkins (commented):
Here's what we came up with looking though this today. However, we ran out of time, and I may need to revise some of these comments once I have a better understanding of what's going on here.

s-sasaki-earthsea-wizard:
@hfattahi
On the older-RSLC / 8-bit `inputDataExceptionMask` question: that case is note 2 of [my earlier comment](https://github.com/isce-framework/isce3/pull/379#issuecomment-5651215648). The fallback currently sets only the `HH` bit positions ([utils.py#L779-L780](https://github.com/isce-framework/isce3/blob/c7a7e86fec3912d171a45aeb71f977768e78c1e9/python/packages/nisar/products/insar/utils.py#L779-L780), [#L800-L801](https://github.com/isce-framework/isce3/blob/c7a7e86fec3912d171a45aeb71f977768e78c1e9/python/packages/nisar/products/insar/utils.py#L800-L801)), so `extract_pol_valid_mask` returns all-invalid for every non-HH polarization of such a granule — confirmed at runtime on a fully-valid grid (`mask` = 11 everywhere, extracted validMask HH=3 and HV/VH/VV=0), for both a uint8 and an absent exception mask. Broadcasting the fallback to all eight bit positions (`* 0xFF00` / `* 0x00FF`) would give exactly the "same sub-swath validity for all polarizations" behavior you describe; the repro script is in the artifact bundle linked from that comment.

For bookkeeping: `c7a7e86fe` covers note 3 of that comment; notes 1 and 4 still apply to the current head.

xhuang-jpl:
Thanks @s-sasaki-earthsea-wizard for finding those major bugs, which have been fixed. @hfattahi  if it is old RSLC uint8, it will fallback to the validSubswathMask, but its values are only 1 or 0, not the subswath mask number.

python/packages/nisar/products/insar/utils.py
```diff
@@ -494,6 +436,209 @@ def generate_dem_rdr(radar_grid_obj,
...
+    the first sub-swath whose per-line valid-sample interval
+    [start, end) contains the sample. An empty interval array claims
+    every in-bounds sample (matching the scalar API's short-circuit),
+    and a dataset without sub-swath information assigns 1 everywhere in
+    bounds.
+
+    Parameters
+    ----------
+    subswaths : isce3.product.SubSwaths
+        The sub-swaths object of the RSLC
```
bhawkins, line 603: I'd just pass `length` and `width` here instead of the subswaths object. Otherwise it's confusing why you'd need both `subswaths` and `intervals`. We do actually need that capability because this function is typically called on a subset of rows at a time.

python/packages/nisar/products/insar/utils.py
```diff
@@ -494,6 +436,209 @@ def generate_dem_rdr(radar_grid_obj,
...
+    numbers = np.zeros_like(in_bounds, dtype=np.uint8)
+    if not intervals:
+        return np.where(in_bounds, np.uint8(1), numbers)
+
+    # Clipped so the per-line gather stays legal; out-of-bounds samples
+    # are excluded through in_bounds
+    azi_gather = np.clip(azi_idx, 0, subswaths.length - 1)
+    remaining = in_bounds
+    for number, interval in enumerate(intervals, start=1):
+        if interval.size == 0:
```
bhawkins, line 628: I don't really understand why we have to handle this case. Would this ever crash if we just `assert interval.size > 0`?

python/packages/nisar/products/insar/utils.py
```diff
@@ -503,117 +648,202 @@ def generate_insar_mask(ref_rslc_obj,
...
+    sec_subswaths = sec_swath.sub_swaths()
+
+    # Fetch each sub-swath's per-line valid-sample interval array once
+    # (1-based API); the per-sample sub-swath tests then run as numpy
+    # array operations instead of scalar SubSwaths.get_sample_sub_swath
+    # calls per output pixel
+    ref_intervals = [ref_subswaths.get_valid_samples_array(s)
+                     for s in range(1, ref_subswaths.num_sub_swaths + 1)]
+    sec_intervals = [sec_subswaths.get_valid_samples_array(s)
+                     for s in range(1, sec_subswaths.num_sub_swaths + 1)]
```
bhawkins, line 710:
```suggestion
    ref_intervals = ref_subswaths.get_valid_samples_arrays_vect()
    sec_intervals = sec_subswaths.get_valid_samples_arrays_vect()
```

python/packages/nisar/products/insar/utils.py
```diff
@@ -503,117 +648,202 @@ def generate_insar_mask(ref_rslc_obj,
...
+    rg_idx_arr : numpy.ndarray
+        1-D slant range indices of the output columns in the reference
+        radar grid; may be fractional or outside the grid
 
     Returns
-    ----------
+    -------
     numpy.ndarray
-        mask at a given frequency
+        uint32 mask of shape (len(azi_idx_arr), len(rg_idx_arr))
```
bhawkins, line 694: Update docstring to reflect that we're returning two mask layers now.

python/packages/nisar/products/insar/utils.py
```diff
@@ -503,117 +648,202 @@ def generate_insar_mask(ref_rslc_obj,
...
+    col_out_of_swath = (rg_idx_arr < 0) | (rg_idx_arr >= ref_swath.samples)
+    # Clipped copy so the per-line gathers stay legal; out-of-swath
+    # columns are zeroed at the end of each row
+    rg_gather = np.clip(rg_idx_int, 0, ref_swath.samples - 1)
+
+    mask = np.zeros((len(azi_idx_arr), len(rg_idx_arr)), dtype=np.uint32)
+    # Polarization dependent valid mask
+    pol_valid_mask = np.zeros((len(azi_idx_arr), len(rg_idx_arr)), dtype=np.uint16)
+
+    for row, i in enumerate(azi_idx_arr):
```
bhawkins, line 745: Index names could be clearer.

python/packages/nisar/products/insar/utils.py
```diff
@@ -503,117 +648,202 @@ def generate_insar_mask(ref_rslc_obj,
...
+        # Rows outside the reference radar grid stay 0
+        if not (0 <= i < ref_swath.lines):
+            continue
+
+        # Geometric coregistration offsets at the truncated reference
+        # indices of the output pixels
+        i_int = int(i)
```
bhawkins, line 752:
```suggestion
        i_int = round(i)
```
Wouldn't it be better to round?

python/packages/nisar/products/insar/utils.py
```diff
@@ -503,117 +648,202 @@ def generate_insar_mask(ref_rslc_obj,
...
+
+        # Sub-swath numbers of the reference RSLC and, at the nearest
+        # secondary sample (int(x + 0.5) of the scalar code, i.e.
+        # truncation toward zero), of the secondary RSLC
+        ref_num = _subswath_numbers(ref_subswaths, ref_intervals,
+                                    i_int, rg_idx_int)
+        sec_num = _subswath_numbers(
+            sec_subswaths, sec_intervals,
+            np.trunc(i_int + az_off + 0.5).astype(np.int64),
+            np.trunc(rg_idx_int + rg_off + 0.5).astype(np.int64))
```
bhawkins, line 766:
```suggestion
            np.round(i_int + az_off).astype(np.int64),
            np.round(rg_idx_int + rg_off).astype(np.int64))
```
Not sure why one would prefer `trunc`

python/packages/nisar/products/insar/utils.py
```diff
@@ -503,117 +648,202 @@ def generate_insar_mask(ref_rslc_obj,
...
+        mask_row = (10 * ref_num + sec_num).astype(np.uint32)
+
+        # Reference RSLC input exception mask bits: keep the low 8 bits,
+        # then widen to uint32 before the shift so the packing is safe
+        # under NEP 50 scalar promotion as well
+        ref_exception_mask_row = ref_exception_mask.row(i_int, rg_gather)
+        mask_row |= (ref_exception_mask_row
+                     .astype(np.uint8).astype(np.uint32) << 16)
```
bhawkins, line 774:
```suggestion
        mask_row |= (ref_exception_mask_row & 0xff).astype(np.uint32) << 16
```
Let's make the bit mask explicit rather than just relying on subtle `astype` behavior.

python/packages/nisar/products/insar/utils.py
```diff
@@ -503,117 +648,202 @@ def generate_insar_mask(ref_rslc_obj,
...
+        # secondary sample (round() of the scalar code, i.e. half to
+        # even, as np.rint); out-of-swath samples are zeroed by gather()
+        sec_i = np.rint(i + az_off).astype(np.int64)
+        sec_j = np.rint(rg_idx_arr + rg_off).astype(np.int64)
+        sec_in_swath = ((sec_i >= 0) & (sec_i < sec_swath.lines) &
+                        (sec_j >= 0) & (sec_j < sec_swath.samples))
+
+        sec_exception_mask_row = sec_exception_mask.gather(sec_i, sec_j, sec_in_swath)
+        mask_row |= (sec_exception_mask_row
+                     .astype(np.uint8).astype(np.uint32) << 8)
```
bhawkins, line 795:
```suggestion
        mask_row |= (sec_exception_mask_row & 0xff).astype(np.uint32) << 8
```

python/packages/nisar/products/insar/utils.py
```diff
@@ -503,117 +648,202 @@ def generate_insar_mask(ref_rslc_obj,
...
+
+    Returns
+    -------
+    numpy.ndarray
+        uint8 array where bit 1 = reference valid (1=valid, 0=invalid)
+        and bit 0 = secondary valid (1=valid, 0=invalid)
+    """
+    # Map polarization to bit position (0-7) based on the standard order
+    pol_to_bit = {'HH': 0, 'HV': 1, 'VH': 2, 'VV': 3,
+                  'LH': 4, 'LV': 5, 'RH': 6, 'RV': 7}
```
bhawkins, line 839: Use the enum from nisar.focus.valid_regions

python/packages/nisar/products/insar/InSAR_L1_writer.py
```diff
@@ -312,30 +312,55 @@ def _add_datasets_to_pixel_offset_group(self):
...
                         Units.unitless,
+                        None,
+                    ),
+                    (
+                        "slantRangeOffset",
+                        np.float32,
+                        "Slant range offset",
+                        Units.meter,
+                        None,
                     ),
```
bhawkins, line 333:
```suggestion
```
Duplicated

python/packages/nisar/products/insar/InSAR_L1_writer.py
```diff
@@ -542,8 +570,19 @@ def add_pixel_offsets_to_swaths_group(self):
...
+                    f"{offset_group_name}/{pol}"
+                )
+                offset_pol_group = self.require_group(offset_pol_group_name)
+
+                # Extract polarization-dependent valid mask
+                valid_mask = extract_pol_valid_mask(pol_valid_mask, pol)
+
+                offset_pol_group['validDataMask'][...] = valid_mask
+                offset_pol_group['validDataMask'].attrs['valid_min'] = np.uint8(0)
+                offset_pol_group['validDataMask'].attrs['long_name'] = to_bytes("Valid data mask")
```
bhawkins, line 585: Would be good to add scalar attributes that summarize the overall fraction of valid pixels.

python/packages/nisar/products/insar/InSAR_L1_writer.py
```diff
@@ -787,19 +826,39 @@ def add_interferogram_to_swaths_group(self, is_unwrapped=False):
...
                         units=ds_unit,
+                        fill_value=fill_value
                     )
+                    if ds_name == 'validDataMask':
+                        # Extract polarization-dependent valid mask
+                        valid_mask = extract_pol_valid_mask(pol_valid_mask, pol)
+
+                        igram_pol_group['validDataMask'][...] = valid_mask
+                        igram_pol_group['validDataMask'].attrs['valid_min'] = np.uint8(0)
+                        igram_pol_group['validDataMask'].attrs['long_name'] = to_bytes("Valid data mask")
```
bhawkins, line 861: Ditto here, some summary attributes would be nice.

python/packages/nisar/workflows/geocode_insar.py
```diff
@@ -119,6 +119,69 @@ def get_mask_ds_input_output(src_freq_path, dst_freq_path, input_hdf5,
 
     return input_rasters, dataset_paths
 
+def get_valid_mask_input_output(src_freq_path, dst_freq_path, pol, input_hdf5,
+                                input_product_type=InputProduct.RUNW,
+                                is_runw_offset_product = False):
+    """Create input raster objects and output dataset paths for valid masks
+
+    Collects the validMask datasets associated with the given frequency and
```
bhawkins, line 127:
```suggestion
    Collects the validDataMask datasets associated with the given frequency and
```

python/packages/nisar/workflows/geocode_insar.py
```diff
@@ -119,6 +119,69 @@ def get_mask_ds_input_output(src_freq_path, dst_freq_path, input_hdf5,
...
+            dst_group_paths.append(f'{dst_freq_path}/pixelOffsets/{pol}')
+         else:
+            src_group_paths.append(f'{src_freq_path}/interferogram/{pol}')
+            dst_group_paths.append(f'{dst_freq_path}/unwrappedInterferogram/{pol}')
+    elif input_product_type is InputProduct.RIFG:
+        src_group_paths.append(f'{src_freq_path}/interferogram/{pol}')
+        dst_group_paths.append(f'{dst_freq_path}/wrappedInterferogram/{pol}')
+    elif input_product_type is InputProduct.ROFF:
+        src_group_paths.append(f'{src_freq_path}/pixelOffsets/{pol}')
+        dst_group_paths.append(f'{dst_freq_path}/pixelOffsets/{pol}')
```
bhawkins, line 174:
```suggestion
        dst_group_paths.append(f'{dst_freq_path}/pixelOffsets/{pol}')
    else:
        assert False, "invalid product"
```

python/packages/nisar/workflows/geocode_insar.py
```diff
@@ -119,6 +119,69 @@ def get_mask_ds_input_output(src_freq_path, dst_freq_path, input_hdf5,
...
+
+    # prepare input valid mask raster
+    for src_group_path, dst_group_path in zip(src_group_paths,dst_group_paths):
+        input_raster_str = f"HDF5:{input_hdf5}:/{src_group_path}/validMask"
```
bhawkins, line 178:
```suggestion
        input_raster_str = f"HDF5:{input_hdf5}:/{src_group_path}/validDataMask"
```

python/packages/nisar/workflows/geocode_insar.py
```diff
@@ -119,6 +119,69 @@ def get_mask_ds_input_output(src_freq_path, dst_freq_path, input_hdf5,
...
+def get_valid_mask_input_output(src_freq_path, dst_freq_path, pol, input_hdf5,
+                                input_product_type=InputProduct.RUNW,
+                                is_runw_offset_product = False):
+    """Create input raster objects and output dataset paths for valid masks
+
+    Collects the validMask datasets associated with the given frequency and
+    polarization, and pairs each one with the HDF5 path it should be written
+    to in the geocoded output product. Which groups are visited depends on the
+    input product type: RUNW yields both the pixel offsets and interferogram
+    masks, while RIFG and ROFF each yield a single mask.
```
bhawkins, line 131: This is super confusing. The output lists always have length one. Maybe we can simplify the interface.

python/packages/nisar/products/insar/utils.py
xhuang-jpl: use the np.rint instead of np.round

python/packages/nisar/products/insar/utils.py
xhuang-jpl: updated

python/packages/nisar/products/insar/utils.py
xhuang-jpl: updated

python/packages/nisar/products/insar/utils.py
xhuang-jpl: updated

python/packages/nisar/products/insar/InSAR_L1_writer.py
xhuang-jpl: Fixed.

python/packages/nisar/workflows/geocode_insar.py
xhuang-jpl: Fixed

python/packages/nisar/products/insar/utils.py
xhuang-jpl: Fixed

python/packages/nisar/products/insar/utils.py
xhuang-jpl: Fixed

python/packages/nisar/products/insar/utils.py
xhuang-jpl: Put the valid check inside

python/packages/nisar/products/insar/utils.py
xhuang-jpl: Fixed, but the _chunk_lines seems cannot be deleted since it is used by the _ensure function

python/packages/nisar/products/insar/utils.py
xhuang-jpl: Fixed

python/packages/nisar/products/insar/utils.py
xhuang-jpl: Fixed

python/packages/nisar/products/insar/utils.py
xhuang-jpl: Fixed

python/packages/nisar/products/insar/utils.py
xhuang-jpl: Fixed

python/packages/nisar/products/insar/InSAR_L1_writer.py
xhuang-jpl: @hfattahi should we copy the scalar attributes from the RSLC or we compute them ourselves for insar?

python/packages/nisar/workflows/geocode_insar.py
xhuang-jpl: Fixed

python/packages/nisar/workflows/geocode_insar.py
xhuang-jpl: Fixed

python/packages/nisar/workflows/geocode_insar.py
xhuang-jpl: Updated the docstring

python/packages/nisar/products/insar/utils.py
xhuang-jpl: the RSLC reader sizes the vector to numberOfSubSwaths and leaves any sub-swath whose validSamplesSubSwath{i} dataset is missing as a 0-size
 entry, so a product may report N sub-swaths with some empty. Falling through to the indexing branch below would raise  IndexError on the 0-length array.

python/packages/nisar/products/insar/utils.py
```diff
@@ -503,117 +668,213 @@ def generate_insar_mask(ref_rslc_obj,
...
+    # columns are zeroed at the end of each row
+    rg_gather = np.clip(rg_idx_int, 0, ref_swath.samples - 1)
+
+    mask = np.zeros((len(azi_idx_arr), len(rg_idx_arr)), dtype=np.uint32)
+    # Polarization dependent valid mask
+    pol_valid_mask = np.zeros((len(azi_idx_arr), len(rg_idx_arr)), dtype=np.uint16)
+
+    for out_row, ref_azi in enumerate(azi_idx_arr):
+        # Rows outside the reference radar grid stay 0
+        if not (0 <= ref_azi < ref_swath.lines):
```
xhuang-jpl:
```suggestion
         ref_azi_int = round(ref_azi)
```

python/packages/nisar/products/insar/utils.py
```diff
@@ -503,117 +668,213 @@ def generate_insar_mask(ref_rslc_obj,
...
+            continue
+
+        # Geometric coregistration offsets at the truncated reference
+        # indices of the output pixels
+        ref_azi_int = round(ref_azi)
```
xhuang-jpl, line 786:
```suggestion
        # indices of the output pixels
```

python/packages/nisar/products/insar/utils.py
bhawkins: We determined that this check is necessary because the `isce3::product::SubSwaths` class has a `numSubSwaths` setter method ([here](https://github.com/isce-framework/isce3/blob/35e48291d127a3408d1ef44d4f185713d47d284b/cxx/isce3/product/SubSwaths.h#L69)) that default initializes the new storage. So it is technically possible to load the sub-swaths from a product and then mutate the class afterwards such that we need this check. IMO that's poor class design and `SubSwaths` should be RAII/read-only, but that's beyond the scope of this PR.

python/packages/nisar/products/insar/InSAR_L1_writer.py
bhawkins: We decided the workflow should compute the scalar attributes because it's the intersection (reference & secondary) that is most relevant for InSAR products. It looks like this has been implemented in [3e2bad2](https://github.com/isce-framework/isce3/pull/379/commits/3e2bad211f9f8b4e0f5bbe6265e9fe7f9e846f98)

bhawkins (approved):
LGTM! Thanks for all your hard work on this patch.

hfattahi:
@xhuang-jpl Preferably let's merge Jungkyo's PR first https://github.com/isce-framework/isce3/pull/381 as there seems to be minor conflict that you will need to resolve.

hfattahi (approved):
LGTM! Thanks @xhuang-jpl for this great addition.

PR #381, opened 2026-09-22
Fix insar-mask showing wrong values in reference and secondary subswaths

oberonia78:
This PR fixes two issue in insar mask. 

image

1) in `geocode_insar.py`, we are using `geocode_obj = isce3.geocode.GeocodeFloat32()` for mask. Where ionosphere_mask =1, and reference_subswath =1 and secondary subswath =1, water_flag = 0, then pixel value become `16777227` in mask layer. However, we are using float32, so `np.uint32(np.float32(16777227)) = 16777228` . Now, the meaning is different (ionosphere_mask =1, and reference_subswath =1 and secondary subswath =2, water_flag = 0). So, the solution is that we need to use float64 for mask layer.

2) in `geocode_insar.py` , the `add_water_to_mask()` function contains the following logic:
```
subswath_mask = mask_layer & 0xFF
subswath_mask[valid] += (100 * water_mask)[valid]
mask_layer = mask_layer | subswath_mask
```

Because of the float32 geocoding issue described above, a pixel with valid reference and secondary subswaths can be changed from: 11 → 12. For a water pixel, add_water_to_mask() then adds 100, then 12 → 112. but the code combines the original mask value and the updated subswath value using bitwise OR: 12 | 112 --> 124. This explains why unexpected values, including 4, appear in secondary_subswath. These are not independent bit fields, so the original and updated low-byte values must not be combined using bitwise OR.

The below is the masks from corrected code. 
image


python/packages/nisar/workflows/geocode_insar.py
hfattahi, line 738: I think this line is not needed any more as you have a new instance of geocode for integer data.

python/packages/nisar/workflows/geocode_insar.py
```diff
@@ -715,7 +733,7 @@ def cpu_run(cfg, input_hdf5, output_hdf5, input_product_type=InputProduct.RUNW):
 
                 desired = ["mask"]
                 geocode_obj.data_interpolator = 'NEAREST'
```
hfattahi: It seems that this line is not needed any more as you have a new instance of geocode for integer data.

python/packages/nisar/workflows/geocode_insar.py
```diff
@@ -737,7 +755,7 @@ def cpu_run(cfg, input_hdf5, output_hdf5, input_product_type=InputProduct.RUNW):
...
```
hfattahi: perhaps this line not needed anymore

python/packages/nisar/workflows/geocode_insar.py
```diff
@@ -815,7 +833,7 @@ def cpu_run(cfg, input_hdf5, output_hdf5, input_product_type=InputProduct.RUNW):
...
```
hfattahi: similar to above, I think this line not needed.

python/packages/nisar/workflows/geocode_insar.py
hfattahi, line 821: Similar to above this line seems like not needed.

xeno0:
Can you also check if the GPU also have this issue?

python/packages/nisar/workflows/geocode_insar.py
```diff
@@ -264,7 +264,13 @@ def add_water_to_mask(cfg, freq, geogrid, dst_h5,
...
             subswath_mask = mask_layer & 0xFF # get the subswath mask
-            subswath_mask[mask] += (100 * water_mask)[mask].astype(np.uint8)
+            subswath_mask[mask] = (
+                subswath_mask[mask] % np.uint32(100)
+                + np.uint32(100) * water_mask[mask].astype(np.uint32)
+            )
+            mask_layer[mask] = (
+                mask_layer[mask] & np.uint32(0xFFFFFF00)
+            ) | subswath_mask[mask]
             dst_h5[mask_h5_path][...] = mask_layer | subswath_mask
```
hfattahi: The changes before this line looks fine to me. I don't understand why this line is still needed though! The three lines above, has places subswath_mask in the lower 8 bits of mask_layer. Now why do we need a bitwise or here between mask_layer and subswath_mask?
Shouldn't this be dst_h5[mask_h5_path][...] = mask_layer

hfattahi (commented):
Thanks @oberonia78 . Please see my comments below.

python/packages/nisar/workflows/geocode_insar.py
hfattahi: To be clear, I mean line 728: geocode_obj.data_interpolator = 'NEAREST' .
Same clarification for comments below.

oberonia78: typo

python/packages/nisar/workflows/geocode_insar.py
```diff
@@ -264,7 +264,13 @@ def add_water_to_mask(cfg, freq, geogrid, dst_h5,
             masked_water_mask = water_mask[mask]
             # Add the water mask to the mask layer
             subswath_mask = mask_layer & 0xFF # get the subswath mask
-            subswath_mask[mask] += (100 * water_mask)[mask].astype(np.uint8)
```
oberonia78, line 267: add more descriptions for mask and subswath_mask

python/packages/nisar/workflows/geocode_insar.py
```diff
@@ -264,7 +264,13 @@ def add_water_to_mask(cfg, freq, geogrid, dst_h5,
             masked_water_mask = water_mask[mask]
             # Add the water mask to the mask layer
             subswath_mask = mask_layer & 0xFF # get the subswath mask
```
oberonia78: subswath_water_mask

xhuang-jpl (approved):
LGTM, thanks @oberonia78

python/packages/nisar/workflows/geocode_insar.py
bhawkins: I was completely lost until I fully understood the packing of the lower 8 bits of the InSAR mask dataset.  Here's the relevant excerpt from the description attribute of the HDF5 dataset:
> Bits 0-7 represent subswath encoding, where the most significant digit represents the water flag of that pixel in the reference RSLC, where 1 is water and 0 is non-water; the second most significant digit corresponds to the subswath number of the reference RSLC, and the least significant digit corresponds to the subswath number of the secondary RSLC; a value of 0 in either digit indicates an invalid sample in the corresponding RSLC.

Re-phrasing for myself based on today's discussion, `mask_layer & 0xFF` encodes four different things:
* The special value of 255 is reserved for invalid data (`_FillValue` attribute). If the pixel has this value, the other three quantities are undefined.
* The secondary subswath (1-based) index is given by `(mask_layer & 0xFF) % 10`, with 0 meaning invalid.
* The reference subswath (1-based) index is given by `((mask_layer & 0xFF) // 10) % 10`, with 0 meaning invalid.
* The water mask is given by `((mask_layer & 0xFF) // 100) % 2` with the interpretation 1: water, 0: not-water.

bhawkins (approved):
LGTM thanks

python/packages/nisar/workflows/geocode_insar.py
bhawkins: Another thing I was confused about was we have to mask the lower 8 bits first since we switch from decimal to binary encoding. That is, a mask value like 367 should be interpreted to mean
* Valid (not equal to `_FillValue`)
* Secondary subswath 1
* Reference subswath 1
* Is water

since `367 & 0xFF == 111`.  Skipping the 8-bit mask would give the wrong answer.

I also suppose my description above was a little wrong with respect to `_FillValue` because you have to look at all bits (`is_valid = mask_layer == 255`), not just the lower 8 bits.  That is, you could have a `mask_layer` value like 511 which doesn't equal 255 even though the lower 8 bits do. I guess all those `0xff | n` cases are sort of ill-formed, though, since `water_digit == 2` isn't defined. Similarly, for `mask_layer & 0xFF` in the range [200, 255] it's unclear whether the subswath index digits should be interpreted or not. But if we stick to the spec then this isn't really a problem.

python/packages/nisar/workflows/geocode_insar.py
oberonia78: deleted.

python/packages/nisar/workflows/geocode_insar.py
oberonia78: deleted.

python/packages/nisar/workflows/geocode_insar.py
oberonia78: deleted.

python/packages/nisar/workflows/geocode_insar.py
oberonia78: No. We need it for the coherence.

python/packages/nisar/workflows/geocode_insar.py
oberonia78: The line has been updated.

python/packages/nisar/workflows/geocode_insar.py
oberonia78: corrected.

hfattahi (approved):
LGTM. Thanks @oberonia78

